PR #7, opened 2018-10-28
Don't insert duplicate ip addresses to the openvpn config file

janaz:
A new set of IP addresses is appended on each restart of the container which eventually results in the openvpn error `Maximum number of 'remote' options (64) exceeded.`


janaz:
The test is failing because of the incorrect md5 hash of `named.root` file. It's fixed in #8

qdm12:
Hi Janaz !

Thanks a lot for your pull request !

- The IP bug is a bug I indeed saw from time to time, thanks for fixing it !
- Thanks for updating the md5 digest and named.roots !

A little note that the `--no-cache` flag might not be useful for multi stage builds as the previous images are discarded.

PR #32, opened 2019-07-15
Fix/improve port forwarding handling

### LINKED ISSUE #31: unknown option "--dport"
Hi !

I'm getting this error when running `portforward.sh`:

````
# docker exec -it pia /portforward.sh
Forwarded port for IP 194.187.249.57 is: 32233
Detecting target VPN interface...tun0
Accepting input traffic through tun0 to port 32233...iptables v1.8.3 (legacy): unknown option "--dport"
Try `iptables -h' or 'iptables --help' for more information.
[ERROR] Undefined error, with status 2)
````

You should use the `-p` option in your `iptables` command, and probably run it twice (one for TCP and one for UDP).

### PR DESCRIPTION
**Pull request by gdlx**
Here's my proposal to fix #31 and apply @gurabli proposal too.

If `PORT_FORWARDING` variable is set to `true`, `portforward.sh` is automatically started in background 10 seconds after OpenVPN command is launched, to let the time to connect before reading the port.

It is then written into a file to allow reading it from another container.

Note the script can still be launched manually as before so there's no breaking change for people already using it that way.

`iptables` command is fixed an split into `tcp` and `udp` rules.

I've also installed `jq` to parse the Json result. This is more robust than a `grep`.

Here's the container log with enabled port forwarding:

````
...
[INFO] Launching OpenVPN
Mon Jul 15 19:36:27 2019 OpenVPN 2.4.7 x86_64-alpine-linux-musl [SSL (OpenSSL)] [LZO] [LZ4] [EPOLL] [MH/PKTINFO] [AEAD] built on May  5 2019
Mon Jul 15 19:36:27 2019 library versions: OpenSSL 1.1.1c  28 May 2019, LZO 2.10
Mon Jul 15 19:36:27 2019 TCP/UDP: Preserving recently used remote address: [AF_INET]185.220.70.135:1198
Mon Jul 15 19:36:27 2019 UDP link local: (not bound)
Mon Jul 15 19:36:27 2019 UDP link remote: [AF_INET]185.220.70.135:1198
Mon Jul 15 19:36:27 2019 WARNING: this configuration may cache passwords in memory -- use the auth-nocache option to prevent this
Mon Jul 15 19:36:27 2019 [f1ea392332562703ee6a20a293eb5f2a] Peer Connection Initiated with [AF_INET]185.220.70.135:1198
Mon Jul 15 19:36:28 2019 TUN/TAP device tun0 opened
Mon Jul 15 19:36:28 2019 /sbin/ip link set dev tun0 up mtu 1500
Mon Jul 15 19:36:28 2019 /sbin/ip addr add dev tun0 local 10.85.10.6 peer 10.85.10.5
Mon Jul 15 19:36:28 2019 Initialization Sequence Completed
[INFO] Reading forwarded port
 * Forwarded port for IP  is: 40519
 * Written forwarded port to /forwarded_port
 * Forwarded port for IP 185.220.70.135 is: 40519
 * Detecting target VPN interface...tun0
 * Accepting input traffic through tun0 to port 40519...DONE
````

### CODE REVIEW

**Comment by gdlx** [comment]
A question also: why do you use `printf` everywhere instead of `echo` ?
It makes the script less readable IMO.
If you needed this to avoid the line ending on some log entries, `echo -n` does the trick 😉

File: README.md [hunk 1]
```diff
@@ -154,6 +154,7 @@ docker run --rm --network=container:pia alpine:3.10 wget -qO- https://ipinfo.io
 | `BLOCK_NSA` | `off` | `on` or `off`, blocks NSA hostnames |
 | `UNBLOCK` | | comma separated string (i.e. `web.com,web2.ca`) to unblock hostnames |
 | `EXTRA_SUBNETS` | | comma separated subnets allowed in the container firewall (i.e. `192.168.1.0/24,192.168.10.121,10.0.0.5/28`) |
+| `PORT_FORWARDING` | | Set to `true` to read the forwarded port |
```
qdm12: Can you set the default to `false` thanks!

portforward.sh
```diff
@@ -19,16 +20,20 @@ if [ "$json" == "" ]; then
...
+printf " * Forwarded port for IP $ip is: $port\n"
+printf " * Detecting target VPN interface..."
 TARGET_PATH="/openvpn/target"
 vpn_device=$(cat $TARGET_PATH/config.ovpn | grep 'dev ' | cut -d" " -f 2)0
 exitOnError $?
 printf "$vpn_device\n"
-printf "Accepting input traffic through $vpn_device to port $port..."
-iptables -A INPUT -i $vpn_device --dport $PORT -j ACCEPT
+printf " * Accepting input traffic through $vpn_device to port $port..."
+iptables -A INPUT -i $vpn_device -p tcp --dport $port -j ACCEPT
```
qdm12: Do you actually have to specify the protocol `tcp`for example? I thought not specifying it would block all protocols.

entrypoint.sh
```diff
@@ -395,6 +393,14 @@ if [ "$PROXY" == "on" ]; then
   printf "DONE\n"
 fi
 
+############################################
+# READ FORWARDED PORT
+############################################
+
+if [ "$PORT_FORWARDING" == "true" ]; then
+  sleep 10 && /portforward.sh &
```
qdm12: Instead of sleeping 10 seconds, get your public IP before using the VPN (using `wget -qO- https://ipinfo.io/ip`) and check for an IP change every 2 seconds? (not too short or it'll be rate limited)

That will work better on very slow connections and is more resilient in my opinion.

portforward.sh
```diff
@@ -19,16 +20,20 @@ if [ "$json" == "" ]; then
     printf "Port forwarding is already activated on this connection, has expired, or you are not connected to a PIA region that supports port forwarding\n"
     exit 1
 fi
-port=`echo $json | grep -Eo [0-9]{3,5}`
```
qdm12: I would rather keep it to using grep to reduce the size and have less dependencies. The json answer is quite simple so no need for jq I think.

qdm12:
I use printf because it's quicker (lol it is) and has formatting built-in and I prefer the trailing `\n` than the `-n` but that's really a silly preference :smile:

gdlx:
> * Never mind but please add that it's written to a file in the readme.

Didn't I ?

> * You could also read the file to check if a port has already been forwarded.

The forwarded port changes on each VPN session, so the file would only be reusable for manual script run, but there would be no guarantee the value doesn't come from a previous session. I don't think the file should be read by the script itself...

> * I like the `&` sign to mix the stdouts together

The goal was more to read the port in background, because openvpn is blocking the script, and the port request has to be done after the session has started.

A cleaner way would be start openvpn in background too, but it's not that easy...

> * I'll probably mix unbound and tinyproxy stdout as well, what do you think about that?

I didn't dig into those parts, but if they can be done before openvpn session is started, then you should stick like this.

README.md
gdlx: It is: https://github.com/qdm12/private-internet-access-docker/blob/55492015cbb62d47d1f97e6067e64f09cd234f18/Dockerfile#L33

Do you mean writing the default value in the Readme ? I think asking to set it to `true` to enable the feature implicitely means it's not enabled by default...

portforward.sh
gdlx: `jq` is a really common package, it just adds 22k to the image and no other dependencies.

Parsing Json with `grep` is really really weak, should PIA add a space or another element in the hash and you script is broken.

I like small images too, but I would not push the zeal to the point of using broken script to save a few kB...

gdlx:
> I use printf because it's quicker (lol it is) and has formatting built-in and I prefer the trailing `\n` than the `-n` but that's really a silly preference 😄

https://unix.stackexchange.com/questions/58310/difference-between-printf-and-echo-in-bash/77564#77564

You're saving 1 second over 1 million execs, and I have inverted results on my laptop:
````
$ time for i in {1..999999}; do echo "$i" >/dev/null; done
real    0m19.731s
user    0m7.016s
sys     0m12.719s

$ time for i in {1..999999}; do printf "$i" >/dev/null; done
real    0m20.228s
user    0m7.469s
sys     0m12.750s
````

You can't really take this as a serious argument. You could use `printf` instead of `echo -n` but keep `echo` to improve readability and use the common standard for most cases.

entrypoint.sh
gdlx: I'll can try this but it will complicate the script and likely make a breaking change with the previous behavior, because the timing logic will have to be included in `porforward.sh`.

README.md
qdm12: Yes but I have a beautiful Default column :smile:

entrypoint.sh
qdm12: Yes you are right, let's keep it that way for this PR, I'll try doing it in a week or so.

portforward.sh
qdm12: Ok fine, let's keep jq onboard then.

qdm12:
Yes of course it was not a serious argument (that's why there was a *lol*) :smile:

Anyway it's printf now, let's keep it like that as readability is subjective, i prefer trailing `\n` when needed whereas you prefer `-n` flag to echo. In my opinion, more will know what `\n` is than what the `-n` flag is to echo.

README.md
gdlx: Sorry, didn't notice, I'll fix that 😊

gdlx:
> Can you also remove the instructions in the readme on how to port forward manually, your way should be the only way from now on (although it's retro compatible)

Done

README.md
gdlx: Done

gdlx:
I've added a few more changes:
- I removed the `2>/dev/null` because `-q0-` already hides errors:
https://github.com/qdm12/private-internet-access-docker/blob/86957a359eb80cd5e84c48e890e43fe930b01423/portforward.sh#L17
- I've added messages for your `exitOnError` function, because the log was useless with all errors being undefined. But honestly, I'd prefer adding `set -e` in the beginning of the script and keep original errors. This often makes troubleshooting easier than "pretty" error messages.
- I've changed this message to make it clear the displayed IP is the VPN remote one:
https://github.com/qdm12/private-internet-access-docker/blob/59c9f028af3e747dcaa6f7f42c9e7dbd0f0d1b10/portforward.sh#L28

gurabli:
Great work! I stopped using port forwarding as it is a potential security risk, but I do have to use it on some trackers.

gdlx:
> Great work! I stopped using port forwarding as it is a potential security risk, but I do have to use it on some trackers.

The risk is low in an unprivileged container.

qdm12:
Hi all, 

@gdlx thanks a lot for your great work. Sorry for being slow to respond, I was away from Internet recently...
Everything sounds good, I'm merging the pull request!

In the next update, I'll also add MUTE_* env variables to allow optionally to mix the logs of tinyproxy and unbound with the openvpn logs.

gdlx:
You should have asked Github to squash before merge, or let me time to clean the PR before merging 😥
No serious consequence though, thanks !

Now I'm wondering if port forwarding could really work for an app being in another container than the VPN one...seems it may need routing iptables rules which would be a nightmare to configure dynamically.

Maybe this feature would be easier to use with the app being bundled in a child image, with your image as parent...it would also be easier to manage port update in the app config. Port forwarding can't be shared by multiple apps anyway...

gurabli:
Well, this is what I wanted to tell, it might not work with other containers, and doing iptables and routing for this, well, no thanks.

Going against Docker rule, but yes, the base VPN docker + the client qBittorrent, Deluge, ruTorrent, Transmission.

qdm12:
Thanks! I did not try port forwarding to be honest, but don't connected containers with `--network="container:pia" ` be accessed through this port forwarding?
If not, maybe writing a script reading your file with the port number and then asking to choose one of the LAN hostname or IP address (so connected to the container) to port forward could work. The issue however is when a reconnection happens, I'm not sure what would happen.

gdlx:
I'm very new to OpenVPN and this kind of containers usage to be honest, so I don't understand the network flow very well when using `--network="container:pia"`.

I'm in the process of forking your image to add my app into it to try with a single container. If it works, I'll try to split my additions into a child container with yours as parent, then I'll submit a new PR here if some changes are needed to make this work.

qdm12:
Yes, although less modular/Dockerish, we could make different Docker image tags with common apps built-in.

But I'll try to figure out the iptables, it shouldn't be too hard, although probably a lot of testing would be involed!

I have currently no internet access at home so it'll likely take a week or so for me to look into it.

groenator:
I condigure port forwarding, everytime I start the container I get this error. Can someone advise why? 

vpn_1     |  * Written forwarded port to /forwarded_port
vpn_1     | [ERROR] Unable to read remote VPN IP, with status 5)

qdm12:
Hi there, is this a new error or did it exist before today? I have changed several things in the code and maybe broke something today.

groenator:
This is an error I got when I started the container.

qdm12:
Hi @groenator 

I have pushed a rework of the port forwarding script.
It might now warn you it cannot get the remote VPN IP but won't fail over it.

For your information, this error is due to `wget -qO- https://duckduckgo.com/\?q=ip` failing, maybe duckduckgo.com is inaccessible. It's only used to show you your public VPN IP address for convenience, so it's not a critical point.

Let me know how it goes!

groenator:
Hi, thanks for fixing it. I will try it tonight and let you know how it
goes.

Regards,

groenator:
Hi there,

I tried the new image and works as expected. Here are the output logs;

vpn_1     | Tue Sep 10 21:28:21 2019 TUN/TAP device tun0 opened
vpn_1     | Tue Sep 10 21:28:21 2019 /sbin/ip link set dev tun0 up mtu 1500
vpn_1     | Tue Sep 10 21:28:21 2019 /sbin/ip addr add dev tun0 local
10.79.10.6 peer 10.79.10.5
vpn_1     | Tue Sep 10 21:28:21 2019 Initialization Sequence Completed
vpn_1     | [INFO] Reading forwarded port
vpn_1     |  * Generating client ID...
vpn_1     |  * Obtaining forward port from PIA server...
vpn_1     |  * Parsing JSON response...
vpn_1     |  * Writing forwarded port to file...
vpn_1     |  * Detecting current VPN IP address...
vpn_1     |  * Forwarded port accessible at 185.220.70.130:38776
vpn_1     |  * Detecting target VPN interface...
vpn_1     |  * Accepting input traffic through tun0 to port 38776...
vpn_1     | [INFO] Port forwarded successfully

Thanks for helping out.

PR #43, opened 2019-09-02
Make forwarded_port file location configurable

Juggels:
When running this container in a Kubernetes pod, it is not possible to directly mount the `/forwarded_port` file from other containers in the pod.

Normally this can be accomplished by creating a shared volume, and mounting this volume in all required containers. Because this container writes the file to `/` by default, it is currently impossible to use this technique, as it would try to mount the given volume on the root filesystem.

This pull request makes the folder in which the file is written configurable, so that a subfolder of the root filesystem can be used.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,6 +31,7 @@ ENV USER= \
     UNBLOCK= \
     EXTRA_SUBNETS= \
     PORT_FORWARDING=false \
+    PORT_FORWARDING_STATUS_FILE="/forwarded_port" \
     PROXY=on \
     PROXY_LOG_LEVEL=Critical \
     PROXY_PORT=8888 \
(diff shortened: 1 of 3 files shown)


qdm12:
Thanks for the PR, merging it now.

Raph521:
While testing out the change to correct the failed wget command due to cert auth failure, I noticed that this change  introduced another bug:

```
[INFO] Reading forwarded port,
/portforward.sh: line 26: syntax error: unexpected "fi" (expecting "then")
```

Looks like there is a "then" missing the new if statement on line 26 of portforward.sh

```
if [ ! -d "${port_status_folder}" ];
  mkdir -p "${port_status_folder}"
fi
```

Juggels:
Oh snap... Don’t understand how I missed that one, my bad :-( Will create a new PR for that later tonight if @qdm12 hasn’t
Gotten to it by then

qdm12:
I have merged #44 thanks guys!

Lucky us there are multiple eyes on the latest changes 😄

PR #44, opened 2019-09-02
Fix portforward script syntax error

Juggels:
So sorry about that one 😳


qdm12:
Thanks, sorry I did not notice 😄

PR #70, opened 2020-02-02
Random region when REGION is empty

rorph:
This change picks a random `ovpn` file when `$REGION` is empty, it doesn't replace the default value tho, so you need to explicit set it, like `--e REGION=""`

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -81,6 +81,14 @@ exitIfUnset PASSWORD
 exitIfNotIn ENCRYPTION "normal,strong"
 exitIfNotIn PROTOCOL "tcp,udp"
 exitIfNotIn NONROOT "yes,no"
+
+if [ -z $REGION ]; then
+  REGION=$(ls -1 /openvpn/$PROTOCOL-$ENCRYPTION/*.ovpn | while IFS= read -r line ; do
+    printf "%06d %s\n" $RANDOM "$line"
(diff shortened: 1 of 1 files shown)


qdm12:
Thanks for the PR! :+1:

Although I'm rewriting the whole entrypoint in Golang, it's like 95% done, instead of shell scripting so I'll integrate your suggestion in the newer version to avoid myself merge conflicts :smile:

I'll make sure to give you credits for the suggestion in the code and readme. I'll leave the PR opened until then.

qdm12:
@rorph I've added your changes to the beta Docker image. You can try it if you want:

The Go rewrite Docker image is now out in 'beta' at `qmcgaw/private-internet-access:go`.

- The documentation is [here](https://github.com/qdm12/private-internet-access-docker/tree/go#private-internet-access-client)
- It should work with the same settings as before
- Feel free to comment here or on the [Slack channel](https://join.slack.com/t/qdm12/shared_invite/enQtOTE0NjcxNTM1ODc5LTYyZmVlOTM3MGI4ZWU0YmJkMjUxNmQ4ODQ2OTAwYzMxMTlhY2Q1MWQyOWUyNjc2ODliNjFjMDUxNWNmNzk5MDk)
- Note that it's lacking port forwarding for now, I will do it in the coming few days
- It might have bugs as it's not totally unit tested yet

PR #84, opened 2020-02-21
Resolve a Few Mullvad Issues

JeordyR:
Changes
* Added VPNSP to README
* Updated GetUser function to strip spaces, mullvad includes spaces on their site by default don't accept spaces through openvpn, so strip them for convenience if users copy/paste directly from site
* Corrected GetMullvadCountry param to use MullvadCountryChoices instead of MullvadCityChoices
* Updated Dockerfile to set COUNTRY to "sweden" by default instead of "Sweden" as the mappings in mullvad.go are all lower-case and the GetValueIfInside function from github.com/qdm12/golibs/params does not .ToLower in its comparison
  * May be worth updating that lib to .ToLower for its comparison, or have an IgnoreCase flag or something

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -79,4 +79,4 @@ RUN apk add -q --progress --no-cache --update openvpn ca-certificates iptables u
     adduser nonrootuser -D -H --uid 1000 && \
     chown nonrootuser -R /etc/unbound /etc/tinyproxy && \
     chmod 700 /etc/unbound /etc/tinyproxy
-COPY --from=builder --chown=1000:1000 /tmp/gobuild/entrypoint /entrypointex
+COPY --from=builder --chown=1000:1000 /tmp/gobuild/entrypoint /entrypoint
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,7 +40,7 @@ ENV VPNSP=pia \
     PORT_FORWARDING=off \
     PORT_FORWARDING_STATUS_FILE="/forwarded_port" \
     # Mullvad only
-    COUNTRY=Sweden \
+    COUNTRY=sweden \
     CITY= \
     ISP= \
     PORT= \
@@ -79,4 +79,4 @@ RUN apk add -q --progress --no-cache --update openvpn ca-certificates iptables u
     adduser nonrootuser -D -H --uid 1000 && \
     chown nonrootuser -R /etc/unbound /etc/tinyproxy && \
     chmod 700 /etc/unbound /etc/tinyproxy
-COPY --from=builder --chown=1000:1000 /tmp/gobuild/entrypoint /entrypoint
+COPY --from=builder --chown=1000:1000 /tmp/gobuild/entrypoint /entrypointex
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -40,7 +40,7 @@ ENV VPNSP=pia \
     PORT_FORWARDING=off \
     PORT_FORWARDING_STATUS_FILE="/forwarded_port" \
     # Mullvad only
-    COUNTRY=Sweden \
+    COUNTRY=sweden \
```
qdm12: It should be case insensitive right? Or did I forget to code that :eyes:

Dockerfile
```diff
@@ -79,4 +79,4 @@ RUN apk add -q --progress --no-cache --update openvpn ca-certificates iptables u
     adduser nonrootuser -D -H --uid 1000 && \
     chown nonrootuser -R /etc/unbound /etc/tinyproxy && \
     chmod 700 /etc/unbound /etc/tinyproxy
-COPY --from=builder --chown=1000:1000 /tmp/gobuild/entrypoint /entrypoint
\ No newline at end of file
+COPY --from=builder --chown=1000:1000 /tmp/gobuild/entrypoint /entrypointex
```
qdm12: What's wrong with entrypoint? :smile:

internal/params/mullvad.go
```diff
@@ -11,7 +11,7 @@ import (
 // GetMullvadCountry obtains the country for the Mullvad server from the
 // environment variable COUNTRY
 func (p *paramsReader) GetMullvadCountry() (country models.MullvadCountry, err error) {
-	choices := append(constants.MullvadCityChoices(), "")
+	choices := append(constants.MullvadCountryChoices(), "")
```
qdm12: Thanks! :+1:

internal/params/openvpn.go
```diff
@@ -21,6 +21,7 @@ func (p *paramsReader) GetUser() (s string, err error) {
 	} else if len(s) == 0 {
 		return s, fmt.Errorf("USER environment variable cannot be empty")
 	}
+	s = strings.Replace(s, ' ', '', -1)
```
qdm12: Why is this needed? You got white spaces where? Trailing spaces?

internal/params/mullvad.go
JeordyR: 👍

Dockerfile
JeordyR: Woops, fix incoming :P

internal/params/openvpn.go
JeordyR: Not 100% needed, but a nice to have if you want.

Mullvad includes spaces in the username they give you on their site by default but don't accept spaces connecting through openvpn, so stripping them for convenience if users copy/paste directly from site.

qdm12: Hmm I guess no one has a username with spaces, so good idea.

You can use `strings.ReplaceAll` instead though :wink`

qdm12: Also I think you forgot to import `"strings"` at the top so the thing does not compile.

As mentioned, you might want to try [the development container](https://github.com/qdm12/private-internet-access-docker#development) as it lints warnings/errors and auto-imports everything for you 👍

Dockerfile
qdm12: I just pushed a small commit updating golibs so it treats all environment variables without case sensitivity (lowercase everything).

Issue #108, opened 2020-03-28
Feature request: Shadowsocks client container tailored for this container


qdm12:
1. What's the feature?

Have a Shadowsocks client container (UDP+TCP proxy) tailored for connecting to this container's Shadowsocks server. Also add documentation/schema to the readme for it.

2. Why do you need this feature?

- Other containers can connect to that shadowsocks client container
- Other containers won't have to be restarted if this vpn container loses its connection
- Automatically reconnects when Shadowsocks server is back online

It should indirectly fix the following issues: #50 #82 #92 

And may fix the issues: #37 , #62 

It would allow the feature request #91 to be developed

gurabli:
Excited to see this in work.
And how should I connect a container? 
And will everything be routed only over VPN, including DNS?

qdm12:
- You can just connect to the container with `--network_mode="container:ssclient"`  for example. The difference is it should never lose its network stack unlike a vpn client container on reconnection.
- It will route only TCP and UDP (so for example no ping ICMP packets), but most traffic uses TCP (http/email/etc.) or UDP (voice/dns) so that should be fine.

gurabli:
I hope you are progressing with the container, as I keep loosing connection every day :)

Frepke:
Any new on this one?

qdm12:
Hello gurabli and Frepke, after consideration, I think that won't solve the problem as updating that new shadowsocks client container would have the same effect as the current problem, so it's really just moving the problem and not solving it.

I think #144 might be a good solution for disconnection problems.

For updating the container, stopping the container will essentially break the connection with connected containers. Maybe there is another Docker networking way to do that.

Otherwise I was thinking of a self updating program in the container to avoid stopping/restarting the container, but that will likely come later once we have a way to live change settings like a web ui and built in http server (see #91 and #147 ).

I'll close the issue but feel free to comment more 😉

PR #126, opened 2020-04-06
Fix containers connected loosing connection when openvpn reconnects

qdm12:
- `persist-remote-ip` openvpn flag added

Might cancel #108 if it proves successful

### DIFF
--- a/internal/mullvad/conf.go
+++ b/internal/mullvad/conf.go
@@ -59,6 +59,7 @@ func (c *configurator) BuildConf(connections []models.OpenVPNConnection, verbosi
 		"tls-cipher TLS-DHE-RSA-WITH-AES-256-GCM-SHA384:TLS-DHE-RSA-WITH-AES-256-CBC-SHA",
 
 		// Added constant values
+		"persist-remote-ip",
 		"mute-replay-warnings",
 		"auth-nocache",
 		"pull-filter ignore \"auth-token\"", // prevent auth failed loops
--- a/internal/pia/conf.go
+++ b/internal/pia/conf.go
@@ -102,6 +102,7 @@ func (c *configurator) BuildConf(connections []models.OpenVPNConnection, encrypt
 		"compress", // allow PIA server to choose the compression to use
 
 		// Added constant values
+		"persist-remote-ip",
 		"auth-nocache",
 		"mute-replay-warnings",
 		"pull-filter ignore \"auth-token\"", // prevent auth failed loops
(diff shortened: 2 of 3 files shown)


gurabli:
> * `persist-remote-ip` openvpn flag added
> 
> Might cancel #108 if it proves successful

I just need to pull the latest tag and this will take effect or an ENV variable is needed (PIA)?

Frepke:
> 
> 
> > * `persist-remote-ip` openvpn flag added
> > 
> > Might cancel #108 if it proves successful
> 
> I just need to pull the latest tag and this will take effect or an ENV variable is needed (PIA)?

For the test, you've to pull with the tag :persist-remote-ip instead of :latest.

gurabli:
> > > * `persist-remote-ip` openvpn flag added
> > > 
> > > Might cancel #108 if it proves successful
> > 
> > 
> > I just need to pull the latest tag and this will take effect or an ENV variable is needed (PIA)?
> 
> For the test, you've to pull with the tag `:persist-remote-ip` instead of `:latest`.

Thanks! I pulled and lets see hwo is it performing. I don't need to see anything in the logs, correct?

Frepke:
> 
> 
> > > > * `persist-remote-ip` openvpn flag added
> > > > 
> > > > Might cancel #108 if it proves successful
> > > 
> > > 
> > > I just need to pull the latest tag and this will take effect or an ENV variable is needed (PIA)?
> > 
> > 
> > For the test, you've to pull with the tag `:persist-remote-ip` instead of `:latest`.
> 
> Thanks! I pulled and lets see hwo is it performing. I don't need to see anything in the logs, correct?

The only thing changed is an extra variable in the .ovpn file. Don't think you see anything in the logs (not sure). Use the container and see how it reacts.

qdm12:
Hello all, sorry I missed those comments somehow!

It just adds `persist-remote-ip` in the ovpn config file, which might solve reconnection. On the other hand, please try #143  (tag `:restart`) (pull request #144) which basically restarts openvpn if it crashes from within the container so that linked containers don't lose their connection.

I tried manually disconnecting the ethernet cable for a few minutes, openvpn tries to restart for a few minutes, then crashes and then gets re-launched successfully. It *should* solve a bunch of problems if it works out for you 😉 

Thanks!

qdm12:
So from what I've gathered, you all don't really have problems of openvpn losing its connection (on `:latest`) but would like to restart/update the pia container without having to restart the other connected containers right?

I think anyway #144 would be better than this modification, that is, if you have openvpn disconnection issues (I don't for example).

Issue #123, opened 2020-04-03
Feature request: IP_STATUS_FILE


C2BB:
Would it be possible to include an option to write the VPN IP out to a file, similar to PORT_FORWARDING_STATUS_FILE?

C2BB:
I slapped this together and added it to my connected rtorrent container to get me by for now.

    command: /bin/sh -c 'while [ 1 ]; do \
        PIA_PORT=$(cat /config/forwarded_port); \
        PIA_IP=$(ip route list dev eth0 | awk "/^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+ via / {print \$1}"); \
        sed -i "/port_range/c\port_range = $PIA_PORT-$PIA_PORT" /config/rtorrent/rtorrent.rc; \
        sed -i "/dht.port.set/c\dht.port.set = $PIA_PORT" /config/rtorrent/rtorrent.rc; \
        sed -i "/network.local_address.set/c\network.local_address.set = $PIA_IP" /config/rtorrent/rtorrent.rc; done'

qdm12:
Nice, thanks! The `ip route list dev eth0` part is definitely useful, I'll adapt it, probably using [horrible hex codes from /proc/net/route](https://github.com/qdm12/private-internet-access-docker/blob/master/internal/firewall/route.go#L32).

It would be nice also to log the actual IP address to use for port forwarding in the Docker logs.

Thanks 👍 

I'll finish this up in the coming week.

C2BB:
Happy to help! Thank you for considering and taking the time.

qdm12:
It should write the IP address to `/ip` by default, as soon as the VPN tunnel is running. Enjoy 😉  And feel free to re-open the issue if something doesn't work.

PR #130, opened 2020-04-12
IP_STATUS_FILE and routing improvements

qdm12:
- Obtains VPN public IP address from routing table
- Logs and writes VPN Public IP address to `/ip` as soon as VPN is up
- Obtain port forward, logs it and writes it as soon as VPN is up
- Routing fully refactored and tested
- Routing reads from `/proc/net/route`
- Routing mutates the routes using `ip route ...`


qdm12:
Fix #123

PR #180, opened 2020-07-07
Stripped extraneous whitespace

WoLpH:


qdm12:
Ah thanks for chasing these whitespaces! I keep cleaning them, but they always come back...

PR #184, opened 2020-07-07
Surfshark needs persist-tun for openvpn restarts

WoLpH:


qdm12:
Hello! Thanks for the many PRs, very appreciated!

However, I think leaving `persist-tun` makes openvpn restart without first exiting. I'm trying to have the restart logic "up" in the layers, in the Go code instead of having openvpn manage it. Ideally it would be a Docker restart but because there are containers depending on it not to restart, that's not possible. Feel free to comment back 😉

WoLpH:
You might very well be right, I can't say I've really researched the matter. I noticed the original VPN files included it and OpenVPN is warning about it at every startup:
```
WARNING: you are using user/group/chroot/setcon without persist-tun -- this may cause restarts to fail
```

It seemed like something that was overlooked during implementation, but if it's intentional that's good too :)

qdm12:
Indeed thanks! I'm in the middle of a 'runner loop' refactoring (for an api then for a web ui) so yes it was planned :smile: Although I might filter out that warning, it is indeed alarming... but not needed :wink: Anyway thanks for digging through the code, I know it's not super clean (especially that monstruous main.go that I'm making leaner now) so good job on that!

Issue #203, opened 2020-07-18
Feature request: Remap PIA port forwarded to a fixed port


qdm12:
Remap PIA port forwarded to a fixed port, for example port `10000`. This would allow the user to simply listen on port 10000 no matter what the actual port forwarded is.

The rule should be (maybe tun0 or eth0):

```
iptables -t nat -A PREROUTING -i tun0 -p tcp --dport <port-forwarded> -j REDIRECT --to-port <published-local-port>
```

Changes are already implemented locally but I need to be able to test port forwarding first.

Refers to #201 
Blocked by #177

qdm12:
@iquam

Yes it's not in the Docker image yet.

Actually can you please try, assuming your container name is `gluetun` and your PIA forwarded port is `5899`:

```sh
docker exec gluetun iptables -t nat -A PREROUTING -i tun0 -p tcp --dport 5899 -j REDIRECT --to-port 9000
docker exec gluetun iptables -t nat -A PREROUTING -i tun0 -p udp --dport 5899 -j REDIRECT --to-port 9000
```

And try to change your qbittorrent to use port `9000`, does it work?

If not, can you also try:

```sh
# Delete the previous rules
docker exec gluetun iptables -t nat -D PREROUTING -i tun0 -p tcp --dport 5899 -j REDIRECT --to-port 9000
docker exec gluetun iptables -t nat -D PREROUTING -i tun0 -p udp --dport 5899 -j REDIRECT --to-port 9000
# Try these rules, on all interfaces and not just the vpn tun0 interface
docker exec gluetun iptables -t nat -A PREROUTING -p tcp --dport 5899 -j REDIRECT --to-port 9000
docker exec gluetun iptables -t nat -A PREROUTING -p udp --dport 5899 -j REDIRECT --to-port 9000
```

qdm12:
@inquam Alternatively you can try running the image with tag `:static-port-forward`. It should redirect the port to `9000`. Please let me know if it works so I can merge it in the main Docker image. Thanks!

inquam:
@qdm12: Been a busy few days so not much time in front of computer. I'll have a look asap and let you know.

qdm12:
You can still try, although with #242 you should be able to re-use the same port forward across reboots if you use the newer (v4) PIA servers (using `-e VPNSP=pia` which is the default). So I might just close the PR for the sake of keeping things simple. Although for the interest of truth, I would still be curious if this remapping works out.

qdm12:
[...]

PR #205, opened 2020-07-19
Static port forward, see #203

qdm12:
- Waiting for a way to test port forwarding (or fix it)


qdm12:
Abandoned due to limited usage.

Issue #298, opened 2020-11-12
Bug: Cannot use HTTP Proxy anymore from browsers


mberino:
**Host OS** (approximate answer is fine too): Ubuntu 18

**Is this urgent?**: No

**What VPN provider are you using**:

**What are you using to run your container?**: Docker run

**What is the version of the program** (See the line at the top of your logs)

```
Running version latest built on 2020-03-13T01:30:06Z (commit d0f678c)
```

**What's the problem** 🤔

When i try to use the HTTP Proxy from another machine on the network it says the proxy server refused connection. 

I am using the following parameters:

```sh
docker run --privileged -d --restart unless-stopped --name=pia --hostname=pia -p 8888:8888 -p 8388:8388/tcp -p 8388:8388/udp -p 8000:8000/tcp  -v /share/homes/mb/config/qmcgaw/auth.conf:/auth.conf:ro \
 --cap-add=NET_ADMIN --device=/dev/net/tun \
 -e REGION="Sweden" -e PROTOCOL=udp -e PIA_ENCRYPTION=strong \
 -e USER=<REDACTED> -e PASSWORD=<REDACTED> \
 -e SHADOWSOCKS=on -e SHADOWSOCKS_PORT=8388 \
 -e SHADOWSOCKS_PASSWORD=<REDACTED>  \
 -e HTTPPROXY=on -e HTTPPROXY_PORT=8888 \
 -e HTTPPROXY_USER=<REDACTED> -e HTTPPROXY_PASSWORD=<REDACTED>\
 -e FIREWALL_OUTBOUND_SUBNETS=192.168.18.0/24 \
   -p 8090:8090 \
 qmcgaw/private-internet-access	
```

In the logs it says 

2020-11-12T13:24:01.002Z        INFO    http proxy: 192.168.18.21:34033 unauthorized 

where the IP address is my other computers address. 

I've tried to find similar issues, but I just can't find any. 

Oh, and it used to work fine before, but I am not sure when it stopped working (I don't use the HTTP Proxy feature heavily).

**Share your logs...**

...*careful to remove i.e. token information with PIA port forwarding*

```log

2020-11-12T13:27:38.153Z        INFO    OpenVPN version: 2.4.9
2020-11-12T13:27:38.157Z        INFO    Unbound version: 1.10.1
2020-11-12T13:27:38.158Z        INFO    IPtables version: v1.8.4
2020-11-12T13:27:38.158Z        INFO    Settings summary below:
OpenVPN settings:
[...]

PR #300, opened 2020-11-15
HTTP proxy authentication problems

qdm12:
- Only accepts HTTP 1.x protocols
- Only checks the credentials when the method is `CONNECT` or the request URL is absolute
- More logging on authorization failures
- Removes the authorization headers before forwarding the HTTP(s) requests
- Refers to #298


qdm12:
Duplicate of #300

PR #302, opened 2020-11-19
Use Openvpn 2.4.8

qdm12:


qdm12:
No longer needed, the problem was on PIA's end. Let's not update to openvpn 2.5.0 for now to avoid compatibility issues.

PR #364, opened 2021-01-30
Feature: updater changes to have more VPN IP addresses

qdm12:
I'll update server information constants in a separate pull request once this one is merged. (done)

### DIFF
--- a/internal/updater/resolver.go
+++ b/internal/updater/resolver.go
@@ -5,14 +5,16 @@ import (
 	"context"
 	"net"
 	"sort"
+	"time"
 )
 
 func newResolver(resolverAddress string) *net.Resolver {
+	d := net.Dialer{}
+	resolverAddress = net.JoinHostPort(resolverAddress, "53")
 	return &net.Resolver{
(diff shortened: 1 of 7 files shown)


Frepke (commented):
Code is partly readable, but not fully understandable for me 😕 
But the result of the code is 👍

internal/updater/resolver.go
```diff
@@ -31,35 +33,76 @@ func newLookupIP(r *net.Resolver) lookupIPFunc {
...
+		if err != nil {
+			return nil, err
+		}
+		for _, ip := range newIPs {
+			key := ip.String()
+			uniqueIPs[key] = struct{}{}
+		}
+		timer := time.NewTimer(timeBetween)
+		select {
+		case <-timer.C:
```
qdm12, line 100: to move above so we don't sleep one  second for the final item

qdm12:
@Frepke I made an explanation video, hopefully it's not toooo much 😄 

https://youtu.be/_4_e-rEztDc

PR #413, opened 2021-04-03
Uplift the 'localSubnet' concept to cover all local ethernet interfaces

### LINKED ISSUE #411: Feature request: Support multiple networks within the gluetun container
**What's the feature?** 🧐

Currently the gluetun firewall only supports the default gateway/subnet when it's enabled: https://github.com/qdm12/gluetun/blob/master/cmd/gluetun/main.go#L204

My setup currently involves:
* gluetun container
* deluge container joined to it (sharing network stack)
* gluetun container joined to docker network allowing Radarr/Sonarr/etc to communicate with Deluge
* gluetun container joined to docker network allowing Traefik proxy to access Deluge's UI to enable administration/etc

Currently gluetun supports reading the default gateway/subnet and setting the `iptables` correctly to allow traffic from one of the networks (the first one, aka the default one from the containers point of view).

I'd like it if gluetun took the approach where it looked for the other network interfaces (eth1, eth2, etc) which in this case, are other docker networks attached to the gluetun container, and performed the same 'Allow each subnet through the INPUT chain'.

Looking at https://github.com/qdm12/gluetun/wiki/Environment-variables#firewall-and-routing the only other way I can see here is adding some sort of 'FIREWALL_INBOUND_SUBNETS' instead, but I'd have to hardcode the docker network CIDRs which I'd prefer not to have to do.

Does this sound sane/feasible? Or am I missing something maybe?

**Optional extra information** 🚀
N/A

michael-robbins:
We cover all links that have their encapsulation set to 'ether' (in my testing this equated to all eth0/eth1/eth2/etc).

My assumption here is we'll be executing inside a docker container, and docker networks are exposed under eth0/eth1/etc, the openvpn tunnel and the local loopback don't have this so we automatically exclude them!

I think I've got it right, but haven't tested it yet (that's next), just raising now for any feedback/direction.

Fixes #411 

Cheers!

### DIFF
--- a/internal/routing/reader.go
+++ b/internal/routing/reader.go
@@ -116,8 +116,6 @@ func (r *routing) LocalSubnets() (localSubnets []net.IPNet, err error) {
 		return localSubnets, fmt.Errorf("cannot list local routes: %w", err)
 	}
 
-	localSubnets = make([]net.IPNet, 0)
-
 	for _, route := range routes {
 		if route.Gw != nil || route.Dst == nil {
 			continue
--- a/internal/routing/reader.go
+++ b/internal/routing/reader.go
@@ -116,7 +116,7 @@ func (r *routing) LocalSubnets() (localSubnets []net.IPNet, err error) {
 		return localSubnets, fmt.Errorf("cannot list local routes: %w", err)
 	}
 
-	localSubnets = make([]net.IPNet, 1)
+	localSubnets = make([]net.IPNet, 0)
 
 	for _, route := range routes {
 		if route.Gw != nil || !localLinks[route.LinkIndex] {
--- a/internal/routing/reader.go
+++ b/internal/routing/reader.go
@@ -89,23 +89,27 @@ func (r *routing) LocalSubnet() (defaultSubnet net.IPNet, err error) {
 }
 
 func (r *routing) LocalSubnets() (localSubnets []net.IPNet, err error) {
-    links, err := netlink.LinkList()
-    if err != nil {
-        return localSubnets, fmt.Errorf("cannot find local subnet: %w", err)
-    }
+	links, err := netlink.LinkList()
+	if err != nil {
+		return localSubnets, fmt.Errorf("cannot find local subnet: %w", err)
+	}
 
-    localLinks := make(map[int]bool)
+	localLinks := make(map[int]bool)
 
-    for _, link := range links {
-        if link.Attrs().EncapType == "ether" {
-            localLinks[link.Attrs().Index] = true
-		    r.logger.Debug("local ethernet link found: %s", link.Attrs().Name)
-        }
-    }
+	for _, link := range links {
+		if link.Attrs().EncapType != "ether" {
+			continue
+		}
 
-    if len(localLinks) == 0 {
-        return localSubnets, fmt.Errorf("cannot find any local interfaces")
-    }
+		localLinks[link.Attrs().Index] = true
+		if r.verbose {
+			r.logger.Info("local ethernet link found: %s", link.Attrs().Name)
+		}
+	}
+
+	if len(localLinks) == 0 {
+		return localSubnets, fmt.Errorf("cannot find any local interfaces")
+	}
 
 	routes, err := netlink.RouteList(nil, netlink.FAMILY_ALL)
 	if err != nil {
--- a/internal/routing/reader.go
+++ b/internal/routing/reader.go
@@ -119,12 +119,12 @@ func (r *routing) LocalSubnets() (localSubnets []net.IPNet, err error) {
 			continue
 		}
 
-		localSubnet = *route.Dst
+		localSubnet := *route.Dst
 		if r.verbose {
 			r.logger.Info("local subnet found: %s", localSubnet.String())
 		}
 
-		append(localSubnets, localSubnet)
+		localSubnets = append(localSubnets, localSubnet)
 	}
 
 	if len(localSubnets) == 0 {
--- a/internal/routing/reader.go
+++ b/internal/routing/reader.go
@@ -88,6 +88,52 @@ func (r *routing) LocalSubnet() (defaultSubnet net.IPNet, err error) {
 	return defaultSubnet, fmt.Errorf("cannot find default subnet in %d routes", len(routes))
 }
 
+func (r *routing) LocalSubnets() (localSubnets []net.IPNet, err error) {
+    links, err := netlink.LinkList()
+    if err != nil {
+        return localSubnets, fmt.Errorf("cannot find local subnet: %w", err)
+    }
+
+    localLinks := make(map[int]bool)
+
+    for _, link := range links {
+        if link.Attrs().EncapType == "ether" {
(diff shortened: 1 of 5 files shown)


michael-robbins:
So running the above now I see the following:
```

File: internal/routing/reader.go [hunk 1]
```diff
@@ -88,6 +88,56 @@ func (r *routing) LocalSubnet() (defaultSubnet net.IPNet, err error) {
 	return defaultSubnet, fmt.Errorf("cannot find default subnet in %d routes", len(routes))
 }
 
+func (r *routing) LocalSubnets() (localSubnets []net.IPNet, err error) {
+	links, err := netlink.LinkList()
+	if err != nil {
+		return localSubnets, fmt.Errorf("cannot find local subnet: %w", err)
+	}
+
+	localLinks := make(map[int]bool)
```
qdm12, line 97: To make a set in Go, have the key as your set value and the value as the empty struct (=0 byte).

That way it's also clearer what the intent is and that the map value don't carry any information 😉 

```suggestion
	localLinks := make(map[int]struct{})
```

internal/routing/reader.go
```diff
@@ -88,6 +88,56 @@ func (r *routing) LocalSubnet() (defaultSubnet net.IPNet, err error) {
...
+
+	for _, link := range links {
+		if link.Attrs().EncapType != "ether" {
+			continue
+		}
+
+		localLinks[link.Attrs().Index] = true
```
qdm12, line 104:
```suggestion
		localLinks[link.Attrs().Index] = struct{}{}
```

michael-robbins:
OK found a bug in the above, I have the correct subnets, but I'm not saving the Interface that's associated with the subnet. Seeing the iptables debug logs:
```
iptables --append OUTPUT -o eth0 -s 172.18.0.2 -d 172.18.0.0/16 -j ACCEPT
iptables --append OUTPUT -o eth0 -s 172.18.0.2 -d 172.19.0.0/16 -j ACCEPT
iptables --append OUTPUT -o eth0 -s 172.18.0.2 -d 172.20.0.0/16 -j ACCEPT
iptables --append INPUT -i eth0 -d 172.18.0.0/16 -j ACCEPT
iptables --append INPUT -i eth0 -d 172.19.0.0/16 -j ACCEPT
iptables --append INPUT -i eth0 -d 172.20.0.0/16 -j ACCEPT
```

I can see it's not using `eth1` and `182.19.0.2` for `172.19.0.0/16`, which is wrong!

I'll fix it up so we pass along the interfaces and interface IP as well as the subnet (probs requires a struct to be passed around containing all of this!)

internal/routing/reader.go
```diff
@@ -88,6 +88,56 @@ func (r *routing) LocalSubnet() (defaultSubnet net.IPNet, err error) {
...
+	if len(localLinks) == 0 {
+		return localSubnets, fmt.Errorf("cannot find any local interfaces")
+	}
+
+	routes, err := netlink.RouteList(nil, netlink.FAMILY_ALL)
+	if err != nil {
+		return localSubnets, fmt.Errorf("cannot list local routes: %w", err)
+	}
+
+	localSubnets = make([]net.IPNet, 1)
```
qdm12: That will blow up 😉

You are making localSubnets as `[]net.IPNet{net.IPNet{}}` here, and then append to that. Effectively you will end up with `[]net.IPNet{net.IPNet{}, yourActualIPNetYouCareAbout}`.

Just make it 0 sized so the append will work as intended (leave it the default return value which is already a zero sized slice).

```suggestion
```

You could also, if you really need memory performance, assign it a certain capacity with `localSubnets = make([]net.IPNet, 0, 10)` for example but that's not super clear and we really don't need it in this particular code section 😉

internal/routing/reader.go
```diff
@@ -88,6 +88,56 @@ func (r *routing) LocalSubnet() (defaultSubnet net.IPNet, err error) {
...
+
+	for _, route := range routes {
+		if route.Gw != nil || !localLinks[route.LinkIndex] {
+			continue
+		}
```
qdm12, line 124: Adapting to that set-like map mentioned above + adding nil check for `route.Dst`

```suggestion
		if route.Gw != nil || route.Dst == nil {
			continue
		} else if _, ok := localLinks[route.LinkIndex]; !ok {
			continue
		}
```

qdm12 (commented):
Nice 💯  Looking good, just a few tiny fixes and comments to address.

Let me know when you test (as I have only a single eth interface in my container haha) and let's merge if it works!!

Thanks!!

internal/routing/reader.go
michael-robbins: Thanks! Yeah I found out about this and explicitly made it zero sized, but I didn't realise that it's already declared as part of the return value (fun times with go!), will remove it!

qdm12 (approved):

michael-robbins:
OK I've rejigged it a bit to actually work as expected with iptables using the correct range/interface/ip!

```
iptables --append OUTPUT -o eth0 -s 172.18.0.2 -d 172.18.0.0/16 -j ACCEPT
iptables --append OUTPUT -o eth1 -s 172.19.0.2 -d 172.19.0.0/16 -j ACCEPT
iptables --append OUTPUT -o eth2 -s 172.20.0.2 -d 172.20.0.0/16 -j ACCEPT
iptables --append INPUT -i eth0 -d 172.18.0.0/16 -j ACCEPT
iptables --append INPUT -i eth1 -d 172.19.0.0/16 -j ACCEPT
iptables --append INPUT -i eth2 -d 172.20.0.0/16 -j ACCEPT
```

We can see the correct interface names and assigned IP addresses are used!

michael-robbins:
@qdm12 if you could have another quick look at my change post-approve, that'd be grand, but I'm happy with the state of it now!

internal/routing/reader.go
```diff
@@ -9,6 +9,12 @@ import (
 	"github.com/vishvananda/netlink"
 )
 
+type LocalNetwork struct {
+	Subnet        net.IPNet
```
qdm12, line 13: Maybe change that to `IPNet` instead of `Subnet` as it's not necessarily a subnet 🤔 

Also having it as `*net.IPNet` might be more appropriate to fit how golang uses it in its standard lib (always as a pointer).

qdm12 (approved):
Just a tiny change if you can, but everything's looking good. I'll merge it this evening/tomorrow. Thanks again for the good work and all the testing 🥇! Happy you made it work for you! 

Also if you have some feedback on code quality/confusing, feel free to rant 😉 Happy to refactor in the future.

internal/routing/reader.go
qdm12: Ah no worry, I'll merge as is 😉  Thanks!

qdm12:
@michael-robbins If you want, feel free to tweak your code further to make it work for https://github.com/qdm12/gluetun/issues/418#issuecomment-817110404 😉 I didn't see that coming 😄 

I'll get to it tomorrow otherwise. Thanks!

Issue #418, opened 2021-04-09
Bug: Container stalls when Firewall is enabled


Sn0wCrack:
**Is this urgent?**: No

**Host OS** (approximate answer is fine too): Ubuntu 20.04

**CPU arch** or **device name**: amd64

**What VPN provider are you using**: Mullvad

**What are you using to run your container?**: Docker Compose

**What is the version of the program** (See the line at the top of your logs)

```
Running version latest built on 2021-04-09T17:47:52Z (commit d1558a3)
```

**What's the problem** 🤔

I've noticed that on the current latest for gluetun OpenVPN, Unbound, etc. do not start and it simply stalls after enabling the firewall rules.

When changing FIREWALL to off, it appears that everything boots just fine

**FIREWALL=on and FIREWALL_DEBUG=on**

```log
=========================================
================ Gluetun ================
=========================================
==== A mix of OpenVPN, DNS over TLS, ====
======= Shadowsocks and HTTP proxy ======
========= all glued up with Go ==========
=========================================
=========== For tunneling to ============
======== your favorite VPN server =======
=========================================
=== Made with ❤️  by github.com/qdm12 ====
=========================================

Running version latest built on 2021-04-09T17:47:52Z (commit d1558a3)


🔧  Need help? https://github.com/qdm12/gluetun/issues/new
💻  Email? quentin.mcgaw@gmail.com
☕  Slack? Join from the Slack button on Github
💸  Help me? https://github.com/sponsors/qdm12
2021/04/10 08:50:59 INFO OpenVPN version: 2.4.10
2021/04/10 08:50:59 INFO Unbound version: 1.10.1
2021/04/10 08:50:59 INFO IPtables version: v1.8.4
2021/04/10 08:50:59 INFO Settings summary below:
|--OpenVPN:
   |--Verbosity level: 1
   |--Run as root: enabled
   |--Provider:
      |--Mullvad settings:
         |--Network protocol: udp
         |--Countries: sweden
         |--Cities: stockholm
|--DNS:
   |--Plaintext address: 1.1.1.1
   |--DNS over TLS:
      |--Unbound:
          |--DNS over TLS providers:
              |--quad9
              |--cloudflare
          |--Listening port: 53
          |--Access control:
              |--Allowed:
                  |--0.0.0.0/0
                  |--::/0
          |--Caching: enabled
          |--IPv4 resolution: enabled
          |--IPv6 resolution: enabled
          |--Verbosity level: 1/5
[...]

PR #419, opened 2021-04-11
Implementing PR feedback and restricting route listing to IPv4 only

michael-robbins:
Implementing feedback from #413 and restricting route listing to IPv4 from #418


michael-robbins:
Implemented your feedback from the other PR, dunno if it's useful or not, but it seems to work as well for me with the above changes implemented!

qdm12 (approved):

PR #528, opened 2021-07-21
Update for Windscribe config

TJJP:
Here is a message from the Windscribe website: OpenVPN compression phaseout is in progress and will be completed by August 3rd 2021.
If you downloaded configs from this page before you saw this message, you need to re-download them now, or simply remove the compress or comp-lzo flags from the config that you downloaded after July 20th 2021.

I think this should fix gluetun from not being able to connect.

PR #558, opened 2021-08-08
Update Windscribe server lists

TJJP:
Fixes # https://github.com/qdm12/gluetun/pull/528#issuecomment-894707939
#528


qdm12:
I'm so sorry, this PR was perfect but I committed the same thing myself since I read a whole bunch of my emails achronologically 😢

Although I actually kept the IP3 as comment since I'm working on adding Wireguard support, so it might get useful soon!

Thanks again for taking the time to dig out that bug and even fixing it 💯

PR #578, opened 2021-08-21
IVPN server data update code and `ISP` filter

qdm12:
- Maint: Use only API, not the ZIP file, to pave the way for Wireguard
- Feat: update server information for IVPN
- Feat: add `ISP` filter for IVPN


qdm12 (commented):
Reviewed

PR #587, opened 2021-08-23
Wireguard support for Surfshark

qdm12:


HarryWindsor:
Any updates when this will be merged?

moyaco:
@qdm12 What is still needed to get this feature complete?

Number of example solutions to get the information required exists today. 
https://github.com/Incognito-Coder/Wiregen (Python)
https://github.com/PolicyPuma4/surfshark-wireguard-tunnel-generator (Golang)
https://github.com/reIyst/SSWG (Bash)

Having a solution that deals with automatically renewing the wireguard config is key.

qdm12:
Sorry for the delay answering.

There is a major refactoring I have to do in order to optionally allow network traffic to the needed API to get keys etc, before the vpn is up. And also another rather large refactoring to support renewing keys periodically. All this will likely take months I'm not gonna lie, but it's definitely on my list, especially the first one since several providers need API communication before the tunnel is up (which sucks but anyway that's how it is).

PR #806, opened 2022-01-22
add line to compose example

kzshantonu:
container_name is required to correctly refer to gluetun container from a different compose file

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,8 @@ version: "3"
 services:
   gluetun:
     image: qmcgaw/gluetun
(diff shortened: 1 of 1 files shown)


qdm12 (changes requested):
Thanks for the contribution!

However, please have it as commented out (`# ` prefixed), together with a comment stating why it should be uncommented.

Otherwise it might conflict with other gluetun instances, that's why I left it unset to have a random container name.

qdm12 (approved):

PR #915, opened 2022-03-28
Update PureVPN Servers Download URL and List

mircoianese:
Hello,

PureVPN OpenVPN configuration files download URL was wrong, so I have updated it (first commit).
Then I also run the command to update the servers list, as those were also outdated.

I've tested the changes and everything seems to be working correctly!

Only thing missing is updating the doc: https://github.com/qdm12/gluetun/wiki/PureVPN

Thanks!


internal/storage/servers.json
```diff
@@ -89601,993 +89601,1995 @@
     ]
   },
   "purevpn": {
-    "version": 2,
-    "timestamp": 1622644308,
+    "version": 3,
```
qdm12, line 89604: keep the version as `2`, this is just about the server model which is the same. Only the timestamp (auto updated) should be changed, which it is :wink: 

```suggestion
    "version": 2,
```

qdm12 (changes requested):
Awesome, thanks for fixing it!
Sorry I was away from my keyboard for a few days, so I didn't get the time to fix your issue.

Please just revert back the version number in the json file, and I'll update the Wiki!

internal/storage/servers.json
mircoianese: Sorry about that, misunderstood the version number :)
I've fixed it! 

Thank you for all your work!

qdm12 (approved):

Issue #610, opened 2021-09-10
Feature request: Wireguard configuration files reading for `VPNSP=custom`


qdm12:
⚠️ Blocked by #611 

**What's the feature?** 🧐

- [x] Parse each configuration and extract:
  - [x] Endpoint IP
  - [x] Endpoint port
  - [x] Allowed Addresses
  - [x] Private key
  - [x] Public key
  - [x] Pre-shared key
  - [x] Checks to ensure this is a client configuration, not a server one.
- [ ] Read all configuration files from `/gluetun/wireguard/configurations`
  - [ ] List all possible servers from in memory
  - [ ] Hot reload / periodic reload / reload on request of the files

qdm12:
From @ksurl:

> You can set to wg.conf as a default but configurable via an environment variable?

From @JBtje

> wg.conf sounds good, configurable via env even better

My take on it:

Let's keep it as constant `/gluetun/wireguard/wg.conf`, not configurable:

- so there is less change of error; one can always bind mount from a different file name.
- Gluetun will support multiple openvpn/wireguard config files read from a directory, so the file name won't matter anymore.

qdm12:
This is more or less done. Wireguard configuration files can be read:

- from `/run/secrets/wg0.conf`
- from `/gluetun/wireguard/wg.conf`


It doesn't read all the configuration files from `/gluetun/wireguard/configurations` or hot reload them, but I don't think that's really necessary/useful for now. Another issue can be created if there is a need for it.

PR #1120, opened 2022-08-26
feat(settings): support Wireguard ini config files

qdm12:
Resolves the first part of #610 

Read Wireguard ini file from `/gluetun/wireguard/wg.conf`


ShayBox:
What about an environment variable to set the path/name for use with multiple containers without having to create multiple directories/sub-directories and rename files?

derekcentrico:
I'm confused on this addition. The wiki is not clear whether any environment is needed when using a volume to the wg0.conf. I'd guess something specifying wireguard and custom but the rest is conflicting. What ought be done to utilize this feature?

Issue #688, opened 2021-10-22
Feature request: add ghcr.io registry as part of release build workflow


ksurl:
### What's the feature 🧐

See [here](https://github.com/ksurl/docker-buildtest/blob/main/.github/workflows/build.yml) for an example on pushing to Docker hub and Github. you will need to create a personal access token for it and add it as a repo secret.  The package will be set to private by default as well, so it needs to be set to public manually

This could be an additional option for pulling the image since docker hub has rate limiting if you're using something like watchtower to do updates.

ksurl:
Any update on this?

qdm12:
@ksurl Your PR worked flawlessly without any extra setup on my side :100: 
Image now available at `ghcr.io/qdm12/gluetun` :tada:

PR #1231, opened 2022-11-09
add ghcr image

ksurl:
Fulfills #688 

need to make package public after first build/push. I'm not sure if you also need to enable packges/container registry on your account/repo

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,6 +93,7 @@ jobs:
     permissions:
       actions: read
       contents: read
+      packages: write
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


qdm12 (dismissed):
Awesome thanks :+1: Let's merge this as is, I'll play around with it shortly

.github/workflows/ci.yml
```diff
@@ -106,6 +107,8 @@ jobs:
           flavor: |
             latest=${{ github.ref == format('refs/heads/{0}', github.event.repository.default_branch) }}
           images: |
+            ghcr.io/qdm12/gluetun
+            ghcr.io/qdm12/private-internet-access
```
qdm12, line 111: Let's keep it to gluetun; private-internet-access is for retro-compatibility only

```suggestion
```

PR #1264, opened 2022-11-30
updating fastestvpn servers

rsquarev:
I hope this will update fastestvpn servers in servers.json and config files. right now i am not able to use fastestvpn with gluetun.

Updater throws below error. I realized the url is not live anymore. i got the working url from fastestvpn support.

image

### DIFF
--- a/internal/provider/fastestvpn/updater/servers.go
+++ b/internal/provider/fastestvpn/updater/servers.go
@@ -13,7 +13,7 @@ import (
 
 func (u *Updater) FetchServers(ctx context.Context, minServers int) (
 	servers []models.Server, err error) {
-	const url = "https://support.fastestvpn.com/download/openvpn-tcp-udp-config-files"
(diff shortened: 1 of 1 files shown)


internal/provider/fastestvpn/updater/servers.go
```diff
@@ -13,7 +13,7 @@ import (
 
 func (u *Updater) FetchServers(ctx context.Context, minServers int) (
 	servers []models.Server, err error) {
-	const url = "https://support.fastestvpn.com/download/openvpn-tcp-udp-config-files"
+	const url = "http://support.fastestvpn.com/download/fastestvpn_ovpn"
```
qdm12, line 16: Use https

```suggestion
	const url = "https://support.fastestvpn.com/download/fastestvpn_ovpn"
```

qdm12 (changes requested):
Nice thanks!

Sorry I was flying around away from Internet, I'll look into this!

qdm12 (approved):

PR #1276, opened 2022-12-06
Curl Install & change flag in conf file

rivolity:
Since there is no curl installed in the image you've provided that is based on alpine 3.15, it will a good idea to add curl to check that the VPN is UP based on the IP address that is different from the ISP address.
For alpine, I've used this command to install it, 
```
apk add --update curl && \
        rm -rf /var/cache/apk/* && \
```
In the conf file of qBittorrent.conf, most users encounter a problem after using the GUI of qBittorrent for the first time, it's duo to the parameter  **HostHeaderValidation=true** that is true, and changing it to false resolved the problem for all users, but we should to connect to the container every time you redeploy the container to change this parameter.
I've added this command line on your Dockerfile to take effect correctly.

`sed -i 's/HostHeaderValidation=true.*/HostHeaderValidation=false/' /config/qBittorrent/config/qBittorrent.conf`


qdm12:
You are trying to merge 122 commits, most of them completely irrelevant to the project. On top of that, it's based on some very old version of Gluetun (3 years ago). The project has evolved a lot since then, and your PR is quite irrelevant now. If it solves your problem, good for you, but I cannot consider merging this this since it goes against the direction this project is going (small docker image, self contained binary, no shell scripting, low attack surface).

PR #1348, opened 2023-01-14
feat(settings): load wireguard inidivual fields as secret files

DennisGaida:
Currently docker secrets aren't supported for wireguard environment variables, but should.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -95,10 +95,13 @@ ENV VPN_SERVICE_PROVIDER=pia \
     # Wireguard
     WIREGUARD_CONF_SECRETFILE=/run/secrets/wg0.conf \
     WIREGUARD_PRIVATE_KEY= \
+    WIREGUARD_PRIVATE_KEY_SECRETFILE=/run/secrets/wireguard_private_key \
     WIREGUARD_PRESHARED_KEY= \
+    WIREGUARD_PRESHARED_KEY_SECRETFILE=/run/secrets/wireguard_preshared_key \
     WIREGUARD_PUBLIC_KEY= \
     WIREGUARD_ALLOWED_IPS= \
     WIREGUARD_ADDRESSES= \
+    WIREGUARD_ADDRESSES_SECRETFILE=/run/secrets/wireguard_addresses \
     WIREGUARD_MTU=1400 \
     WIREGUARD_IMPLEMENTATION=auto \
     # VPN server filtering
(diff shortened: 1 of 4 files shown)


qdm12:
Thanks for the contribution!

However, I am closing this because the implementation for secret files is not part of this PR, and because

- #1120 will get merged soon
- We can then have that single ini wireguard config file as a secret, instead of having to deal with multiple secret files

qdm12:
I fixed & updated your code (compilation, some logic was wrong etc.); however whilst doing this, it still bugs be regarding the ini wireguard secret file to be added soon. Since both are from the 'secrets' settings source, which should have precedence? The precedence system works by source and I want to keep it that way to avoid having horrible shadowy logic here and there.

I still see no point to have separate secret files instead of one standardized wireguard ini file as a secret file (most users would just bring over some wireguard ini file from somewhere).

I'll change this to a draft PR, and will likely close it once we have a secret ini file setup soon-ish (after release v3.33.0). Although you can now build the image using your `patch-1` branch and use it if you want 😉

qdm12:
We'll have secret files for everything using this precedence (blocked by #1759):

1. Secret ini file
2. Secret single values (goal of this pr)
3. ini file
4. environment variables

Both 1 and 2 will be from the same "secrets" source, so there will have to be a bit of ugly logic there to read both and merge settings, but that's fine.

You should also be able to specify a mix of all of them, for example have only a server public key in the ini file and the private key as a single secret file.

Now I'm a bit shared if 1. or 2. should be first, but to respect existing order for non secret sources, I think having the ini secret file have precedence makes more sense.

qdm12:
6096b7ad4b8278084a8d8a9c8274eb6daefbd255 reads a wireguard config file from `/run/secrets/wg0.conf` (changeable path with `WIREGUARD_CONF_SECRETFILE`), filling the first step `1.`.

I've updated your branch to fill the second step `2.` 😉 although from a users' standpoint it does exactly what you added (same paths, variable names).

3. is done with PR https://github.com/qdm12/gluetun/pull/1120 (8 months ago), and env variables were already there for 4.

All in all, I think we should be good now 😉

qdm12 (approved):

qdm12:
FYI now, if a user sets a wireguard secret config file, all fields are extracted, not just private key, preshared key and addresses. This makes more sense than before where it would only extract sensitive fields.

PR #1393, opened 2023-02-12
Add updater vars to docker-compose example

colereynolds:
I'm proposing the addition of `UPDATER_PERIOD` and `UPDATER_VPN_SERVICE_PROVIDERS` to the [docker-compose example](https://github.com/qdm12/gluetun#setup) in the README to prevent the issue described in the discussion linked below. This will help bring awareness to the static list of servers that are shipped with each release and prevent connection issues when the IPs in the user's selected region(s) have become stale and there's no clear indicator given to the users to direct them to updating their local servers.json file. Knowing that releases can happen daily or bi-annually and the updater is disabled by default, I feel this is probably a no-brainer to prevent others, especially new users, from running into the same issue.

https://github.com/qdm12/gluetun/discussions/1356#discussioncomment-4947210

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,9 @@ services:
       # - WIREGUARD_ADDRESSES=10.64.222.21/32
       # Timezone for accurate log times
       - TZ=
+      # Server list updater. See https://github.com/qdm12/gluetun/wiki/Updating-Servers#periodic-update
+      - UPDATER_PERIOD=
+      - UPDATER_VPN_SERVICE_PROVIDERS=
 ```
 
 🆕 Image also available as `ghcr.io/qdm12/gluetun`


qdm12:
Hello, first sorry for the huge delay I am just coming out of very busy times and am again able to work on Gluetun.

> there's no clear indicator given to the users to direct them to updating their local servers.json file

I think there is, if the healthcheck fails, the info log `unhealthy: <error> (see https://github.com/qdm12/gluetun/wiki/Healthcheck)` is logged, and clicking on this link shows the reasons and fixes to do. There is an even bigger log when Openvpn fails with `TLS Error: TLS key negotiation failed to occur within 60 seconds (check your network connectivity)`. Did you not see those logs in your case?

colereynolds:
I did not. If you peek at the log (in my opening comment) in the discussion linked above, all I saw was this: 
`INFO [healthcheck] program has been unhealthy for 16s: restarting VPN`

It sent me down a rabbit hole of thinking something else was wrong when really the servers for the region I had selected had all become stale. If I had seen that health check error or anything indicating that it was related to the static server list, that would have saved me time. I feel it would be best to suggest those parameters to new users so they at least know about the servers.json file and can configure the updater accordingly. Enabling the updater by default could also work but that could cause issues if say a specific provider changed their DNS names before you noticed and pushed an update for it.

qdm12:
Ok, so logic was a bit intricate to understand with the many timers and signals, but I figured it out. It would only log the wiki link when gluetun would switch from healthy to unhealthy basically, but not the first time when it would stay unhealthy from the start (and restart the vpn). That got fixed in fe97e28461d294dfd2ebc6245e1c5ee59371f73c 👍 

Now your PR still adds value, since I keep on seeing issues created by people not even trying to update the servers list, so it's after all worth including it in the minimalist example. Merging it 💯

qdm12 (approved):

qdm12:
Just FYI / don't get mad at me 😄 

After getting issues of people still unaware of updating servers data themselves, I decided to re-remove the `UPDATER_VPN_SERVICE_PROVIDERS` since in 99% of cases, the user should only care about the current vpn provider they use. On the other hand, I:

- re-worked the wiki a lot, and now there is a single page on servers data including how to list servers and how to update them
- added comments to the docker-compose.yml

      ```yml
      # Server list updater
      # See https://github.com/qdm12/gluetun-wiki/blob/main/setup/servers.md#update-the-vpn-servers-list
      - UPDATER_PERIOD=
      ```

Hopefully this will clear it up further.

There are also plenty of logged information as well as issue template description pointing to updating servers, but it seems a few people still don't want to read, so I guess it's a bit of a infinite battle against humanity 😄

colereynolds:
I think that's a great change. Hopefully that will resolve all of stale server list issues. In my opinion, if they don't read the README or the issue template before asking for help, they deserve to suffer😂

PR #1380, opened 2023-02-01
feat(nordvpn): new API endpoint and wireguard support

bubuntux:
(EDIT by qdm12):

Try it with:

```
docker run -it --rm --cap-add=NET_ADMIN -e VPN_SERVICE_PROVIDER=nordvpn -e WIREGUARD_PRIVATE_KEY=CD75jsr7Dvjt+G0IHB2gKLDJ7agWPP4x707rhcU/nnY= -e VPN_TYPE=wireguard qmcgaw/gluetun:pr-1598
```

PR #1427, opened 2023-02-27
add generics to some methods

bubuntux:
extracted from #1380


internal/configuration/settings/helpers/copy.go
```diff
@@ -141,22 +131,12 @@ func CopyIPPrefix(original netaddr.IPPrefix) (copied netaddr.IPPrefix) {
...
-		return nil
-	}
-
-	copied = make([]string, len(original))
-	copy(copied, original)
-	return copied
-}
-
-func CopyUint16Slice(original []uint16) (copied []uint16) {
+func CopySlice[K string | uint16](original []K) (copied []K) {
```
qdm12, line 134:
```suggestion
func CopySlice[K any](original []K) (copied []K) {
```

internal/configuration/settings/helpers/merge.go
```diff
@@ -139,13 +139,13 @@ func MergeWithHTTPHandler(existing, other http.Handler) (result http.Handler) {
 	return other
 }
 
-func MergeStringSlices(a, b []string) (result []string) {
+func MergeSlices[K string | uint16](a, b []K) (result []K) {
```
qdm12, line 142:
```suggestion
func MergeSlices[K comparable](a, b []K) (result []K) {
```

internal/configuration/settings/helpers/override.go
```diff
@@ -127,20 +127,11 @@ func OverrideWithHTTPHandler(existing, other http.Handler) (result http.Handler)
 	return existing
 }
 
-func OverrideWithStringSlice(existing, other []string) (result []string) {
+func OverrideWithSlice[K string | uint16](existing, other []K) (result []K) {
```
qdm12, line 130:
```suggestion
func OverrideWithSlice[K any](existing, other []K) (result []K) {
```

internal/configuration/settings/helpers/override.go
```diff
@@ -127,20 +127,11 @@ func OverrideWithHTTPHandler(existing, other http.Handler) (result http.Handler)
...
-	copy(result, other)
-	return result
-}
-
-func OverrideWithUint16Slice(existing, other []uint16) (result []uint16) {
-	if other == nil {
-		return existing
-	}
-	result = make([]uint16, len(other))
+	result = make([]K, len(other))
```
qdm12, line 134: Let's use `T` instead of `K`. `K` is used in Go example page if it's used as a key for a map, here we can just use `T` for `Type`.

internal/configuration/settings/helpers/merge.go
```diff
@@ -139,13 +139,13 @@ func MergeWithHTTPHandler(existing, other http.Handler) (result http.Handler) {
...
 	if a == nil && b == nil {
 		return nil
 	}
 
-	seen := make(map[string]struct{}, len(a)+len(b))
-	result = make([]string, 0, len(a)+len(b))
+	seen := make(map[K]struct{}, len(a)+len(b))
+	result = make([]K, 0, len(a)+len(b))
```
qdm12, line 148: Let's change `K` to `T` since from a caller perspective, `K` is not really used as a key in map, but only as the Type of elements of a slice.

internal/configuration/settings/helpers/copy.go
```diff
@@ -141,22 +131,12 @@ func CopyIPPrefix(original netaddr.IPPrefix) (copied netaddr.IPPrefix) {
...
 	if original == nil {
 		return nil
 	}
 
-	copied = make([]uint16, len(original))
+	copied = make([]K, len(original))
```
qdm12, line 139: Let's change `K` to `T` since it's not used a map key.

internal/configuration/settings/helpers/copy.go
```diff
@@ -103,16 +103,6 @@ func CopyIPNet(original net.IPNet) (copied net.IPNet) {
...
-func CopyIPNetPtr(original *net.IPNet) (copied *net.IPNet) {
-	if original == nil {
-		return nil
-	}
-
-	copied = new(net.IPNet)
-	*copied = CopyIPNet(*original)
-	return copied
-}
-
```
qdm12, line 115: Let's keep this here. I understand it's unused, but I left it there since it should be used soon (for DNS).

internal/configuration/settings/firewall.go
```diff
@@ -64,8 +64,8 @@ func (f *Firewall) mergeWith(other Firewall) {
 // settings object with any field set in the other
 // settings.
 func (f *Firewall) overrideWith(other Firewall) {
-	f.VPNInputPorts = helpers.OverrideWithUint16Slice(f.VPNInputPorts, other.VPNInputPorts)
-	f.InputPorts = helpers.OverrideWithUint16Slice(f.InputPorts, other.InputPorts)
+	f.VPNInputPorts = helpers.OverrideWithSlice(f.VPNInputPorts, other.VPNInputPorts)
+	f.InputPorts = helpers.OverrideWithSlice(f.InputPorts, other.InputPorts)
 	f.OutboundSubnets = helpers.OverrideWithIPNetsSlice(f.OutboundSubnets, other.OutboundSubnets)
```
qdm12, line 69: Could `OverrideWithIPNetsSlice` be thrown away in favor of `OverrideWithSlice`

internal/configuration/settings/firewall.go
```diff
@@ -53,8 +53,8 @@ func (f *Firewall) copy() (copied Firewall) {
 // It merges values of slices together, even if they
 // are set in the receiver settings.
 func (f *Firewall) mergeWith(other Firewall) {
-	f.VPNInputPorts = helpers.MergeUint16Slices(f.VPNInputPorts, other.VPNInputPorts)
-	f.InputPorts = helpers.MergeUint16Slices(f.InputPorts, other.InputPorts)
+	f.VPNInputPorts = helpers.MergeSlices(f.VPNInputPorts, other.VPNInputPorts)
+	f.InputPorts = helpers.MergeSlices(f.InputPorts, other.InputPorts)
 	f.OutboundSubnets = helpers.MergeIPNetsSlices(f.OutboundSubnets, other.OutboundSubnets)
```
qdm12, line 58: Could `MergeIPNetsSlices` be thrown away in favor of `MergeWithSlice`

internal/provider/utils/filtering.go
```diff
@@ -85,25 +86,12 @@ func filterServer(server models.Server,
 	return false
 }
 
-func filterByPossibilities(value string, possibilities []string) (filtered bool) {
+func filterByPossibilities[K string | uint16](value K, possibilities []K) (filtered bool) {
```
qdm12, line 89: I think (don't get mad at me 🙏) it was better before. Using `strings.EqualFold(fmt.Sprint(value), fmt.Sprint(possibility))` is kinda confusing. What we could do however is change `filterByPossibilitiesUint16` to take in a `T comparable` and use `==` instead of `strings.EqualFold` so it could extend to other comparable types. We should keep a separate string focused function which ignores casing.

internal/storage/filter.go
```diff
@@ -106,25 +107,12 @@ func filterServer(server models.Server,
...
-}
-
-// TODO merge with filterByPossibilities with generics in Go 1.18.
-func filterByPossibilitiesUint16(value uint16, possibilities []uint16) (filtered bool) {
-	if len(possibilities) == 0 {
-		return false
-	}
-	for _, possibility := range possibilities {
-		if value == possibility {
+		if strings.EqualFold(fmt.Sprint(value), fmt.Sprint(possibility)) {
```
qdm12, line 115: same comment as in `internal/provider/utils/filtering.go`

qdm12 (approved):

qdm12:
Sorry that took so long, but it's finally rebased/merged 😉 (wanted to use netip.* new types for ip addresses and prefixes which was somehow blocking that generics PR).

PR #1476, opened 2023-03-25
fix(vpnunlimited): lower TLS security level to 3

qdm12:
Fixes #1430

### DIFF
--- a/internal/provider/vpnunlimited/openvpnconf.go
+++ b/internal/provider/vpnunlimited/openvpnconf.go
@@ -2,6 +2,7 @@ package vpnunlimited
 
 import (
 	"github.com/qdm12/gluetun/internal/configuration/settings"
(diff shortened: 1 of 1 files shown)


internal/provider/vpnunlimited/openvpnconf.go
```diff
@@ -18,5 +19,14 @@ func (p *Provider) OpenVPNConfig(connection models.Connection,
 			"route-metric 1",
 		},
 	}
+
+	if settings.Version != openvpn.Openvpn24 {
+		// VPN Unlimited's certificate is sha1WithRSAEncryption and sha1 is now
+		// rejected by openssl 3.x.x which is used by OpenVPN >= 2.5.
+		// We lower the security level to 3 to allow this algorithm,
```
qdm12, line 26:
```suggestion
		// We lower the security level to 0 to allow this algorithm,
```

PR #1543, opened 2023-04-23
feat(protonvpn): port forwarding support with NAT-PMP

### LINKED ISSUE #1488: Feature request: ProtonVPN Port Forwarding
### What's the feature 🧐

The current method of port forwarding appears to be unsupported by gluetun. ProtonVPN has a guide on their site to set up port forwarding: https://protonvpn.com/support/port-forwarding-manual-setup/. This requires the binary `natpmpc` which currently does not exist in the container image.

_No response_

qdm12:
Fixes #1488


internal/provider/protonvpn/portforward.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+
+		if assignedInternalPort != assignedExternalPort {
+			logger.Warn(fmt.Sprintf("internal port assigned %d differs"+
+				" from external port assigned %d",
+				assignedInternalPort, assignedExternalPort))
+		}
+	}
+
+	port = assignedExternalPort
```
nichogx, line 60: Looks like a bug here.

`_, assignedInternalPort, assignedExternalPort, assignedLiftetime, err :=` creates a new `assignedExternalPort` inside the loop, instead of assigning to `var assignedExternalPort uint16` that's outside the loop, so this function will always return zero.

nichogx: fixed in #1649

qdm12: Awesome 💯 !

PR #1585, opened 2023-05-21
fix(firewall): prevent IP family mix in `acceptOutputFromIPToSubnet`

qdm12:


qdm12:
Done in https://github.com/qdm12/gluetun/commit/4ea474b8961bbef21d97cfb31a113c393a2e4657

PR #1596, opened 2023-05-24
Update httpproxy.go to fix error message

filippobuletto:
Error message reading secret file for http proxy password wrongly refer to openvpn instead of http proxy


qdm12 (approved):
Thanks!!

PR #1722, opened 2023-07-03
Expand IPv6 support by checking via pinging too

samip5:
This should help with not trying to use IPv6 when address exists, but no connectivity.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/breml/rootcerts v0.2.11
 	github.com/fatih/color v1.15.0
 	github.com/golang/mock v1.6.0
+	github.com/mehrdadrad/ping v0.2.5
 	github.com/qdm12/dns v1.11.0
 	github.com/qdm12/golibs v0.0.0-20210822203818-5c568b0777b6
 	github.com/qdm12/gosettings v0.3.0-rc13
@@ -19,8 +20,8 @@ require (
 	github.com/stretchr/testify v1.8.4
 	github.com/vishvananda/netlink v1.2.1-beta.2
 	github.com/youmark/pkcs8 v0.0.0-20201027041543-1326539a0a0a
-	golang.org/x/net v0.10.0
-	golang.org/x/sys v0.8.0
+	golang.org/x/net v0.11.0
+	golang.org/x/sys v0.9.0
 	golang.org/x/text v0.10.0
 	golang.zx2c4.com/wireguard v0.0.0-20230325221338-052af4a8072b
 	golang.zx2c4.com/wireguard/wgctrl v0.0.0-20230215201556-9c5414ab4bde
@@ -43,7 +44,7 @@ require (
 	github.com/vishvananda/netns v0.0.0-20200728191858-db3c7e526aae // indirect
 	go4.org/intern v0.0.0-20211027215823-ae77deb06f29 // indirect
 	go4.org/unsafe/assume-no-moving-gc v0.0.0-20230221090011-e4bae7ad2296 // indirect
-	golang.org/x/crypto v0.9.0 // indirect
+	golang.org/x/crypto v0.10.0 // indirect
 	golang.org/x/exp v0.0.0-20230522175609-2e198f4a06a1 // indirect
 	golang.org/x/sync v0.1.0 // indirect
 	golang.zx2c4.com/wintun v0.0.0-20230126152724-0fa3db229ce2 // indirect
--- a/go.sum
+++ b/go.sum
@@ -72,6 +72,8 @@ github.com/mdlayher/netlink v1.6.2/go.mod h1:O1HXX2sIWSMJ3Qn1BYZk1yZM+7iMki/uYGG
 github.com/mdlayher/socket v0.1.1/go.mod h1:mYV5YIZAfHh4dzDVzI8x8tWLWCliuX8Mon5Awbj+qDs=
 github.com/mdlayher/socket v0.2.3 h1:XZA2X2TjdOwNoNPVPclRCURoX/hokBY8nkTmRZFEheM=
 github.com/mdlayher/socket v0.2.3/go.mod h1:bz12/FozYNH/VbvC3q7TRIK/Y6dH1kCKsXaUeXi/FmY=
+github.com/mehrdadrad/ping v0.2.5 h1:q8ABlpxsup6WIRSYFKkCuizJuZ40k7thX1vW49/T3io=
+github.com/mehrdadrad/ping v0.2.5/go.mod h1:14BBOf7wioB6ftnrb9Vv06Zh1sf0Is2u5MZuKrXahis=
 github.com/miekg/dns v1.1.40 h1:pyyPFfGMnciYUk/mXpKkVmeMQjfXqt3FAJ2hy7tPiLA=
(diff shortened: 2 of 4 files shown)


qdm12:
This has multiple problems unfortunately:

- it does networking before the VPN is up which takes some milliseconds. And because ipv6 detection is required for the firewall to be setup, it means it will delay the firewall setup which means traffic from other containers/devices connected to Gluetun would leak data at start. Gluetun was designed to have the firewall setup as fast as possible, and this goes against it.
- this relies on `ipv6.test-ipv6.com` and if this goes down, gluetun becomes unusable. We could make this configurable though.
- There may be cases where the router/ISP doesn't support IPv6 (so ping would fail), but the host supports it (for internal networks etc.), and this doesn't cover such case.
- ping-ing usually requires root (I used that ping library in the past for the healthcheck) which I would like to avoid to eventually one day have a root-less container/program.

The current approach checks your IP routes for IPv6, which is not perfect, but it works fine without networking and is fast.
The better approach (feel free to PR) is to try creating a dummy route or rule (craft it such that it does not impact anything) with the `FamilyV6` family, and check the returned error. This should allow to check if ipv6 is supported or not better, since some hosts might have no ipv6 routes, although they support ipv6.

I'll close this PR, but feel free to have a shot at the 'better approach', I think it should just be around ~50 lines changed in the IsIPv6Supported method.

samip5:
> The current approach checks your IP routes for IPv6, which is not perfect, but it works fine without networking and is fast.
> The better approach (feel free to PR) is to try creating a dummy route or rule (craft it such that it does not impact anything) with the `FamilyV6` family, and check the returned error. This should allow to check if ipv6 is supported or not better, since some hosts might have no ipv6 routes, although they support ipv6.

I don't feel that the dummy route is a good way to check, as even if the host has IPv6, one might still want to disable trying to connect to IPv6 addresses by the VPN eg gluetun container. Currently when there is no IPv6 internet connectivity, the VPN takes ages to start up because it goes though all of the servers and may fail due to "network not reacheable" errors.

I'm open to hearing what would be better to handle such a case.

qdm12:
@samip5 First and foremost, thank you for the PR 💯 
I am very sorry I didn't mention this, I visited tens of issues and email notifications before your PR so I was a bit burnt out on the commenting, my bad 😢

Now  jumping back in the meaty ipv6:

> I don't feel that the dummy route is a good way to check
> I'm open to hearing what would be better to handle such a case.

Thinking more about it, there are two distinct things to detect:

- Does the container support IPv6? **Dummy route approach**, the result would be used for firewall configuration
- Does the container support IPv6 to the outside world? **Check IPv6 default route**, the result would be used for IPv6 VPN servers filtering

If there is no IPv6 default route, I don't believe it's possible to reach an IPv6 address outside your LAN right? And, if so, there is no need to ping an ipv6 address right?

> as even if the host has IPv6 (and default routes for it), one might still want to disable trying to connect to IPv6 addresses by the VPN eg gluetun container.

Why would you not connect over IPv6 if you can use it to connect?

> Currently when there is no IPv6 internet connectivity, the VPN takes ages to start up because it goes though all of the servers and may fail due to "network not reacheable" errors.

For the time being, this can only be the case if you have an ipv6 default route existing so Gluetun assumes IPv6 is supported. Is this your case? You can check with `docker run --rm --cap-add=NET_ADMIN alpine:3.18 ip -6 route`

---

@buroa 

The intention was to redirect to make another PR, since that one was an inappropriate fix. I guess we could have continued and rename things, but I thought it would be clearer. Also I was commenting burnt out as mentioned at the start of this comment.

There isn't a possible redesign really, that is how container networking works. One could delay the start of other connected containers, but that would be too much to ask the user to do. Feel free to create an issue if something else crosses your mind.

> Not following this comment at all...

You can have IPv6 for example in your LAN or Docker network, but your ISP and/or router may not support IPv6.

> It just requires permissions, not root.

This particular library requires root since you need ICMP to ping, pinging over UDP doesn't work well. I used it in the past for the healthcheck as mentioned.

samip5:
> I am very sorry I didn't mention this, I visited tens of issues and email notifications before your PR so I was a bit burnt out on the commenting, my bad

You do hopefully realize that it's not a reason/excuse to behave like you did even if understandable.

> > I don't feel that the dummy route is a good way to check
> > I'm open to hearing what would be better to handle such a case.
> 
> Thinking more about it, there are two distinct things to detect:
> 
> * Does the container support IPv6? **Dummy route approach**, the result would be used for firewall configuration
> * Does the container support IPv6 to the outside world? **Check IPv6 default route**, the result would be used for IPv6 VPN servers filtering
> 
> If there is no IPv6 default route, I don't believe it's possible to reach an IPv6 address outside your LAN right? And, if so, there is no need to ping an ipv6 address right?

Regarding container supporting IPv6, yes the dummy route could be used for that.
Regarding if container supports IPv6 to the outside world, one cannot use just routes to decide that, there needs to be active ingredient that checks it as if route exists but an external firewall prevents gluetun from actively making outbound IPv6 connections, that doesn't work great or poor [happy eyeballs implementation ](https://en.wikipedia.org/wiki/Happy_Eyeballs) by ISP(s) which results in the same so not great right?
If it doesn't have an IPv6 route, yes I think it's safe to assume no support.


> > as even if the host has IPv6 (and default routes for it), one might still want to disable trying to connect to IPv6 addresses by the VPN eg gluetun container.
> 
> Why would you not connect over IPv6 if you can use it to connect?

Currently the detection fails flat if you have an external to gluetun policy that doesn't permit IPv6 traffic thus resulting in poor auto detection result as that's why pinging would be useful.
There should be a way to disable IPv6 entirely even if the autodetection thinks different as there are situations where gluetun is used in quite a glue setup (eg, [pod-gateway](https://github.com/angelnu/pod-gateway) and that doesn't have IPv6 support as VXLAN is tricky on k8s) so end result is trying to connect to IPv6 endpoints without it succeeding as it had been blocked and it taking forever to connect as a result.

qdm12:
> You do hopefully realize that it's not a reason/excuse to behave like you did even if understandable.

It is a good valid reason, can't you simply accept an apology and move forward?

On top of that, this PR is an inappropriate change and that change was not discussed in an issue previously,  closing it with an explanation still makes sense to me today. Anyway moving on;

For the firewall setup, we use the dummy route. Even if there is no ipv6 default route, we have to block ipv6 to prevent leaks if ipv6 is supported internally. This is fast obviously so fits the first firewall enable.

Now to find if ipv6 is reachable to the outside world:
1. Check ipv6 default route; if there is none, no support; if there is go to step 2
2. Query a (configurable) ipv6 address. This should be ideally https (definitely not icmp or udp which is often blocked by routers).
The big problem with step 2 is that no network is allowed except through the vpn (firewall blocked), so that's a bit of a dead end.

I think replacing step 2 with an env variable `VPN_DISABLE_IPV6` would do the trick, if your router blocks ipv6, the user can set it to on to avoid trying connecting over ipv6. Maybe there is a better name for that variable, feel free to suggest.

samip5:
> I think replacing step 2 with an env variable `VPN_DISABLE_IPV6` would do the trick, if your router blocks ipv6, the user can set it to on to avoid trying connecting over ipv6. Maybe there is a better name for that variable, feel free to suggest.

You're open to an PR that would instead of an pinger approach, use that pretty much the same way as I had done in this PR or how should that be implemented?

Issue #1541, opened 2023-04-23
Wiki issue: Update NordVPN to include support for `SERVER_GROUPS`


vdrover:
### URL to the Wiki page

https://github.com/qdm12/gluetun/wiki/NordVPN

### What's the issue?

Server groups are also supported. eg, you can use `SERVER_GROUPS` in docker compose instead of `SERVER_REGIONS`.

The current groups from NordVPN can be found in the CLI. For example:

```
$ sudo nordvpn groups
 > Africa_The_Middle_East_And_India
 > Double_VPN
 > Onion_Over_VPN
 > Standard_VPN_Servers
 > Asia_Pacific
 > Europe
 > P2P
 > The_Americas
```

qdm12:
This data should be added through #1380 first

qdm12:
Although what's the use case? 🤔 

The region group will now be taken into account (`The_Americas`, etc.) and you will be able to use the `SERVER_REGIONS` filter.

But the other ones, do they have any use? That would add a considerable amount of data to the servers data because each server has 3-5 groups, and given the high number of nordvpn servers (they make half of the servers data in gluetun!), that would add a few MBs to the Docker image which I would ideally avoid.

vdrover:
I’m not asking to change the docker image. I’m saying that the current image already does support SERVER_GROUPS. Updating the wiki would save folks fumbling around like I did to see if it works :)

qdm12:
`SERVER_GROUPS` is not even a defined environment variable, so it's simply ignored. Gluetun randomly picks any VPN server from all the servers available.

vdrover:
Understood. 

Then to answer your use-case question, I was hoping to use random double_vpn servers.

qdm12:
Ok cool, I created #1643 (which includes your request), please subscribe to that one instead 👍 Thanks!

PR #1806, opened 2023-08-18
Category filtering for NordVPN

### LINKED ISSUE #1643: Feature request: Filter NordVPN servers by group
### What's the feature 🧐

NordVPN servers have groups, it would be nice to be able to filter by group.

- [ ] Change `internal/providers/nordvpn/updater` code
- [ ] Update `internal/storage/servers.json`
- [ ] Update Github wiki NorVPN servers list

Blocked by #1380 

Requested in:

- https://github.com/qdm12/gluetun/issues/719#issuecomment-1258403893
- https://github.com/qdm12/gluetun/issues/1541

AdamHebby:
Aims to fix #1643 

I am _not_ a GO developer, please read this code carefully!

image

image

After talking to NordVPN support, as long as a server supports P2P, a P2P connection will be established once it's determined there is a need for it (IE seeing P2P traffic). There's no way of knowing if it has connected to the P2P server other than contacting support with the IP address or results from `curl https://ipleak.net/json/`. I did ask if there was a way of starting the connection and specifying P2P from the get-go to avoid any switching but they said there was no documentation on this.

I've also [seen online](https://www.reddit.com/r/nordvpn/comments/q6astb/comment/hgbawxq/) that this switch from standard to P2P can result in you being reconnected to a "better" server, but this could also introduce latency.

TL;DR Nord claims it "just works"

Also unsure on testing? I've added a single test case for the categories - please let me know if there's anything more I can do here.

### DIFF
--- a/internal/provider/nordvpn/updater/models.go
+++ b/internal/provider/nordvpn/updater/models.go
@@ -139,6 +139,19 @@ func (s *serverData) hasVPNService(services map[uint32]serviceData) (ok bool) {
 	return false
 }
 
+// categories returns the list of categories for the server.
+func (s *serverData) categories(groups map[uint32]groupData) (categories []string) {
+	categories = make([]string, 0, len(s.GroupIDs))
+	for _, groupID := range s.GroupIDs {
+		data, ok := groups[groupID]
+		if !ok || data.Type.Identifier == "regions" {
+			continue
+		}
+		categories = append(categories, data.Title)
+	}
+	return categories
+}
+
 // ips returns the list of IP addresses for the server.
 func (s *serverData) ips() (ips []netip.Addr) {
 	ips = make([]netip.Addr, 0, len(s.IPs))
--- a/internal/provider/utils/filtering.go
+++ b/internal/provider/utils/filtering.go
@@ -61,6 +61,10 @@ func filterServer(server models.Server,
 		return true
 	}
 
+	if filterAnyByPossibilities(server.Categories, selection.Categories) {
+		return true
+	}
+
 	if filterByPossibilities(server.Region, selection.Regions) {
 		return true
 	}
@@ -99,5 +103,19 @@ func filterByPossibilities[T string | uint16](value T, possibilities []T) (filte
 			return false
 		}
 	}
+	return true
+}
+
+func filterAnyByPossibilities(values, possibilities []string) (filtered bool) {
+	if len(possibilities) == 0 {
+		return false
+	}
+
+	for _, value := range values {
+		if !filterByPossibilities(value, possibilities) {
+			return false // found a valid value
+		}
+	}
+
 	return true
 }
--- a/internal/storage/filter.go
+++ b/internal/storage/filter.go
@@ -82,6 +82,10 @@ func filterServer(server models.Server,
 		return true
 	}
 
+	if filterAnyByPossibilities(server.Categories, selection.Categories) {
+		return true
+	}
+
 	if filterByPossibilities(server.Region, selection.Regions) {
 		return true
 	}
@@ -120,6 +124,20 @@ func filterByPossibilities[T string | uint16](value T, possibilities []T) (filte
 			return false
 		}
 	}
+	return true
+}
+
+func filterAnyByPossibilities(values, possibilities []string) (filtered bool) {
+	if len(possibilities) == 0 {
+		return false
+	}
+
+	for _, value := range values {
+		if !filterByPossibilities(value, possibilities) {
+			return false // found a valid value
+		}
+	}
+
 	return true
 }
 
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -10,6 +10,7 @@ issues:
         - goerr113
         - containedctx
         - goconst
+        - maintidx
     - path: "internal\\/server\\/.+\\.go"
       linters:
         - dupl
--- a/Dockerfile
+++ b/Dockerfile
@@ -109,6 +109,7 @@ ENV VPN_SERVICE_PROVIDER=pia \
     SERVER_COUNTRIES= \
     SERVER_CITIES= \
     SERVER_HOSTNAMES= \
+    SERVER_CATEGORIES= \
     # # Mullvad only:
     ISP= \
     OWNED_ONLY=no \
--- a/internal/configuration/settings/errors.go
+++ b/internal/configuration/settings/errors.go
@@ -5,6 +5,7 @@ import "errors"
 var (
 	ErrCityNotValid                    = errors.New("the city specified is not valid")
 	ErrControlServerPrivilegedPort     = errors.New("cannot use privileged port without running as root")
+	ErrCategoryNotValid                = errors.New("the category specified is not valid")
 	ErrCountryNotValid                 = errors.New("the country specified is not valid")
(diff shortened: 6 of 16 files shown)


AdamHebby:
@qdm12 Can I get your review on this please? Also I saw you were concerned about the potential file size increase, it's rather minimal at .9MB increase.

internal/provider/nordvpn/updater/models.go
```diff
@@ -83,6 +85,18 @@ func (s *serverData) region() (region string) {
 	return ""
 }
 
+// categories returns the list of categories for the server.
+func (s *serverData) categories() (categories []string) {
+	cats := make([]string, 0, len(s.Groups))
+
```
qdm12, line 91: nit just for the sake of using the returned variable and not create another 😉 

```go
	categories = make([]string, 0, len(s.Groups))
```

You would have to also update `cats` -> `categories`

internal/provider/nordvpn/updater/models.go
```diff
@@ -83,6 +85,18 @@ func (s *serverData) region() (region string) {
...
+	for _, group := range s.Groups {
+		if group.Type.Identifier == "legacy_group_category" {
+			cats = append(cats, group.Identifier)
+		}
```
qdm12, line 95: Quick question: why do we only take legacy group categories? Is there a new group category we can use instead?

internal/provider/nordvpn/updater/models.go
```diff
@@ -47,6 +47,8 @@ type serverData struct {
 		} `json:"metadata"`
 	} `json:"technologies"`
 	Groups []struct {
+		// Identifier can be 'legacy_standard', 'legacy_p2p'.
+		Identifier string `json:"identifier"`
```
qdm12, line 51: If the group can only be `legacy_standard` or `legacy_p2p`, would it be reasonable instead to just set the `PortForward` server boolean when encountering `legacy_p2p` and not have categories at all?

internal/provider/utils/filtering.go
```diff
@@ -97,3 +101,19 @@ func filterByPossibilities[T string | uint16](value T, possibilities []T) (filte
 	}
 	return true
 }
+
+func filterMultipleByPossibilities[T string | uint16](values []T, possibilities []T) (filtered bool) {
```
qdm12, line 105: nit since this is only used for []string (categories), no need to have generic for now:

```suggestion
func filterMultipleByPossibilities(values, possibilities []string) (filtered bool) {
```

internal/provider/utils/filtering.go
```diff
@@ -97,3 +101,19 @@ func filterByPossibilities[T string | uint16](value T, possibilities []T) (filte
...
+
+	pass := true
+	for _, value := range values {
+		if !filterByPossibilities(value, possibilities) {
+			pass = false
+			break
+		}
+	}
+
+	return pass
```
qdm12, line 118: Correct me if I'm wrong, but I think this is meant to be:

```suggestion
	for _, value := range values {
		filtered = filterByPossibilities(value, possibilities)
		if filtered {
			return true
		}
	}

	return false
```

internal/storage/filter.go
```diff
@@ -119,6 +123,22 @@ func filterByPossibilities[T string | uint16](value T, possibilities []T) (filte
...
+	for _, value := range values {
+		if !filterByPossibilities(value, possibilities) {
+			pass = false
+			break
+		}
+	}
+
+	return pass
+}
+
```
qdm12, line 141: same comment as for the other filterMultipleByPossibilities definition 😉

qdm12 (commented):
Nice work 👍 💯 

I left a few comments, the most important one being:

> If the group can only be `legacy_standard` or `legacy_p2p`, would it be reasonable instead to just set the `PortForward` server model boolean to true when encountering `legacy_p2p` and not have categories at all?

That of course removes a lot of the fun in this PR, but it does make sense so far to me; even if there are other categories, I'm not certain there would be a usage to filter by category except for p2p? 🤔

internal/provider/nordvpn/updater/models.go
AdamHebby: I'm not sure what you mean here? Are you saying to rename cats to categories?
image

If I were to then remove the `make` it would add dupes to categories. Or are you saying to instead remove from categories and invert the IF logic?

internal/provider/nordvpn/updater/models.go
AdamHebby: Just double checked and turns out I missed `legacy_dedicated_ip`
```
"legacy_dedicated_ip"
"legacy_p2p"
"legacy_standard"
```

NGL I'm not sure what you mean about the PortForward bit, what relation to this does it have?

Side note: NordVPN does not support port forwarding

internal/provider/utils/filtering.go
AdamHebby: Should I also update `filterByPossibilities` since that's what I copied?

internal/provider/utils/filtering.go
AdamHebby: image

It's kinda based on this, but from my understanding, a TRUE evaluation means to filter out the value? Pls read filtering.go since you'll understand it more than me and advise pls

internal/provider/nordvpn/updater/models.go
qdm12: Yes my bad I may had mixed with another provider.
Another question: wasn't the server groups meant to be for example `Double_VPN`, `Africa_The_Middle_East_And_India`, etc. as mentioned in https://github.com/qdm12/gluetun/issues/1541#issue-1680084182 ?

internal/provider/nordvpn/updater/models.go
qdm12: No just have it as:

```go
func (s *serverData) categories() (categories []string) {
	categories = make([]string, 0, len(s.Groups))
```

and that will work 😉 `:=` is used to declare new variables, but `categories` is already there as the return variable.
If you use `make([]string, 0, len(s.Groups))`, the capacity of the slice is len(s.Groups), but the length is 0, so appending will start at index 0, then 1 etc.

internal/provider/nordvpn/updater/models.go
qdm12: Ok! Should we use the `title` instead since it's a bit more user friendly?

qdm12 (commented):
Sorry for the delay re-reviewing, here are some answers on previous conversations! 👍

internal/provider/utils/filtering.go
qdm12: You're correct in your understanding of `filterByPossibilities`, but your function returns "do not filter" as soon as any value from `values` is correct; we want to check all values are correct. I'm fixing this 😉

internal/provider/nordvpn/updater/models.go
qdm12: Using the title, so now it's one of: Dedicated IP, Double VPN, Onion Over VPN, P2P, Standard VPN servers

qdm12:
Actually never mind, checking more the response from https://api.nordvpn.com/v2/servers?limit=10

it shows `Legacy category` for anything not `regions`, so let's keep it with categories.

PR #1836, opened 2023-09-01
Implement support for Wireguard in PIA

### LINKED ISSUE #612: Feature request: Wireguard support for PIA
**What's the feature?** 🧐

- Support Wireguard for PIA

**Optional extra information** 🚀

- [pia-foss/manual-connections](https://github.com/pia-foss/manual-connections/)
- [`connect_to_wireguard_with_token.sh`](https://github.com/pia-foss/manual-connections/blob/master/connect_to_wireguard_with_token.sh)
- [`get_token.sh`](https://github.com/pia-foss/manual-connections/blob/e7d4806a9eabc89bc5da17c9e0e2af96bbdf526b/get_token.sh#L75) the token is generated with a call to https://privateinternetaccess.com/gtoken/generateToken using the `PIA_USERNAME` and `PIA_PASSWORD`. (thanks @marciocm)
- Hostname and IP can be found at https://serverlist.piaservers.net/vpninfo/servers/v5

## TODOs

- [ ] make PIA API package from port forwarding code since Wireguard also requires the token etc.
- [ ] Obtain dynamically Wireguard public key (and other data) and plug that in connection builder
- [ ] Add PIA to `VPNSP` for Wireguard

FireMasterK:
Closes #612
Untested, as I don't have a PIA account

Warning: I don't really write a lot of Go code, so the structure may be bad!


garlik82:
I'm willing to test this. Already built the image but what should I put in my compose file?

FireMasterK:
> I'm willing to test this. Already built the image but what should I put in my compose file?

You should be able to set your account credentials in the openvpn user credentials, and set `VPN_TYPE=wireguard`. You need not configure any wireguard keys, as ephemeral keys are automatically registered and used in the API.

garlik82:
> > I'm willing to test this. Already built the image but what should I put in my compose file?
> 
> You should be able to set your account credentials in the openvpn user credentials, and set `VPN_TYPE=wireguard`. You need not configure any wireguard keys, as ephemeral keys are automatically registered and used in the API.

ERROR VPN settings: provider settings: VPN provider name is not valid for Wireguard: value is not one of the possible choices: private internet access must be one of airvpn, custom, ivpn, mullvad, nordvpn, surfshark or windscrib.

This is what I get

FireMasterK:
> ERROR VPN settings: provider settings: VPN provider name is not valid for Wireguard: value is not one of the possible choices: private internet access must be one of airvpn, custom, ivpn, mullvad, nordvpn, surfshark or windscrib.
> 
> This is what I get.
> 
> If I put custom instead of PIA I get the error saying no endpoint set

I've fixed this in the latest commit, please build and try it again.

garlik82:
> > ERROR VPN settings: provider settings: VPN provider name is not valid for Wireguard: value is not one of the possible choices: private internet access must be one of airvpn, custom, ivpn, mullvad, nordvpn, surfshark or windscrib.
> > This is what I get.
> > If I put custom instead of PIA I get the error saying no endpoint set
> 
> I've fixed this in the latest commit, please build and try it again.

Now I get these: 

ERROR [vpn] finding a VPN server: filtering servers: no server found: for VPN wireguard; protocol udp; region dk copenhagen; encryption preset strong
--


<br class="Apple-interchange-newline">

FireMasterK:
> > I've fixed this in the latest commit, please build and try it again.
> 
> Now I get this:
> 
> ## ERROR [vpn] finding a VPN server: filtering servers: no server found: for VPN wireguard; protocol udp; region dk copenhagen; encryption preset strong

Hi, please try again, this should now be fixed!

garlik82:
How do I know the right thing to put in SERVER_REGIONS= ?

FireMasterK:
> How do I know the right thing to put in SERVER_REGIONS= ?

Anything that works with OpenVPN should most likely work with Wireguard too.

garlik82:
> > How do I know the right thing to put in SERVER_REGIONS= ?
> 
> Anything that works with OpenVPN should most likely work with Wireguard too.
 
It doesn't. Must likely it's searching on servers.json for a wireguard server but PIA only has ovpn server in the built in server list.

FireMasterK:
> It doesn't. Must likely it's searching on servers.json for a wireguard server but PIA only has ovpn server in the built in server list.
> 
> https://raw.githubusercontent.com/qdm12/gluetun/master/internal/storage/servers.json (builtin server list) https://serverlist.piaservers.net/vpninfo/servers/v6 (pia server list)

Could you give it a go again? I managed to reproduce those errors without an account, and each of them should be fixed now.

You may need to disable the firewall with `FIREWALL=off` (not sure how to work around it currently)

garlik82:
almost there.

RROR [vpn] getting Wireguard connection: Post "https://www.privateinternetaccess.com/gtoken/generateToken": dial tcp: lookup www.privateinternetaccess.com on 1.1.1.1:53: write udp 172.20.0.13:35675->1.1.1.1:53: write: operation not permitted

internal/provider/privateinternetaccess/wireguardconfigurator.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	port, err := strconv.ParseUint(responseData.ServerPort, 10, 16)
+	if err != nil {
+		return settings, err
+	}
+	connection.Port = uint16(port)
+
+	psk := publicKey.String()
+	wireguardSettings.PreSharedKey = &psk
```
qdm12, line 96: Are they using the public key as the pre-shared key!?? That is so silly if that's the case.

internal/provider/utils/noportforward.go
```diff
@@ -37,3 +40,7 @@ func (n *NoPortForwarding) KeepPortForward(context.Context, uint16, netip.Addr,
 	string, Logger) (err error) {
 	return fmt.Errorf("%w: for %s", ErrPortForwardingNotSupported, n.providerName)
 }
+
+type NoWireguardConfigurator interface {
+	GetWireguardConnection(ctx context.Context, connection models.Connection, wireguardSettings settings.Wireguard, ipv6Supported bool) (settings wireguard.Settings, err error)
+}
```
qdm12, line 46: You should implement it like NoPortForwarding instead of just having an interface

internal/updater/providers.go
```diff
@@ -22,12 +22,16 @@ func (u *Updater) updateProvider(ctx context.Context, provider Provider,
 	existingServersCount := u.storage.GetServersCount(providerName)
 	minServers := int(minRatio * float64(existingServersCount))
 	servers, err := provider.FetchServers(ctx, minServers)
+	providerWireguardKeyUnknown := false
+	if providerName == "private internet access" {
```
qdm12, line 26: use the constant `providers.PrivateInternetAccess` instead of writing a string directly

internal/vpn/run.go
```diff
@@ -40,6 +41,17 @@ func (l *Loop) Run(ctx context.Context, done chan<- struct{}) {
 			vpnRunner, serverName, err = setupOpenVPN(ctx, l.fw,
 				l.openvpnConf, providerConf, settings, l.ipv6Supported, l.starter, subLogger)
 		} else { // Wireguard
+			// save auth in case of PIA for port forwarding and getting the wg config
+			if *settings.Provider.Name == "private internet access" {
```
qdm12, line 45: use the constant `providers.PrivateInternetAccess` instead of writing a string directly

internal/vpn/run.go
```diff
@@ -40,6 +41,17 @@ func (l *Loop) Run(ctx context.Context, done chan<- struct{}) {
...
+				if *settings.OpenVPN.User != "" {
+					err := l.openvpnConf.WriteAuthFile(*settings.OpenVPN.User, *settings.OpenVPN.Password)
+					if err != nil {
+						l.crashed(ctx, fmt.Errorf("writing auth to file: %w", err))
+						continue
+					}
+				}
```
qdm12, line 52: we should inject those credentials directly to the code fetching the wireguard config, instead of using that openvpn auth file.

internal/vpn/wireguard.go
```diff
@@ -22,7 +21,19 @@ func setupWireguard(ctx context.Context, netlinker NetLinker,
...
+	wireguardConfigurator, ok := providerConf.(provider.WireguardConfigurator)
+	if ok {
+		// Use already defined variables
+		wireguardSettings, err = wireguardConfigurator.GetWireguardConnection(ctx, connection, settings.Wireguard, ipv6Supported)
+		if err != nil {
+			return nil, "", fmt.Errorf("getting Wireguard connection: %w", err)
+		}
+	} else {
+		wireguardSettings = utils.BuildWireguardSettings(connection, settings.Wireguard, ipv6Supported)
+	}
```
qdm12, line 36: It would be better to inject `ctx` + `providerConf` into `utils.BuildWireguardSettings` and modify `buildWireguardSettings` to check if providerConf has the `GetWireguardConfig` method.

Also, if you do this with a type assertion as you do now, it will pass for all providers since each have the `GetWireguardConfig` method from the `NoWireguardXXX` interface; I would suggest to simply remove the `NoWireguardXXX` interface you defined, and that would nicely solve it.

internal/vpn/wireguard.go
```diff
@@ -3,7 +3,6 @@ package vpn
 import (
 	"context"
 	"fmt"
-
```
qdm12, line 6: nit unneeded change, revert (re-add empty line)

internal/models/server.go
```diff
@@ -44,7 +44,7 @@ var (
 	ErrWireguardPublicKeyEmpty = errors.New("wireguard public key field is empty")
 )
 
-func (s *Server) HasMinimumInformation() (err error) {
+func (s *Server) HasMinimumInformation(providerWireguardKeyUnknown bool) (err error) {
```
qdm12, line 47: instead check the `Hostname` field and determine if it's a PIA server from this? So there is no need for the extra argument 🤔

.gitignore
```diff
@@ -1 +1,2 @@
 scratch.txt
+.idea/
```
qdm12, line 2: remove before merging

internal/provider/privateinternetaccess/tokenutils.go
```diff
@@ -0,0 +1,72 @@
...
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/url"
+	"strings"
+)
+
+func fetchToken(ctx context.Context, client *http.Client,
+	tokenType string, authFilePath string) (token string, err error) {
```
qdm12, line 13: since this is used outside openvpn-only, we should inject the username and password and not use the openvpn auth file anymore (especially if it's for Wireguard, that makes it super strange)

internal/provider/privateinternetaccess/tokenutils.go
```diff
@@ -0,0 +1,72 @@
...
+
+	var path string
+
+	switch tokenType {
+	case "client":
+		path = "/api/client/v2/token"
+	case "gtoken":
+		path = "/gtoken/generateToken"
+	default:
+		return "", fmt.Errorf("token type %q is not supported", tokenType)
```
qdm12, line 32: you could even panic here, since that would be a programming error:

```suggestion
		panic(fmt.Sprintf("token type %q is not supported", tokenType))
```

internal/provider/privateinternetaccess/wireguardconfigurator.go
```diff
@@ -0,0 +1,104 @@
...
+	client, err := newHTTPClient(connection.Hostname)
+
+	// generate a new private key
+	privateKey, err := wgtypes.GeneratePrivateKey()
+
+	// generate public key from private key
+	publicKey := privateKey.PublicKey()
+
+	// fetch token from PIA's API
+	token, err := fetchToken(ctx, client, "gtoken", p.authFilePath)
```
qdm12, line 30: ⚠️ This calls www.privateinternetaccess.com and that's a deal breaker for the killswitch and to avoid leaking data outside the tunnel (especially from other containers/devices connecting through gluetun). For example if the tunnel takes 5 seconds to setup, data will leak for 5 seconds; if the tunnel never manages to setup, it would leak for hours.
That's also why I did not implement this yet, but don't worry, your code will be used at some point.

I'm working on a few other moving parts, and the idea is that connecting to privateinternetaccess.com over https (and its dns resolution) would be the only one allowed outside the vpn, temporarily. But this is far from done, so expect this pull request to stay opened for a few months unfortunately.

qdm12 (commented):
Thanks for the contribution!

Unfortunately this requires connectivity before the VPN is up, and that's a deal breaker for Gluetun. I'm slowly working towards a solution but this is far from done. The PR won't be useless though, it will be used at some point in the future 👍 !

Nikamura:
any way I could help to get support added for PIA wireguard?

internal/provider/privateinternetaccess/wireguardconfigurator.go
bryanvaz: Theoretically, if the agent were to connect via openvpn via a sideband, download the necessary wireguard configs over the ovpn tunnel, then bring up the primary wireguard tunnel, would that address the leakage risk?

Obviously the scenario assumes all tunnelled traffic is only routed over wireguard, and thus is blocked till the wireguard tunnel is up. I would assume this should be the default behaviour anyways in the event of the PIA dropping the wireguard tunnel to rotate IPs.

PR #1874, opened 2023-09-20
fix(portforward): rework loop entirely

qdm12:


buthed010203:
I get this error on startup when using this branch:
image

internal/portforward/loop.go
```diff
@@ -1,64 +1,139 @@
...
+	}
+}
+
+func (l *Loop) Stop() (err error) {
+	l.runCancel()
+	<-l.runDone
+
+	if l.service != nil {
+		return l.service.Stop()
+	}
```
qdm12, line 130: this should be moved to execute in the `run` function, in case Stop is called when Start is running in the `run` function, which would cause a Stop on an already start-aborted service.

internal/portforward/service/interfaces.go
```diff
@@ -0,0 +1,16 @@
...
+
+type PortAllower interface {
+	SetAllowedPort(ctx context.Context, port uint16, intf string) (err error)
+	RemoveAllowedPort(ctx context.Context, port uint16) (err error)
+}
+
+type Logger interface {
+	Info(s string)
+	Warn(s string)
+	Error(s string)
```
qdm12, line 15: Do we need each of these?

internal/portforward/service/service.go
```diff
@@ -0,0 +1,44 @@
...
+		// Fixed injected objets
+		client:      client,
+		portAllower: portAllower,
+		logger:      logger,
+	}
+}
+
+func (s *Service) GetPortForwarded() (port uint16) {
+	s.startStopMutex.Lock()
+	defer s.startStopMutex.Unlock()
```
qdm12, line 42: Add a port rw mutex for the sake of clarity.
Eventhough using startStopMutex works in this case, especially since Start and Stop should be quite fast to execute, it's better to separate them.

internal/portforward/service/settings.go
```diff
@@ -0,0 +1,59 @@
...
+	"fmt"
+	"net/netip"
+
+	"github.com/qdm12/gluetun/internal/configuration/settings"
+	"github.com/qdm12/gluetun/internal/provider"
+	"github.com/qdm12/gosettings"
+)
+
+type Settings struct {
+	Settings      settings.PortForwarding
```
qdm12, line 14: Rename to `UserSettings`?

internal/portforward/service/settings.go
```diff
@@ -0,0 +1,59 @@
...
+	s.Gateway = gosettings.OverrideWithValidator(s.Gateway, update.Gateway)
+	s.ServerName = gosettings.OverrideWithString(s.ServerName, update.ServerName)
+	s.Interface = gosettings.OverrideWithString(s.Interface, update.Interface)
+	s.VPNProvider = gosettings.OverrideWithString(s.VPNProvider, update.VPNProvider)
+}
+
+var ErrVPNProviderNotSet = errors.New("VPN provider not set")
+
+func (s *Settings) validate() (err error) {
+	if s.VPNProvider == "" {
```
qdm12, line 55: check for non-nil port forwarder, valid gateway, non empty servername (pia only), non empty interface and non empty vpn provider

internal/portforward/service/settings.go
```diff
@@ -0,0 +1,59 @@
...
+type Settings struct {
+	Settings      settings.PortForwarding
+	PortForwarder provider.PortForwarder
+	Gateway       netip.Addr // needed for PIA and ProtonVPN
+	ServerName    string     // needed for PIA
+	Interface     string     // needed for PIA and ProtonVPN, tun0 for example
+	VPNProvider   string     // used to validate new settings
+}
+
+func (s *Settings) UpdateWith(partialUpdate Settings) (err error) {
```
qdm12, line 22: Add comment it deep copies the settings and leaves original settings untouched

internal/portforward/loop.go
qdm12: Keeping this as is, so we access the service stop error;
calling Stop on an start-aborted service will simply return `nil` which is fine in this case.

PR #1890, opened 2023-10-02
support for AirVPN entry-IP 3 with Wireguard protocol

le0m:
Add support for entry-IP 3 in AirVPN's Wireguard protocol. See [specs](https://airvpn.org/specs/), i incurred in this problem today.

AirVPN also supports Wireguard's standard port `51820` with entry-IP 3, but this is my very first time writing GO code so i don't know how to conditionally add it to the validation [here](https://github.com/qdm12/gluetun/blob/d4df87286e1e14c5471d09800ad8408285b44e58/internal/configuration/settings/wireguardselection.go#L73).

Tested this locally with the updater CLI command, seems to work.

### DIFF
--- a/internal/provider/airvpn/updater/servers.go
+++ b/internal/provider/airvpn/updater/servers.go
@@ -20,14 +20,14 @@ func (u *Updater) FetchServers(ctx context.Context, minServers int) (
 	}
 
 	// every API server model has:
-	// - Wireguard server using IPv4In1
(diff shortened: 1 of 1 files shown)


xEsurient:
Holy this pull is old, but I wonder how the prioritization works in Gluetun, I can see on the server list that for OpenVPN this is already done.

PR #1992, opened 2023-12-03
Update purevpn's zip file url based on

yifangd:
https://support.purevpn.com/purevpn/openvpn-files


qdm12:
Awesome, I updated the commit to carry updated servers data as well, thanks!!

Issue #1967, opened 2023-11-16
Bug: `tcp-client` OpenVPN setting not recognized


ralienpp:
### Is this urgent?

No

### Host OS

Linux Mint 21.2

### CPU arch

x86_64

### VPN service provider

Custom

### What are you using to run the container

docker-compose

### What is the version of Gluetun

2023-11-08T10:11:52.918Z (commit 8318be3)

### What's the problem 🤔

I cannot configure `gluetun` to connect to a custom OpenVPN server that uses TCP for transport. However, I can connect to it directly, if I use the OpenVPN client of my host system.

What I think is wrong: the logic of `ExtractProto` is unaware of `tcp-client` as an option, there's no condition for this in the switch: https://github.com/qdm12/gluetun/blob/674f0a32095c306edda07862b0015a7217c94bb8/internal/updater/openvpn/extract.go#L28

If I change my config to say `tcp`, the container starts, but according to the log it incorrectly identifies the protocol as UDP: `├── Protocol: UDP` at startup. And yet, if you go to the very end it you see that the client tries to open a TCP connection: `TCP connection established with [AF_INET]11.12.13.14:1194`

This line `OpenSSL: error:0A000410:SSL routines::sslv3 alert handshake failure` makes me think that perhaps the error is caused by a failed attempt to agree on a cipher with the server? If that were the case, how to confirm it?

### Share your logs (at least 10 lines)

```plain text
gluetun  | Running version latest built on 2023-11-08T10:11:52.918Z (commit 8318be3)
gluetun  | 2023-11-16T20:27:38Z INFO [routing] default route found: interface eth0, gateway 172.19.0.1, assigned IP 172.19.0.2 and family v4
gluetun  | 2023-11-16T20:27:38Z INFO [routing] local ethernet link found: eth0
gluetun  | 2023-11-16T20:27:38Z INFO [routing] local ipnet found: 172.19.0.0/16
gluetun  | 2023-11-16T20:27:38Z INFO [firewall] enabling...
gluetun  | 2023-11-16T20:27:38Z INFO [firewall] enabled successfully
gluetun  | 2023-11-16T20:27:39Z INFO [storage] merging by most recent 17689 hardcoded servers and 17689 servers read from /gluetun/servers.json
[...]

PR #1995, opened 2023-12-07
Tweak handling of "tcp-client" config value

ralienpp:
Make it the same case as "tcp", so the protocol name can be used without changes in iptables commands.

This deals with the side-effect mentioned in https://github.com/qdm12/gluetun/issues/1967.

### DIFF
--- a/internal/openvpn/extract/extract.go
+++ b/internal/openvpn/extract/extract.go
@@ -81,7 +81,11 @@ func extractProto(line string) (protocol string, err error) {
 	}
 
 	switch fields[1] {
-	case "tcp", "tcp4", "tcp6", "tcp-client", "udp", "udp4", "udp6":
+	case "tcp-client":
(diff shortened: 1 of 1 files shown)


qdm12:
1. Sorry for the late reply, I don't monitor closed issues (#1967) so I didn't see the problem. Next time feel free to open a new issue 😉 
2. Thank you so much for opening a PR fixing the problem!
3. I decided to push 75fd86962542eab693d7698c0e9e731b2f391bd1 which makes the firewall convert `tcp-client` -> `tcp`, instead of changing the parsed value, since I'm not 100% sure `tcp-client` and `tcp` are the same for OpenVPN (and documentation/conversations on this are not very available), so I would rather keep the openvpn config `proto tcp-client` line intact and only change where the protocol name for the firewall (only one place).

Cheers!

PR #1742, opened 2023-07-11
feat(dns): use qdm12/dns@v2.0.0-beta-rc6

qdm12:
- Replace Unbound subprocess with a Go built DNS over TLS forwarder
- DNSSEC validation is dropped for now (it's sort of unneeded)
- Faster start up
- Clearer error messages
- Allow for more Gluetun-specific customization
- Fix #137

Issue #137, opened 2020-04-12
Feature request: DNS over HTTPS in Go to replace Unbound


qdm12:
We should adopt https://github.com/qdm12/dns/releases/tag/v2.0.0-rc5 even if it has no DNSSEC validation builtin yet. Main reason being most domains are not DNSSEC secured, and so your DNS provider (cloudflare, google etc.) can lie about the records for these unsecured domains, even if you can securely proof these are unsecured. So not really a point so much, although this will be done (see https://github.com/qdm12/dns/pull/97)

This is blocked by an upgrade of the qdm12/gosettings library to v0.4.0-rc9

denizdogan:
I just wanted to check in on the progress here! Are we there yet? 😄

qdm12:
It's still a work in progress at https://github.com/qdm12/dns/tree/v2.0.0-beta I can't give an exact timeline, but it's almost done.

jcheroske:
It would be great if a specific DoT server could be specified, and not just providers. Not sure if this is doable now, but I haven't been able to find it.

frepke:
> It's still a work in progress at https://github.com/qdm12/dns/tree/v2.0.0-beta I can't give an exact timeline, but it's almost done.

Hi Quentin, any information available?
And can we expect an implementation in Gluetun when it's ready?

qdm12:
Can anyone try image `qmcgaw/gluetun:pr-1742`? It uses DNS over TLS in Go which I wrote with my own lil' fingers (arguably better than dns over https). Options should all be the same as before. The main difference is it doesn't do DNSSEC validation, which, after a lot of digging, turns out to be not that useful. I'll work on it though, but it doesn't feel like a blocking priority to drop Unbound.

This will allow a lot more options and customizations (even Prometheus metrics at some point, it's already in the dns server Go code).

pduchnovsky:
> Can anyone try image `qmcgaw/gluetun:pr-1742`? It uses DNS over TLS in Go which I wrote with my own lil' fingers (arguably better than dns over https). Options should all be the same as before. The main difference is it doesn't do DNSSEC validation, which, after a lot of digging, turns out to be not that useful. I'll work on it though, but it doesn't feel like a blocking priority to drop Unbound.
> 
> This will allow a lot more options and customizations (even Prometheus metrics at some point, it's already in the dns server Go code).

What needs to be tested exactly ?
Also, how safe is it to test this version in 'production' at this stage ? :D

[...]

PR #1998, opened 2023-12-09
Support domain endpoints [WIP]

ShrirajHegde:
Support domain as endpoint in both config files and environment.

### DIFF
--- a/internal/configuration/sources/files/wireguardselection.go
+++ b/internal/configuration/sources/files/wireguardselection.go
@@ -64,7 +64,11 @@ func parseWireguardPeerSection(peerSection *ini.Section,
 
 		ip, err := netip.ParseAddr(host)
 		if err != nil {
-			return fmt.Errorf("%w: %w", ErrEndpointHostNotIP, err)
+			ips, err := net.LookupIP(host)
+			if err != nil {
+				return fmt.Errorf("resolving domain: %w", err)
+			}
+			ip, _ = netip.ParseAddr(ips[0].String())
 		}
 
 		endpointPort, err := port.Validate(portString)


ShrirajHegde:
Domain resolution only works for `/gluetun/wireguard/wg0.conf` file for now. 

I can't find the code that generates the error : `ERROR reading from environment variables: VPN provider: server selection: environment variable VPN_ENDPOINT_IP: ParseAddr("example.com"): unexpected character (at "example.com")`

qdm12:
Unfortunately this is a lot more complicated to do than this, and I'm slowly working towards a solution.

The reason why no hostname is allowed is https://github.com/qdm12/gluetun-wiki/blob/main/faq/others.md#server-information

> Gluetun uses IP addresses instead of hostnames to connect to VPN servers, to avoid doing a DNS resolution at start.
> The main reason is Gluetun should not have connectivity before it establishes the VPN connection, such that connected containers (or other machines) won't leak their data out for the few starting seconds.

Essentially the solution is to optionally allow DNS traffic for very specific requests (matching a regex), and this will be done after #1742 is merged - still a few months of work really. It does require some firewall fiddling too, as well as some refactoring of the async logic within gluetun (already in progress, done half way).

I'll close this PR for now since it cannot resolve it, and resolving it is rather complicated and a-job-for-me-since-i'm-already-working-on-it (right now I'm finishing DNSSEC support in https://github.com/qdm12/dns/tree/v2.0.0-beta so it can fully replace unbound for private DNS in Gluetun)😸

qdm12:
Also about:

> Domain resolution only works for /gluetun/wireguard/wg0.conf file for now.

Yes and it should not. Firewall gets enabled at

https://github.com/qdm12/gluetun/blob/657b4b787f247d6f6c5180ca4766b2c97dba6ecf/cmd/gluetun/main.go#L215C2-L215C14

And all settings are read before it gets enabled at

https://github.com/qdm12/gluetun/blob/657b4b787f247d6f6c5180ca4766b2c97dba6ecf/cmd/gluetun/main.go#L183

Reading settings should be blazing fast and without external network, since we need the settings to configure the firewall as well. If we do network IO like DNS resolution when reading the settings, it means the firewall won't be enabled for a few more milliseconds which is not what we want. I left the firewall enabling after the settings reading, but within this PR it should be before then (creating another mess of firewall configuration too).

ShrirajHegde:
>Essentially the solution is to optionally allow DNS traffic for very specific requests 

How about using Cloudfare's DNS over HTTPS ? You can make an HTTP request to IP `1.1.1.1`.

https://developers.cloudflare.com/1.1.1.1/encryption/dns-over-https/make-api-requests/

qdm12:
That doesn't help, it would allow other traffic to go through. All this is still blocked by #137

ShrirajHegde:
@qdm12 What about another minimal DNS server that is in the same docker network as gluetun? That can be configured as DNS, without allowing all traffic from gluetun before VPN is up? It can even run on a domain whitelist mode.

This would really help people who want to access their home network with DDNS.

PR #2016, opened 2023-12-19
fix(windscribe): openvpn uses `DEFAULT:@SECLEVEL=0`

qdm12:


qdm12:
Not helping the issue so closing it.

PR #2025, opened 2023-12-31
feat(wireguard): Add persistent keepalive config

abihf:
TODO:

- [ ] Implement other config sources

### DIFF
--- a/internal/configuration/settings/wireguard.go
+++ b/internal/configuration/settings/wireguard.go
@@ -44,6 +44,9 @@ type Wireguard struct {
 	// It defaults to "auto" and cannot be the empty string
 	// in the internal state.
 	Implementation string `json:"implementation"`
+	// PersistenKeepAlive enable sending keep alive signal to server.
+	// So NAT view it as active connection
+	PersistenKeepAlive int `json:"persistent_keepalive"`
 }
 
 var regexpInterfaceName = regexp.MustCompile(`^[a-zA-Z0-9_]+$`)
--- a/internal/configuration/sources/env/wireguard.go
+++ b/internal/configuration/sources/env/wireguard.go
@@ -14,6 +14,10 @@ func (s *Source) readWireguard() (wireguard settings.Wireguard, err error) {
 	wireguard.Interface = s.env.String("VPN_INTERFACE",
 		env.RetroKeys("WIREGUARD_INTERFACE"), env.ForceLowercase(false))
 	wireguard.Implementation = s.env.String("WIREGUARD_IMPLEMENTATION")
+	wireguard.PersistenKeepAlive, err = s.env.Int("WIREGUARD_IMPLEMENTATION", env.ForceLowercase(false))
+	if err != nil {
+		return wireguard, err // already wrapped
+	}
 	wireguard.Addresses, err = s.env.CSVNetipPrefixes("WIREGUARD_ADDRESSES",
 		env.RetroKeys("WIREGUARD_ADDRESS"))
 	if err != nil {
--- a/internal/wireguard/config.go
+++ b/internal/wireguard/config.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net"
 	"net/netip"
+	"time"
 
 	"golang.zx2c4.com/wireguard/wgctrl"
 	"golang.zx2c4.com/wireguard/wgctrl/wgtypes"
@@ -42,9 +43,11 @@ func makeDeviceConfig(settings Settings) (config wgtypes.Config, err error) {
 		}
 		preSharedKey = &preSharedKeyValue
 	}
-
+	var persistenKeepAlive time.Duration
+	if settings.PersistenKeepAlive > 0 {
+		persistenKeepAlive = time.Second * time.Duration(settings.PersistenKeepAlive)
+	}
 	firewallMark := settings.FirewallMark
-
 	config = wgtypes.Config{
 		PrivateKey:   &privateKey,
 		ReplacePeers: true,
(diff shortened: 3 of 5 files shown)


internal/configuration/settings/wireguard.go
```diff
@@ -44,6 +44,9 @@ type Wireguard struct {
 	// It defaults to "auto" and cannot be the empty string
 	// in the internal state.
 	Implementation string `json:"implementation"`
+	// PersistenKeepAlive enable sending keep alive signal to server.
+	// So NAT view it as active connection
+	PersistenKeepAlive int `json:"persistent_keepalive"`
```
qdm12, line 49: let's make this a `*time.Duration` directly here and then you can use s.env.Duration to parse it for env variables.

Also add to the setDefaults method to set it to zero if it's nil, and add the field to the copy, mergeWith etc. methods below.

internal/configuration/settings/wireguard.go
```diff
@@ -44,6 +44,9 @@ type Wireguard struct {
 	// It defaults to "auto" and cannot be the empty string
 	// in the internal state.
 	Implementation string `json:"implementation"`
+	// PersistenKeepAlive enable sending keep alive signal to server.
+	// So NAT view it as active connection
```
qdm12, line 48: nit

```suggestion
	// PersistenKeepAlive enables sending keep alive signals to the Wireguard 
	// server so NAT views it as an active connection. It defaults to 0 which
	// indicates to not use the setting in the Wireguard configuration.
```

internal/configuration/sources/env/wireguard.go
```diff
@@ -14,6 +14,10 @@ func (s *Source) readWireguard() (wireguard settings.Wireguard, err error) {
 	wireguard.Interface = s.env.String("VPN_INTERFACE",
 		env.RetroKeys("WIREGUARD_INTERFACE"), env.ForceLowercase(false))
 	wireguard.Implementation = s.env.String("WIREGUARD_IMPLEMENTATION")
+	wireguard.PersistenKeepAlive, err = s.env.Int("WIREGUARD_IMPLEMENTATION", env.ForceLowercase(false))
```
qdm12, line 17: change the var name 😉 

```suggestion
	wireguard.PersistenKeepAlive, err = s.env.Duration("WIREGUARD_PERSISTENT_KEEP_ALIVE", env.ForceLowercase(false))
```

internal/wireguard/config.go
```diff
@@ -42,9 +43,11 @@ func makeDeviceConfig(settings Settings) (config wgtypes.Config, err error) {
 		}
 		preSharedKey = &preSharedKeyValue
 	}
-
+	var persistenKeepAlive time.Duration
+	if settings.PersistenKeepAlive > 0 {
+		persistenKeepAlive = time.Second * time.Duration(settings.PersistenKeepAlive)
+	}
 	firewallMark := settings.FirewallMark
-
```
qdm12, line 47: nit no need to remove the empty line

internal/wireguard/config.go
```diff
@@ -42,9 +43,11 @@ func makeDeviceConfig(settings Settings) (config wgtypes.Config, err error) {
 		}
 		preSharedKey = &preSharedKeyValue
 	}
-
```
qdm12, line 45: nit no need to remove the empty line

internal/wireguard/config.go
```diff
@@ -42,9 +43,11 @@ func makeDeviceConfig(settings Settings) (config wgtypes.Config, err error) {
...
+	var persistenKeepAlive time.Duration
+	if settings.PersistenKeepAlive > 0 {
+		persistenKeepAlive = time.Second * time.Duration(settings.PersistenKeepAlive)
+	}
```
qdm12, line 49: With settings.PersistentKeepAlive as time.Duration directly, no need to do conversion etc here.

internal/wireguard/config.go
```diff
@@ -68,6 +71,7 @@ func makeDeviceConfig(settings Settings) (config wgtypes.Config, err error) {
 					IP:   settings.Endpoint.Addr().AsSlice(),
 					Port: int(settings.Endpoint.Port()),
 				},
+				PersistentKeepaliveInterval: &persistenKeepAlive,
```
qdm12, line 74: Leave this to `nil` if the persistent keep alive is `0` perhaps? Or maybe it makes no difference if it's nil or 0.

internal/wireguard/settings.go
```diff
@@ -45,6 +45,9 @@ type Settings struct {
 	// Implementation is the implementation to use.
 	// It can be auto, kernelspace or userspace, and defaults to auto.
 	Implementation string
+	// PersistenKeepAlive enable sending keep alive signal to server.
+	// So NAT view it as active connection
```
qdm12, line 49: // PersistenKeepAlive enables sending keep alive signals to the Wireguard 
	// server so NAT views it as an active connection. It defaults to 0, which means
    // it is not to be used.

internal/wireguard/settings.go
```diff
@@ -45,6 +45,9 @@ type Settings struct {
...
+	PersistenKeepAlive int
```
qdm12, line 50: Change the type to time.Duration

qdm12 (changes requested):
Nice thanks. A few changes needed though.

No need to add it to all the settings sources, but add it to the Wireguard toml configuration parsing (I think in the settings/sources/files directory). Thanks!

qdm12:
Done in commit c87c0e12fe50a63a1fe785953bce6f68a03aaf40

PR #2089, opened 2024-02-04
Increase healthcheck start period

Valentin-Metz:
This is relevant for some weaker devices (RaspberryPi).
On mine it currently takes 23 seconds to start.
Increasing this period will not affect faster devices. If a healthcheck succeeds early, the service is considered online.
This is relevant for services that wish to wait for `gluetun` to confirm its health status before coming online themselves.

Example docker-compose section:

```
depends_on:
   glueton:
      condition: service_healthy
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -202,7 +202,7 @@ ENV VPN_SERVICE_PROVIDER=pia \
     PGID=
 ENTRYPOINT ["/gluetun-entrypoint"]
 EXPOSE 8000/tcp 8888/tcp 8388/tcp 8388/udp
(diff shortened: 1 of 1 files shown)


qdm12:
What happens in your logs? Does Gluetun restart internally 3-4 times because it's "unhealthy"?

Valentin-Metz:
> What happens in your logs? Does Gluetun restart internally 3-4 times because it's "unhealthy"?

It restarts ~twice usually.
The solution to fix this was to manually supply an increased startup period in my `docker-compose.yml`.
Apparently it is possible to overwrite individual components of the default health-check settings without overriding the rest.
Closing the pull request, but making this comments so that others who are in a similar situation might find it.

Issue #2024, opened 2023-12-30
Feature request: Adding secrets _FILE backend to credentials => OpenVpn/WireGuard


nolimitech:
### What's the feature 🧐

Implements secrets files. 
Keep the environment variables but add new one with _FILE at the end.
Exemple:
- OPENVPN_USER_FILE=/run/secrets/nordvpn_user
- OPENVPN_PASSWORD_FILE=/run/secrets/nordvpn_password

Should be a MUST. Using .env file should not be the way.

Keep the good work. Great job.

qdm12:
There is already such feature, see https://github.com/qdm12/gluetun-wiki/blob/eb05fb39aee4e299f077796fb8660dafa4aaa3cb/setup/advanced/docker-secrets.md?plain=1

qdm12:
Just the suffix is `_SECRETFILE` not `_FILE`. This may change in the future, but not for now.

Issue #1113, opened 2022-08-21
v4.0.0 Gluetun


qdm12:
### What's the feature 🧐

There are several pending breaking changes that will simplify the code and allow for faster development.
There is no fixed date really for a v4.0.0 release, but it will ultimately happen.

Changes to be made are listed here:

- Remove `FIREWALL_DEBUG` in favor of `LOG_LEVEL=debug`
- Remove Surfshark region retro compatibility (search for `// TODO .*v4`) and remove `RetroLoc` from server model
- Split `DOT_PRIVATE_ADDRESS` in two: ips and cidr ranges
- Rename `UNBLOCK` to a better name
- ProtonVPN: remove `name` field from updater code
- Throw out v0 http server
- MAYBE have the http server as json rpc instead of json rest?

ksurl:
> Rename `UNBLOCK`

How about `DOMAIN_ALLOW_LIST`

moookino:
can you add hostname resolution for VPN_ENDPOINT instead of IP address?

qdm12:
@moookino this is out of scope, see issue #788 - hiding your comment 😉

burnbrigther:
How about ExpressVPN lightway_udp support?  Please? :)

iaacautomation:
On Kubernetes there is a always a good internal DNS setup, so having a way to completely disable the DNS server would be useful. I think it was removed recently, which takes control away from user.

qdm12:
@iaacautomation please voice your concern on #3279 not here. And answer that question over at #3279 

> Using the Kubernetes DNS server would mean all your DNS traffic would leak out the VPN tunnel. The built-in DNS server now "smartly" forwards requests for local names to whatever nameservers were defined in /etc/resolv.conf at container start, effectively it should forward local name requests to the Kubernetes DNS server, whilst still forwarding public name requests through the VPN tunnel. What's missing from this approach?

Marking both your commment and this comment as off-topic.

PR #2109, opened 2024-02-14
Add support for wireguard secrets.

raldone01:
Removed the secret Source and inlined secret handling for simplicity.

Secrets are important because I like to commit my env and yml files.

Addresses: #2024

Note:

I had lots of trouble following all the different sources.
I don't think they add much value and are overly complex.
I would advise to remove the others too and use just one simple config loading module.

This was my first time writing go code so if you want any changes let me know.
Please test it against your setup for the various providers.
I can also update the documentation/wiki but first I want to be sure this is satisfactory.


clayrosenthal:
+1 for this notation. It is commonly used by many docker images, they have a reference to it in the [docs](https://docs.docker.com/compose/use-secrets/). And it is a more secure pattern to use files. Only suggestion would be adding more env vars to be able to be set by this mechanism

qdm12:
Any thoughts on the existing mechanism: https://github.com/qdm12/gluetun/issues/2024#issuecomment-1957343857 ?

raldone01:
I saw the existing implementation which is non standard and does not work for wireguard.

I orignally planned to just add support for wireguard using the existing implementation but found the code quite difficult and complicated to follow.

That there are two pull requests regarding this feature and an issue should be quite telling that there is some room for improvement here.

I also wrote my thoughts on the top. This PR reduces code quite a bit. The comments and longer test actually have it roughly even out with locs.

What you do with this information and PR is up to you.

qdm12 (commented):
Thanks for the contribution, including test and comments 💯 

However, I feel like this is trying to fix a non-existent problem (those `_SECRETFILE` variables do work, right?), EXCEPT adding a secret file reading for Wireguard.

> I had lots of trouble following all the different sources.
> I don't think they add much value and are overly complex.
> I would advise to remove the others too and use just one simple config loading module.

Fair point, and I agree and disagree:
- it was a single loading module a few years ago, and it was unmanageable. That's why it was split up, and source reading was separated from settings definitions (and defaults, validation etc.)
- I'm planning one day to migrate to https://github.com/qdm12/gosettings/tree/v0.4.0-rc9 which handles internally merging of sources and transformation of keys (i.e. `SOME_ENV` becomes `some-env` for flags), I've done the migration for all my projects but Gluetun because there is a lot of nifty details and retro-compatibility that makes this hard. That would result in a single package `internal/configuration` and no subpackages at all 😉

Anyway, for now we should keep the existing structure with those 3 sources.

On top of this, doing major refactor of the configuration directory (here this is a bit of a major ones for a few configuration keys) usually results in mayhem, since it's really easy to miss details (even for me) given retro-compatibility, lowercasing, etc. so I prefer to avoid unless very necessary and rare.

qdm12:
Also see previous PR: https://github.com/qdm12/gluetun/pull/1348

qdm12:
#1348 got merged so now Wireguard secrets (as config file or as individual field-files) are supported. Therefore this PR becomes a bit unneeded, EXCEPT to add a comment to #1113 about using just `_FILE` suffix for both the files source and the secrets source (and drop the `_SECRETFILE` suffix), and probably merge both sources together as a consequence.

qdm12:
- Added the merging of secrets and files sources as a bullet point to https://github.com/qdm12/gluetun/issues/1113
- The 'settings' library was upgraded (from v0.3.x to v0.4.1) which simplified the code a bit, and at least removed ~1500 lines of code.

Closing this since it's no longer necessary, but thanks for the discussion and coding put into it 💯

raldone01:
Awesome! I will try it soon.

Issue #1103, opened 2022-08-09
Feature request: Add support for ProtonVPN to identify and choose servers based upon subscription tier and/or feature set.


DiamondPrecisionComputing:
### What's the feature 🧐

I considered hijacking #1076 but decided that this might be a larger ask than a single feature.

The issue is that not all paid ProtonVPN servers support P2P. (By default, all paid servers that support P2P also support port forwarding.) Proton provides a server list that tells you what servers support P2P but, in their infinite wisdom, does not document what keys and values in their API determine what the server supports.
(It should probably be noted that their free servers support neither P2P nor port forwarding.)

Updating the worker that pulls and parses the JSON data from Proton's API probably shouldn't be that difficult. But, the option to let end users add variables to specify which features they require and then picking a server based on that criteria is a significantly larger ask.

Ultimately, I can see these features being useful for some. Feel free to prioritize based upon which features are more widely used, or reject all together if the demand is not there. :)

1. Free vs paid tier
  A. This is already implemented but I believe it is based upon whether "free" is located in the server name. There is a key that dictates what tier the server is. Unsure if there will be significant change in data accuracy based upon parsing out the server name vs using the key-value pair in the API.

2. Adblocker (NetShield)
  A. Gluetun offers adblocking built-in but implementing this requires no code change, only documentation updates. Simply add the suffix +f1 to the username for blocking malware only. Add suffix +f2 for blocking malware, ads, and trackers.

3. P2P and Port Forwarding
  A. By default, paid servers that support P2P also support Port Forwarding. The Features key-pair determines if this is supported on the server.
  B. I don't believe Gluetun currently looks to verify P2P is supported on the selected server. This could cause issues when using clients that use the P2P protocol for communication as Gluetun would continue to error until the connection was unhealthy and then try another server.

4. TOR
  A. Select servers auto tunnel traffic through a TOR relay. The Features key-pair determines if this is supported on the server.

5. Secure Core/Multihop
  A. Gluetun already provides this as it looks to see if the entry and exit countries are different. The Features key-pair determines if this is supported on the server.  Unsure if there will be significant change if code is updated of this.

A few additional notes.

- Secure Core servers only support Adblocker
- TOR servers only support
- P2P/Port Forwarding servers also support Adblocker
- Normal servers only support Adblocker
- All paid servers, except secure core, support Streaming, Adblocker, 10Gbps local server connection
[...]

PR #2182, opened 2024-03-25
feat(protonvpn): Added ProtonVPN feature selection

### LINKED ISSUE #1582: Feature request: Protonvpn filtering by features
### What's the feature 🧐

Each server JSON object from their API has a feature number corresponding to a feature detailed in https://github.com/ProtonVPN/protonvpn-nm-lib/blob/31d5f99fbc89274e4e977a11e7432c0eab5a3ef8/protonvpn_nm_lib/enums.py#L44-L49

Gluetun should allow to filter servers by feature

_No response_

kvanzuijlen:
I know it's not completely done yet (missing tests, missing todo's for moving feature-related stuff to the new Features struct, docs, etc) but I wanted to open a PR to see if you agree with my chosen direction.

Closes #1582, partially #1103


internal/configuration/settings/serverselection.go
```diff
@@ -54,21 +54,33 @@ type ServerSelection struct { //nolint:maligned
 	// be filtered. This is used with VPN Secure.
 	// TODO extend to providers using FreeOnly.
 	PremiumOnly *bool `json:"premium_only"`
+	// OpenVPN contains settings to select OpenVPN servers
+	// and the final connection.
+	OpenVPN OpenVPNSelection `json:"openvpn"`
+	// Wireguard contains settings to select Wireguard servers
+	// and the final connection.
+	Wireguard WireguardSelection `json:"wireguard"`
```
qdm12, line 62: nit Please move back to the bottom of the struct to reduce deltas / git blame 😉 Thanks!

internal/configuration/settings/serverselection.go
```diff
@@ -107,6 +123,11 @@ func (ss *ServerSelection) validate(vpnServiceProvider string,
 		return fmt.Errorf("for VPN service provider %s: %w", vpnServiceProvider, err)
 	}
 
+	err = validateFeatureFilters(*ss, vpnServiceProvider)
+	if err != nil {
+		return fmt.Errorf("for VPN service provider %s: %w", vpnServiceProvider, err)
+	}
+
 	if *ss.OwnedOnly &&
 		vpnServiceProvider != providers.Mullvad {
```
qdm12, line 132: Why not incorporate OwnedOnly etc. (below) within validateFeatureFilters?

internal/configuration/settings/serverselection.go
```diff
@@ -107,6 +123,11 @@ func (ss *ServerSelection) validate(vpnServiceProvider string,
 		return fmt.Errorf("for VPN service provider %s: %w", vpnServiceProvider, err)
 	}
 
+	err = validateFeatureFilters(*ss, vpnServiceProvider)
+	if err != nil {
+		return fmt.Errorf("for VPN service provider %s: %w", vpnServiceProvider, err)
```
qdm12, line 128:
```suggestion
		return err
```

since the provider name is already in the returned error

internal/configuration/settings/serverselection.go
```diff
@@ -54,21 +54,33 @@ type ServerSelection struct { //nolint:maligned
...
+	SecureCoreOnly *bool `json:"secure_core_only"`
+	// TorOnly is true if VPN servers without tor should
+	// be filtered. This is used with ProtonVPN.
+	TorOnly *bool `json:"tor_only"`
+	// P2POnly is true if VPN servers not for p2p should
+	// be filtered. This is used with ProtonVPN.
+	P2POnly *bool `json:"p2p_only"`
+	// IPv6Only is true if VPN servers without IPv6 support should
+	// be filtered. This is used with ProtonVPN.
+	IPv6Only *bool `json:"ipv6_only"`
```
qdm12, line 83: What does this filter do? 🤔 
Because AFAIK there are only IPv4 addresses to reach Protonvpn servers.

.gitignore
```diff
@@ -1 +1,2 @@
 scratch.txt
+.idea/
```
qdm12, line 2: nit remove before merge 😉 Or we can do a separate PR for it

```suggestion
```

internal/models/server.go
```diff
@@ -29,11 +29,20 @@ type Server struct {
...
 	IPs         []netip.Addr `json:"ips,omitempty"`
+	Features    *Features    `json:"features,omitempty"`
+}
+
+type Features struct {
+	SecureCore bool `json:"secure_core,omitempty"`
+	Tor        bool `json:"tor,omitempty"`
+	P2P        bool `json:"p2p,omitempty"`
+	Streaming  bool `json:"stream,omitempty"`
+	IPv6       bool `json:"ipv6,omitempty"`
```
qdm12, line 45: What's the advantage over having them directly as fields of `Server` as they were?
I see it makes it a bit tidier indeed, but to keep it simple/retro-compatible/no-change-needed, I would keep the same structure as before. You can however add a `// TODO v4 ...` comment describing it would be nice to have them in a Features struct though.

internal/provider/protonvpn/updater/api.go
```diff
@@ -23,6 +23,7 @@ type logicalServer struct {
 	Region      *string          `json:"Region"`
 	City        *string          `json:"City"`
 	Servers     []physicalServer `json:"Servers"`
+	Features    int              `json:"Features"`
```
qdm12, line 26: nit I guess these are positive only numbers, so let's use uint

```suggestion
	Features    uint              `json:"Features"`
```

Or maybe event `uint16` I think? 🤔

internal/provider/protonvpn/updater/iptoserver.go
```diff
@@ -10,7 +10,7 @@ import (
 type ipToServer map[string]models.Server
 
 func (its ipToServer) add(country, region, city, name, hostname string,
-	free bool, entryIP netip.Addr) {
+	free bool, entryIP netip.Addr, features *models.Features) {
```
qdm12, line 13: nit just return the whole bunch of features after `free bool`, or define an unexported `features` struct in this package, up to you 😉 (and then add the `free bool` to it too.

internal/provider/utils/filtering.go
```diff
@@ -45,7 +45,7 @@ func filterServer(server models.Server,
 		return true
 	}
 
-	if *selection.StreamOnly && !server.Stream {
+	if *selection.StreamOnly && (!server.Stream || !server.Features.Streaming) {
```
qdm12, line 48: We should have a single `Stream` field, so probably just set the value of `Features.Streaming` into `Stream`.

internal/storage/filter.go
```diff
@@ -67,7 +67,7 @@ func filterServer(server models.Server,
 		return true
 	}
 
-	if *selection.StreamOnly && !server.Stream {
+	if *selection.StreamOnly && (!server.Stream || !server.Features.Streaming) {
```
qdm12, line 70: same comment here, we need to merge both fields

qdm12 (commented):
Nice it's looking pretty good, just a few minor comments 😉 👍

internal/configuration/settings/serverselection.go
kvanzuijlen: Will do!

internal/configuration/settings/serverselection.go
kvanzuijlen: Not sure if this was a feature or a "subscription tier" filter. If it is a feature then it makes sense to move it as well!

internal/models/server.go
kvanzuijlen: it was mostly not to add too many parameters to the [iptoserver.go add](https://github.com/qdm12/gluetun/pull/2182/files#diff-24ccd1d540dfbe5e9a4e94c5fb8e12ac3ffa3bc0a425a2cd0d70557c51b2eefdR12-R13) function. I'm also fine with just making it part of the server struct though.

internal/provider/protonvpn/updater/api.go
kvanzuijlen: Not familiar enough with Go 😅. Both uint16 and uint sound fine to me, with uint having the advantage of not having to be changed if a feature is added and uint16 not reserving unneeded memory space (if I understand correctly). Let me know which of the 2 has your preference!

internal/provider/protonvpn/updater/iptoserver.go
kvanzuijlen: Will wait with this to see what you decide about the features struct in models/server.go

internal/provider/utils/filtering.go
kvanzuijlen: Will do! (optionally with a TODO for v4, depending on what you decide about the features struct)

internal/configuration/settings/serverselection.go
kvanzuijlen: Not sure! It's indeed not a supported feature by any current server, but I just wanted to implement all features listed in [protonvpn-nm-lib](https://github.com/ProtonVPN/protonvpn-nm-lib/blob/master/protonvpn_nm_lib/enums.py#L44-L49)

internal/provider/protonvpn/updater/api.go
qdm12: `uint` has the advantage to reject negative inputs and have compilation safety against negative numbers.... just in case ™️ 

`uint16` restricts it further (from 64 or 32 bits to 16 bits on all platforms), and perhaps highlight more it's used as a bit array and not just a number. Let's try uint16 😉

internal/models/server.go
qdm12: Let's have these new fields directly in the Server struct, and add an unexported `features` (no first capital letter for unexported types) struct in the updater package to keep the code as clean as it is 😉

internal/provider/protonvpn/updater/iptoserver.go
kvanzuijlen: The features are defined at logical server level, "free" at physical server level, didn't feel right to put it both in the same struct, but will do if you prefer it.

kvanzuijlen:
Hi,

I applied (most of) your suggestions from the code review. I also added minimal unit tests (basically added to all the _test files) for all files changed in this PR. I also built my own image and started using it already. It seems to work, as far as I can tell.

kvanzuijlen:
@qdm12 I've been running this code for a couple of days now myself (https://github.com/users/kvanzuijlen/packages/container/package/gluetun) and as far as I can tell it now fully works :).

internal/configuration/settings/serverselection.go
qdm12: Good point indeed, I've separated the validation in `validateSubscriptionTierFilters` and `validateFeatureFilters`

internal/configuration/settings/serverselection.go
qdm12: Removed due to the strange filter option name `IPV6_ONLY` and because it's unused. We can always re-add it later if needed.

internal/provider/protonvpn/updater/iptoserver.go
qdm12: This makes sense, especially since free is more of a subscription tier filter as you mentioned 😉

qdm12:
1. I rebased your branch on master
2. Extracting features was actually semi-functional... but fear not, it's Go's `switch` which behaves strangely compared to other programming languages. When encountering a true case in the switch, it gets out of the switch and doesn't try other cases. That means if the features bits match one feature, it won't find other features. This is fixed by assigning the features struct fields directly without a switch:

    ```go
    features := features{
    	secureCore: featuresBits&1 != 0,
    	tor:        featuresBits&2 != 0,
    	p2p:        featuresBits&4 != 0,
    	stream:     featuresBits&8 != 0,
    	// ipv6: featuresBits&16 != 0, - unused.
    }
    ```

   FYI this could also been fixed using the `fallthrough` keyword at the end of each case block in the switch 😉 
3. IPv6 only filter got removed. It's a bit of a confusing option `IPV6_ONLY` and also does not seem to be used.
4. Re-updated servers data
5. Added new environment variable options to Dockerfile

    ```Dockerfile
    SECURE_CORE_ONLY= \
    TOR_ONLY= \
    P2P_ONLY= \
    ```

qdm12 (approved):

qdm12:
@kvanzuijlen are you aware of a difference between 'p2p' and 'port forwarding' in terms of feature for Protonvpn? I'm looking at https://github.com/qdm12/gluetun/issues/1103#issuecomment-2255404791 and thinking of removing P2P_ONLY in favor of PORT_FORWARD_ONLY. Also are **all** p2p servers able to port forward?

Issue #1989, opened 2023-11-30
Bug: net/http: TLS handshake timeout but some HTTPS works


maksimkurb:
### Is this urgent?

None

### Host OS

Ubuntu 22.04

### CPU arch

x86_64

### VPN service provider

Custom

### What are you using to run the container

Kubernetes

### What is the version of Gluetun

Running version latest built on 2023-11-28T07:51:53.661Z (commit c326b61)

### What's the problem 🤔

I've configured gluetun to connect to my custom WireGuard server, but container is restarting indefinitely due to failure of downloading some HTTPS URLs:

```
2023-11-30T06:25:50Z ERROR [vpn] cannot get version information: Get "https://api.github.com/repos/qdm12/gluetun/commits": net/http: TLS handshake timeout
```

Firstly, I thought that VPN tunnel is not setted up properly, but then I attached to the pod and tried to `wget` some HTTP and HTTPS pages. It appears that HTTP traffic works fine, but HTTPS works only for certain websites, and it is slower. E.g. `wget httpS://ifconfig.co` is running for 4 seconds, but `wget https://dl-cdn.alpinelinux.org` (and many other domains) is hangs forever.

By wget'ting `http://ifconfig.co` I know that WireGuard tunnel is connected, because I see IP address of my VPN server.
Also, if I use this VPN config on other device, HTTPS traffic works perfectly.

```
/ # time wget -nv -O - http://ifconfig.co
138.[REDACTED] <--- IP Address of VPN Node
2023-11-30 06:22:01 URL:http://ifconfig.co/ [14/14] -> "-" [1]
real    0m 0.38s
user    0m 0.00s
sys     0m 0.00s

/ # time wget -nv -O - https://ifconfig.co
138.[REDACTED] <--- IP Address of VPN Node
2023-11-30 06:22:13 URL:https://ifconfig.co/ [14/14] -> "-" [1]
real    0m 4.63s
user    0m 0.02s
sys     0m 0.00s

/ # time wget -nv -O - http://ipinfo.io/
{
  "ip": "138.[REDACTED]", <--- IP Address of VPN Node
  "city": "Frankfurt am Main",
  "region": "Hesse",
  "country": "DE",
  "loc": "[REDACTED]",
  "org": "[REDACTED]",
  "postal": "[REDACTED]",
  "timezone": "Europe/Berlin",
  "readme": "https://ipinfo.io/missingauth"
}
[...]

PR #2221, opened 2024-04-13
feat(openvpn): Add OPENVPN_MTU env to change default tun-mtu

### LINKED ISSUE #2044: allow to change MTU for all VPNs
### What's the feature 🧐

In certain circumstances MTU needs to be smaller as some networks have issues. Wireguard side seems to have now(link below) MTU parameter but not OpenVPN. Can there be a feature that allows setting lower MTU irrespective of if it is Wireguard or OpenVPN.
I resolve OpenVPN MTU limit problem simply executing below after starting the container, however this should be configurable parameter.
docker exec f86573009ed6 /sbin/ip link set dev tun0 up mtu 1300

https://github.com/qdm12/gluetun/issues/1650

_No response_

PseudoResonance:
This resolves #2044.

tun-mtu directive will be omitted if OPENVPN_MTU is unset.

Additionally lowers the default AirVPN TunMTU to 1320 to resolve connection issues, such as the following: #1417 #1989


qdm12:
Does using for example `OPENVPN_MSSFIX=1320` fix it instead?

qdm12:
> Edit: Never mind, it appears to be working again? I'm not sure what changed, but I'll leave it running for a bit and see.

Yes take your time and let's find out.

I was changing your code actually to have a common `VPN_MTU` common for Wireguard and OpenVPN (and remove `WIREGUARD_MTU`), and default it to `1320` for all. However, I was a bit skeptical this would break OpenVPN setups using the default MTU (I think 1500?). It might also be problematic if the MSSFIX option is used (in that case we should default it to 1500 for retro-compatibility???). Anyway a bit messy in the end, so if we can leave it using OPENVPN_MSSFIX that would be great. Maybe something to write into #1113 (add VPN_MTU, remove OPENVPN_MSSFIX+WIREGUARD_MTU)

PseudoResonance:
After around a month, I think it's been working well with just MSSFIX set. It did have issues a few times for some reason, but after it restarted once or twice it worked. I think the behavior was similar with my MTU flag too though.

qdm12:
Let's close this then, thanks!

Issue #2244, opened 2024-04-29
Bug: Error while connecting


needsupport:
### Is this urgent?

None

### Host OS

Ubuntu

### CPU arch

x86_64

### VPN service provider

Custom

### What are you using to run the container

docker run

### What is the version of Gluetun

latest

### What's the problem 🤔

I can not connect. wire guard config is fine. I tested it 

### Share your logs (at least 10 lines)

```plain text
|   └── Wireguard settings:
|       ├── Private key: 
|       ├── Interface addresses:
|       |   └── 172.17.80.199/32
|       ├── Allowed IPs:
|       |   ├── 0.0.0.0/0
|       |   └── ::/0
|       └── Network interface: tun1
|           └── MTU: 1400
├── DNS settings:
|   ├── Keep existing nameserver(s): no
|   ├── DNS server address to use: 
|   └── DNS over TLS settings:
|       ├── Enabled: yes
|       ├── Update period: every 24h0m0s
|       ├── Unbound settings:
|       |   ├── Authoritative servers:
|       |   |   └── cloudflare
|       |   ├── Caching: yes
|       |   ├── IPv6: no
|       |   ├── Verbosity level: 1
|       |   ├── Verbosity details level: 0
|       |   ├── Validation log level: 0
|       |   ├── System user: root
|       |   └── Allowed networks:
|       |       ├── 0.0.0.0/0
|       |       └── ::/0
|       └── DNS filtering settings:
|           ├── Block malicious: yes
|           ├── Block ads: no
|           ├── Block surveillance: no
|           └── Blocked IP networks:
|               ├── 127.0.0.1/8
|               ├── 10.0.0.0/8
|               ├── 172.16.0.0/12
|               ├── 192.168.0.0/16
|               ├── 169.254.0.0/16
|               ├── ::1/128
|               ├── fc00::/7
|               ├── fe80::/10
|               ├── ::ffff:127.0.0.1/104
|               ├── ::ffff:10.0.0.0/104
|               ├── ::ffff:169.254.0.0/112
|               ├── ::ffff:172.16.0.0/108
|               └── ::ffff:192.168.0.0/112
├── Firewall settings:
|   ├── Enabled: yes
|   └── VPN input ports:
|       └── XX
├── Log settings:
|   └── Log level: info
├── Health settings:
|   ├── Server listening address: 127.0.0.1:9999
|   ├── Target address: cloudflare.com:443
|   ├── Duration to wait after success: 5s
[...]

PR #2249, opened 2024-05-01
fix: crash in custom wireguard connection with no server names

junr03:
Currently accessing `selection.Names[0]` can cause an index out of bounds panic if the user did not set any server names in their config.

This change deletes the line that would cause the crash, while leaving the existing conditional set in the [line below](https://github.com/qdm12/gluetun/compare/master...junr03:gluetun:fix-crash?expand=1#diff-3cd37d747c18d398d306f661c1ff4639931b86ca948354ced26d5e1b74b7190fL62).

### DIFF
--- a/internal/provider/custom/connection.go
+++ b/internal/provider/custom/connection.go
@@ -57,7 +57,6 @@ func getWireguardConnection(selection settings.ServerSelection) (
 		Port:       *selection.Wireguard.EndpointPort,
 		Protocol:   constants.UDP,
 		PubKey:     selection.Wireguard.PublicKey,
-		ServerName: selection.Names[0],
 	}
 	if len(selection.Names) > 0 {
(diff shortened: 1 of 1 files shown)


qdm12:
Hi @junr03 thank you for this wonderful PR! 💯 
Unfortunately I jumped into fixing and pushing a fix on my side without having a look at the pull requests section, my apologies.
Next time feel free to write you opened a PR for it in the issue (for example https://github.com/qdm12/gluetun/issues/2244 here), so I spot it and merge it, I'm always happy to get small and/or first-time contributors involved 😉 
Thanks again 🎖️

junr03:
No worries! And funny enough I did the same but in reverse. I posted the PR without checking in the open issues first :) 

I used to be a full time OSS contributor, but a few years out have made my etiquette rusty!

Thanks for landing a fix either way.

PR #2253, opened 2024-05-02
Current Gluetun with OpenVPN 2.4 and OpenSSL 1.1

qdm12:


qdm12:
No longer needed; there is still the openvpn-2.4 branch if needed.

Issue #2147, opened 2024-03-08
Bug: cannot use custom provider with PIA port forwarding code


Carl-Elkhoury:
### Is this urgent?

No

### Host OS

Ubuntu Jammy

### CPU arch

x86_64

### VPN service provider

Private Internet Access

### What are you using to run the container

Portainer

### What is the version of Gluetun

Running version latest built on 2024-03-07T12:32:25.391Z (commit 3254fc8)

### What's the problem 🤔

As Wireguard isn't officially supported for PIA, I was able to run my PIA wireguard configuration via custom service provider which works great. 
I was trying to add port forwarding by setting VPN_PORT_FORWARDING_PROVIDER to private internet access as written in the wiki [here]( https://github.com/qdm12/gluetun-wiki/blob/main/setup/options/port-forwarding.md) that states: Choose the custom port forwarding code to use. This is useful when using the custom provider with Wireguard.

When I do I get error "panic: server name cannot be empty" (see logs below)
When I try to add the server name, it seems like it is not an accepted parameter under custom provider: "VPN settings: provider settings: server selection: for VPN service provider custom: the server name specified is not valid: one or more values is set but there is no possible value available"

Would you know what I am doing wrong?

Thank you for the help! Love your work on gluetun, this service is awesome 

### Share your logs (at least 10 lines)

```plain text
LOG1:

2024-03-08T18:59:30Z INFO [port forwarding] starting
panic: server name cannot be empty
goroutine 21 [running]:
github.com/qdm12/gluetun/internal/provider/privateinternetaccess.(*Provider).PortForward(0xc001a86918?, {0x109c7e0?, 0xc000194730?}, {{0x7ea4ae8e01b8, 0xc00010ebe0}, {{0x0, 0xffff0a148e01}, 0xc000010018}, 0xc001e9e450, {0x0, ...}})
	github.com/qdm12/gluetun/internal/provider/privateinternetaccess/portforward.go:33 +0x82e
github.com/qdm12/gluetun/internal/portforward/service.(*Service).Start(0xc0000c6a90, {0x109c7e0, 0xc000194730})
	github.com/qdm12/gluetun/internal/portforward/service/start.go:31 +0x391
github.com/qdm12/gluetun/internal/portforward.(*Loop).run(0xc00014d400, {0x109c7e0, 0xc000194730}, 0x0?, 0x0?, 0xc00007fbc0, 0x0?)
[...]

Issue #2048, opened 2024-01-16
PIA: Port forwarding obtaining signature payload timeout


MillsyBot:
Hello!

First off: thanks for making such a cool product! 

Now to business: I have been using PIA (sans port forwarding) for sometime and have really enjoyed it. I am attempting now to add the port forwarding feature for a current use case that I have. I believe that I have configured things properly, however it is quite possible that I missed something. Here are the relevant environment variables

```bash
VPN_PORT_FORWARDING=on
FIREWALL_VPN_INPUT_PORTS=
VPN_ENDPOINT_PORT=
VPN_PORT_FORWARDING_PROVIDER=private internet access
FIREWALL_INPUT_PORTS=
VPN_PORT_FORWARDING_STATUS_FILE=/gluetun/forwarded_port
VPN_PORT_FORWARDING_LISTENING_PORT=0
```

And here are the logs that I am getting.

```bash
2024-01-15T21:18:34-07:00 INFO [vpn] You are running on the bleeding edge of latest!
2024-01-15T21:18:34-07:00 INFO [port forwarding] starting
2024-01-15T21:19:04-07:00 ERROR [vpn] port forwarding for the first time: refreshing port forward data: fetching port forwarding data: obtaining signature payload: Get "https://10.31.110.1:19999/getSignature?token=<token>": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
```

I have verified on a few locations that claim they support port forwarding, however the results are the same. I attempted to reach the 10.31.110.1 port 19999 from inside the gluetun container, and that was also a bust.

Thanks in advance for any help. Sorry if this one is too obvious!

qdm12:
Hi there! I am not sure, I was going to say this VPN server probably doesn't support port forwarding... What server hostname/ip are you using, just to double check? If anyone has the same issue, please chime in as well, thanks!

MillsyBot:
Hey! Thanks for the response.

```
SERVER_NAMES=vancouver433
SERVER_HOSTNAMES=ca-vancouver.privacy.network
SERVER_REGIONS=CA Vancouver
```

I have been using Vancouver to make the attempts, however I have cycled a few different servers in various regions (Venezuela, Mexico, Norway).

Is there a preferred or recommended order of preference on these variables? like only use `SERVER_NAMES` or use all three or use only one?

ZulliB:
I have been dealing with the same issue for a while and here is the error I get:
[...]

PR #2254, opened 2024-05-02
fix(pia): port forward using server IP instead of gateway ip

qdm12:
For issue #2320  (also related issues are #2048  and #2147)

⚠️ comment in https://github.com/qdm12/gluetun/issues/2147 that users should now use SERVER_HOSTNAMES instead of SERVER_NAMES.


qdm12:
No longer needed.

PR #2272, opened 2024-05-09
feat: add ability to configure the amount of times the healthcheck fails

cdloh:
Adds the ability to configure the amount of times that the healthcheck can fail before the system will restart the VPN

### Context

My internet gets a bit flaky when it's maxed out attempting to put traffic through the VPN. There's nothing actually wrong with the connection and traffic is flowing fine but the healthcheck times out or refuses to connect every so often.

This then causes the VPN to flap and restart and then connect again and get stuck in a loop.

This change allows the healthcheck to attempt a few more times. I've been running this for a while and it solves my issue and keeps the connection healthy.

If you're happy with this feature I'm happy to clean it up a bit more and update the wiki as required.


cdloh:
Looks like you pushed some healthcheck changes in latest. I'll test to see if it largely solves my issues and discard this if it does 👍

qdm12:
> Adds the ability to configure the amount of times that the healthcheck can fail before the system will restart the VPN

Why not just increase `HEALTH_VPN_DURATION_INITIAL`?

> Looks like you pushed some healthcheck changes in latest. I'll test to see if it largely solves my issues and discard this if it does

Indeed, sorry maybe I missed your PR before. The change might help, it essentially increases the dialing timeout on every try (instead of fixed 3 seconds), and the health initial vpn duration was increased as well.

I don't think the number of attempts adds much value, although it could replace the existing timer (but retro-compatibility, so we're stuck until v4). Now feel free to discuss ideas etc. to improve it as well!

qdm12:
Closing this due to 6042a9e3c24d21ef900d15a880c382884f342613 changing how the healthcheck works. Thanks for the PR and taking the time to propose these changes still 💯 🎖️

Issue #1785, opened 2023-08-05
Feature request: run a script on interface up/down


Diaoul:
### What's the feature 🧐

Run a shell script when interface is up (or down)

### Extra information and references

It can be useful to set a dynamic ip in a whitelist somewhere (like [here](https://www.myanonamouse.net/api/endpoint.php/3/json/dynamicSeedbox.php))

joshhoughton:
Agreed, this would be extremely useful!

pauldiloreto:
+1, would love to have this as an option.

Does anyone have a work around for this scenario?

mtfurlan:
For wireguard, we just need to set [PostUp/PostDown](https://github.com/pirate/wireguard-docs?tab=readme-ov-file#preup) in the config.

Given that the both openvpn and wireguard have support for this, I think a reasonable interface would be something along the lines of
```
environment:
    POST_INTERFACE_UP_SCRIPT=/gluetun/tun_up.sh
volumes:
    - /path/to/tun_up.sh:/gluetun/tun_up.sh
```

qdm12:
Prioritizing this since it's kind of long overdued. Also it will be as a command to be cross-OS instead of a script path (but you could do `/bin/sh -c "my shell commands"` if you want.

This is also kind of needed to make sense of newly opened PR #2392  (more details why https://github.com/qdm12/gluetun/discussions/2393#discussioncomment-10288125)

qdm12:
Up command is finally available: https://github.com/qdm12/gluetun-wiki/blob/main/setup/advanced/vpn-port-forwarding.md#custom-port-forwarding-up-command
Working on the down command right now

qdm12:
- [↪️ Documentation for up & down commands](https://github.com/qdm12/gluetun-wiki/blob/main/setup/advanced/vpn-port-forwarding.md#custom-port-forwarding-updown-command)
- Up command added in a035a151bd44ff85087b05f62fc72bcda63af60a
- Down command added in 0374c14e429df16734d2cfa0784027f43624c247

Enjoy! 🎉

PR #2399, opened 2024-08-06
feat(portforwarding): Allow running script upon port forwarding success

### LINKED ISSUE #2351: Feature request: Webhook call when forwarded port change
### What's the feature 🧐

I use gluetun with PIA and forwarding. I have seen many ways to read the forwarded port from gluetun (most of them using the fwd_port file, wich I read is being deprecated) but I think the best way would be gluetun that makes a web request by itself calling a user-configured webhook when the port is being changed and maybe at the start of gluetun too.
Making the web request entirely configurable with a template inside the docker compose file (or a separate config file) would be ideal. 
I understand it can take some time to implement such a thing, so a simple GET or POST request with the forwarded port to a user-configurable endpoint it'd be nice too.

I found [a program (written in Go)](https://github.com/afiestas/gluetun-sync) that implement this, if you want to take a look 

Thank you for your work!

_No response_

lavalleeale:
This PR allows an environment variable (VPN_PORT_FORWARDING_STATUS_SCRIPT) to be set that will run a script whenever there is an update to port forwarding status that could be used to sync with other programs that need to know what ports have been forwarded.

EDIT by qdm12:

- Fixes #2351 
- Partly addresses #1785 
- May refer to #616, #1555 (more to completed)


internal/portforward/service/script.go
```diff
@@ -0,0 +1,29 @@
...
+	"strings"
+)
+
+func (s *Service) runPortForwardedScript(ports []uint16) (err error) {
+	// run bash script with ports as arguments
+	portStrings := make([]string, len(ports))
+	for i, port := range ports {
+		portStrings[i] = fmt.Sprint(int(port))
+	}
+	portsString := strings.Join(portStrings, " ")
```
qdm12, line 16: we should have this as a comma separated value ideally, it's easier to split by comma then taking a variable number of arguments I think (across shells)

internal/portforward/service/script.go
```diff
@@ -0,0 +1,29 @@
...
+
+	scriptPath := s.settings.Scriptpath
+	s.logger.Info("running port forward script " + scriptPath)
+	cmd := exec.Command(scriptPath, portsString)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	err = cmd.Run()
```
qdm12, line 23: This should be context (ctx context.Context) dependent, I think there is a exec.CommandContext for that.

internal/portforward/service/start.go
```diff
@@ -69,6 +69,14 @@ func (s *Service) Start(ctx context.Context) (runError <-chan error, err error)
 		return nil, fmt.Errorf("writing port file: %w", err)
 	}
 
+	if s.settings.Scriptpath == "" {
```
qdm12, line 72: This needs to change anyway, but just in case ™️ it should be

```suggestion
	if s.settings.Scriptpath != "" {
```

internal/portforward/service/start.go
```diff
@@ -80,7 +88,8 @@ func (s *Service) Start(ctx context.Context) (runError <-chan error, err error)
 	s.keepPortDoneCh = keepPortDoneCh
 
 	go func(ctx context.Context, portForwarder PortForwarder,
-		obj utils.PortForwardObjects, runError chan<- error, doneCh chan<- struct{}) {
+		obj utils.PortForwardObjects, runError chan<- error, doneCh chan<- struct{},
+	) {
```
qdm12, line 92: nit unneeded change, please revert 😉

qdm12 (changes requested):
Nice, thank you for this PR 👍 

We need to notably:

- change this to a run command (instead of script file)
- make code ready for a down command (later 🙏, if anyone needs such command)
- [make a PR](https://github.com/qdm12/gluetun-wiki/compare) to the github wiki to document this (probably in the setup/advance/vpn-port-forwarding document) and mention the command is a single command, which doesn't understand shell specific syntax such as `&&`, and one should use `/bin/sh -c "my shell syntax"` to do so if they want (bash isn't installed by default, only sh)

Thanks!

lavalleeale:
Ok, I believe that I have resolved all the issues with the initial implementation, I am not a huge fan of comma separating the ports, but that is how the current code works to give maximum compatibility

internal/configuration/settings/portforward.go
```diff
@@ -29,6 +29,11 @@ type PortForwarding struct {
 	// to write to a file. It cannot be nil for the
 	// internal state
 	Filepath *string `json:"status_file_path"`
+	// Command is the port forwarding status command
+	// to use. It can be the empty string to indicate not
+	// to run a command. It cannot be nil for the
+	// internal state
+	Command *string `json:"status_command"`
```
qdm12, line 36: Sorry to re-iterate on this, can you rename this to `UpCommand`? To plan ahead for an eventual `DownCommand` 🤔 Thanks!

internal/portforward/service/command.go
```diff
@@ -0,0 +1,32 @@
...
+	portsString := strings.Join(portStrings, ",")
+
+	rawCommand := strings.ReplaceAll(s.settings.Command, "{{PORTS}}", portsString)
+	s.logger.Info("running port forward command " + rawCommand)
+	command := strings.Split(rawCommand, " ")
+	// it is a user input and we trust it so we can ignore the gosec warning
+	cmd := exec.CommandContext(ctx, command[0], command[1:]...) // #nosec G204
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	err = cmd.Run()
```
qdm12, line 26: This should be replaced by `github.com/qdm12/golibs/command` (soon to be local in #2418) and plugged in the logger somehow, to avoid race conditions on os.Stdout (and have smushed logs). I'll push this on your branch no worries😉

internal/portforward/service/command.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+)
+
+func (s *Service) runUpCommand(ctx context.Context, ports []uint16) (err error) {
+	// run command replacing {{PORTS}} with the ports (space separated)
```
qdm12, line 12: any better suggestion than using CSV to seperate values? That could be added in #1113 !

internal/portforward/service/settings.go
```diff
@@ -12,6 +12,7 @@ type Settings struct {
 	Enabled        *bool
 	PortForwarder  PortForwarder
 	Filepath       string
+	Command        string
```
qdm12, line 15: Same here, please rename to `UpCommand` 😉

internal/portforward/service/start.go
```diff
@@ -69,6 +69,14 @@ func (s *Service) Start(ctx context.Context) (runError <-chan error, err error)
 		return nil, fmt.Errorf("writing port file: %w", err)
 	}
 
+	if s.settings.Command != "" {
+		err = s.runUpCommand(ctx, ports)
+		if err != nil {
+			_ = s.cleanup()
+			return nil, fmt.Errorf("running port forward command: %w", err)
+		}
```
qdm12, line 77: I wonder, if the command fails, should port forwarding fail to start, and possibly then crash the container 🤔 
Maybe it would make more sense to just log it as an error, especially given port forwarding is *working*, it's just the user provided script that failed (which is somehow likely to happen imo)

qdm12 (commented):
Nice job! 👍 
Just 2 renames to do 😉 
And once done I'll handle the using-that-command-package-and-plug-lines-to-the-logger. I just worked on it yesterday so it's still fresh in my mind 😉 !

internal/configuration/settings/portforward.go
```diff
@@ -162,6 +175,9 @@ func (p *PortForwarding) read(r *reader.Reader) (err error) {
 			"PRIVATE_INTERNET_ACCESS_VPN_PORT_FORWARDING_STATUS_FILE",
 		))
 
+	p.Command = r.Get("VPN_PORT_FORWARDING_UP_COMMAND",
```
qdm12, line 178: Also please add `VPN_PORT_FORWARDING_UP_COMMAND=` to the Dockerfile's ENV list thanks!

internal/portforward/service/command.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+)
+
+func (s *Service) runUpCommand(ctx context.Context, ports []uint16) (err error) {
```
qdm12, line 11: I'll add a test to make sure for example this works: `/bin/sh -c "echo {{PORTS}}"` - given it plugs arguments by splitting them by spaces, I'm not sure it would handle that case with double quotes correctly. Will check!

internal/configuration/settings/portforward.go
```diff
@@ -29,6 +29,11 @@ type PortForwarding struct {
 	// to write to a file. It cannot be nil for the
 	// internal state
 	Filepath *string `json:"status_file_path"`
+	// Command is the port forwarding status command
```
qdm12, line 32: Self comment to self-resolve: don't forget to merge PR https://github.com/qdm12/gluetun-wiki/pull/89 !

qdm12:
Hi @lavalleeale could you give me (force) push access to your Gluetun branch `master`? I'm done with the changes, just waiting to force push them there.

thibautd:
Thanks for working on this, can't wait !

iceteaSA:
Anything happening here?

qdm12:
@lavalleeale I still don't have push access to your fork, can you please allow force push access? Otherwise I'll create another PR with another branch, but would like to give you the credits for it ideally!

lavalleeale:
Done!

qdm12:
Thanks @lavalleeale 
Just waiting for CI to pass, and I'll re-recheck my own still unresolved comments to make I've done them (has been a while already!).

qdm12 (approved):

qdm12:
Merged, thanks! I've tested it locally, it looks like it's working fine.
I also added VPN_PORT_FORWARDING_DOWN_COMMAND` in the next commit, which does what its name says.
And also pushed f9bdb219d0837b6d8efad38bab77657523428246 to support quoting better i.e. `echo "hello"` would previously be changed to `echo "hello` which would fail the parsing step, now it's fixed.

KSmanis:
@qdm12 Thanks for merging! I've already successfully migrated my setup from polling the control server to this hook, but I'd rather not follow the master branch. Any plans on a new release soon?

thibautd:
For those wanting to use this new feature with Transmission, I wrote a simple Python script to automatically update the `Peer Port` in Transmission. It's written in Python and uses the [uv](https://github.com/astral-sh/uv) package manager.

**docker-compose.yml**
```
services:
  gluetun:
    build: ./gluetun
    container_name: gluetun
    cap_add:
      - NET_ADMIN
    devices:
      - /dev/net/tun:/dev/net/tun
    environment:
      - Your other env vars here...
      - VPN_PORT_FORWARDING_UP_COMMAND=/bin/uv run --script /transmission-sync.py {{PORTS}}
    restart: unless-stopped
    networks:
      - internal
  transmission:
    image: lscr.io/linuxserver/transmission:latest
    container_name: transmission
    network_mode: "service:gluetun"
    volumes:
      - # Your volumes here
    restart: unless-stopped
```

I had to create a custom Docker image for gluetun to install Python, `uv` and my script inside:

**gluetun/Dockerfile**
```
FROM qmcgaw/gluetun:latest

RUN apk add --no-cache python3

COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
COPY transmission-sync.py /
```

Finally, the Python script wich uses https://pypi.org/project/transmission-rpc/.

**gluetun/transmission-sync.py**
```
# /// script
# requires-python = ">=3.12"
# dependencies = [
#     "requests",
#     "transmission-rpc",
# ]
# ///

import sys
import requests
from transmission_rpc import Client

PORTS = [int(p) for p in sys.argv[1].split(",")]

print('Set port: ', PORTS)

CLIENT = Client(host="127.0.0.1", port=9091)
CLIENT.set_session(peer_port=PORTS[0])

print('Done')
```

I put that here in case it can be useful :-)

Marsu31:
Thank you for this great feature !!! :-D
My 50 cents, for qbittorrent container:

`VPN_PORT_FORWARDING_UP_COMMAND=wget -qO - --post-data "json={\"listen_port\":{{PORTS}}}" http://127.0.0.1:${QBITTORRENT_UI_PORT}/api/v2/app/setPreferences`

Prerequisites:
Whitelist localhost connection in qbittorent options.
Have environment variable `QBITTORRENT_UI_PORT` set or replace by port.

Have a great holiday :-)

ramennbowls:
Thank you @Marsu31 for your command. I managed to simplify it like so 👇
```env
VPN_PORT_FORWARDING_UP_COMMAND=wget -qO- --post-data=json={\"listen_port\":{{PORTS}}} http://127.0.0.1:8080/api/v2/app/setPreferences
```
---
If your Gluetun container starts before qBittorrent, the command will fail because there is no service to contact yet. personally have qBittorrent depend on Gluetun, and will not start until it's healthy. 
```yml
qbittorrent:
    ...
    depends_on:
      gluetun:
        condition: service_healthy
```
For reliability, consider adding the `--retry-connrefused` and `--tries=6` flags.
- `--retry-connrefused` tells wget to retry if the address doesn't resolve (incorrect or likely down). Optionally, you can use the `--waitretry` flag to specify the time to wait in seconds between retries. By default, `--retry-connrefused` waits 1 second but doubles with each retry _(1s → 2s → 4s → 8s ...)_.
- `--tries` allows you to specify the maximum number of retries. I recommend setting this to 6, because `--retry-connrefused` doubles its wait between retries (unless specified otherwise) which totals to ~30 seconds worth of retry.

All in all, I'd suggest using this command 👇
```env
VPN_PORT_FORWARDING_UP_COMMAND=wget --retry-connrefused --tries=6 -qO- --post-data=json={\"listen_port\":{{PORTS}}} http://127.0.0.1:8080/api/v2/app/setPreferences
```
---
Personally, instead of using the `-qO-` flag, I prefer to use `-nv` which gives you a less verbose output so you know what's going on, but Gluetun seems to interpret all outputs from the command as errors so 🤔
```md
ERROR [port forwarding] failed: Connection refused.
ERROR [port forwarding] URL:http://127.0.0.1:8080/api/v2/app/setPreferences [0/0] -> "setPreferences" [4] *👈 not an error*
```

redge76:
How do you guys handle the gluetun container restart ? 
If qbittorrent is connected to the gluetun network, it lose connectivity until it is restarted.
Some background: https://stackoverflow.com/questions/47088261/restarting-an-unhealthy-docker-container-based-on-healthcheck
(the second option does not work with qbittorent ...)

Marsu31:
Thank you @ramennbowls for your improvements !  :-)

The lack of logging made me not see the error you pointed out.

To see output not in "ERROR" category I had to surround the command with shell and redirect stderr like this:

`VPN_PORT_FORWARDING_UP_COMMAND=/bin/sh -c '/usr/bin/wget -O- --retry-connrefused --post-data "json={\"listen_port\":{{PORTS}}}" http://127.0.0.1:${QBITTORRENT_UI_PORT}/api/v2/app/setPreferences 2>&1'`

I don't use `-nv` nor `-q` to be sure to see something. It doesn't happen often, also there is no performance problem (for me ;-) ).

Issue #2334, opened 2024-06-24
Bug: cannot remove allowed input port from firewall


mjsjml:
### Is this urgent?

No

### Host OS

Ubuntu 22.04

### CPU arch

x86_64

### VPN service provider

AirVPN

### What are you using to run the container

docker-compose

### What is the version of Gluetun

Running version latest built on 2024-06-17T22:37:52.988Z (commit 93ed87d)

### What's the problem 🤔

When gluetun fails a healthcheck and gets a new vpn connection, it can't resume port forwarding.  The desired ports will be closed on the new connection.  Things used to work fine with an older version of gluetun (don't know which one).

ERROR [vpn] cannot remove allowed input port from firewall: removing allowed port 63799 on interface tun0: command failed: "ip6tables --delete INPUT -i tun0 -p tcp --dport 63799 -j ACCEPT": ip6tables: Bad rule (does a matching rule exist in that chain?).: exit status 1

### Share your logs (at least 10 lines)

```plain text
2024-06-24T00:21:55.555343458Z 2024-06-23T17:21:55-07:00 INFO [routing] default route found: interface eth0, gateway 172.18.0.1, assigned IP 172.18.0.10 and family v4
2024-06-24T00:21:55.555374628Z 2024-06-23T17:21:55-07:00 INFO [routing] adding route for 0.0.0.0/0
2024-06-24T00:21:55.555422948Z 2024-06-23T17:21:55-07:00 INFO [firewall] setting allowed subnets...
2024-06-24T00:21:55.558813825Z 2024-06-23T17:21:55-07:00 INFO [routing] default route found: interface eth0, gateway 172.18.0.1, assigned IP 172.18.0.10 and family v4
2024-06-24T00:21:55.558863994Z 2024-06-23T17:21:55-07:00 INFO [routing] adding route for 192.168.2.0/24
2024-06-24T00:21:55.558878991Z 2024-06-23T17:21:55-07:00 INFO [routing] adding route for 192.168.3.0/24
2024-06-24T00:21:55.559692529Z 2024-06-23T17:21:55-07:00 INFO [dns] using plaintext DNS at address 1.1.1.1
2024-06-24T00:21:55.560152708Z 2024-06-23T17:21:55-07:00 INFO [http server] http server listening on [::]:8010
2024-06-24T00:21:55.560488916Z 2024-06-23T17:21:55-07:00 INFO [healthcheck] listening on 127.0.0.1:9999
[...]

PR #2411, opened 2024-08-14
fix(firewall): delete chain rules by line number

qdm12:
Image is with tag `:pr-2411`. It now deletes iptables rules by line number to go around iptables not finding existing rules for deletion, fixing #2334 

- [x] Initial implementation
- [x] Some unit tests here and there
- [x] Self-tested locally 
- [x] Feedback received


internal/firewall/delete.go
```diff
@@ -0,0 +1,84 @@
...
+	"fmt"
+	"os/exec"
+	"strings"
+)
+
+func isDeleteInstruction(instruction string) bool {
+	fields := strings.Fields(instruction)
+	for i := 0; i < len(fields); i += 2 {
+		switch fields[i] {
+		case "-D", "--delete": //nolint:goconst
```
qdm12, line 14: we should make this doesn't delete a rule by line number; otherwise we should not try to find the rule line number from this.

PR #2416, opened 2024-08-15
feat(settings): add storage filepath settings

### LINKED ISSUE #2074: Feature request: remove hardcoded link to /gluetun
### What's the feature 🧐

Currently, the location for files like the `servers.json` is all hardcoded to `/gluetun`.

Feature request: please implement the `HOME` variable in gluetun and the Dockerfile. Using gluten on an orchestrator like `Kubernetes` or `Nomad` is difficult because of this hardcoded issue. To have gluetun running, I must mount an additional volume just for a single file. The entire container has restrictions like read-only filesystem except for several locations, like secrets, etc.

When using `Nomad` you get a `/local` directory, where you can write files; I would also be happy with a solution that adds a variable like `NOMAD=1` which will change the location of the `/gluetun` directory to `/local` so that I do not need an additional volume.

_No response_

jfroy:
This patch adds a new storage settings object with a single field to
control the storage filepath.

One use case is to set the filepath to the empy string, which will
disable reading and writing of the servers.json file. This is useful for
deployments that want to minimize I/O and want to set tight memory
limits (writing the servers.json spikes memory).

Fixes #2074

### DIFF
--- a/internal/configuration/settings/storage.go
+++ b/internal/configuration/settings/storage.go
@@ -22,6 +22,7 @@ func (s Storage) validate() (err error) {
 			return fmt.Errorf("filepath is not valid: %w", err)
 		}
 	}
+	return nil
 }
 
 func (s *Storage) copy() (copied Storage) {
--- a/internal/configuration/settings/storage.go
+++ b/internal/configuration/settings/storage.go
@@ -1,7 +1,9 @@
 package settings
 
 import (
-	"github.com/qdm12/gluetun/internal/constants"
+	"fmt"
+	"path/filepath"
+
 	"github.com/qdm12/gosettings"
 	"github.com/qdm12/gosettings/reader"
 	"github.com/qdm12/gotree"
@@ -45,7 +47,7 @@ func (s StorageSettings) toLinesNode() (node *gotree.Node) {
 	if *s.Filepath == "" {
 		return gotree.New("Storage settings: disabled")
 	}
-	node = gotree.New("Storage settings:")		
+	node = gotree.New("Storage settings:")
 	node.Appendf("Filepath: %s", *s.Filepath)
 	return node
 }
--- a/internal/configuration/settings/storage.go
+++ b/internal/configuration/settings/storage.go
@@ -14,24 +14,27 @@ type StorageSettings struct {
 }
 
 func (s StorageSettings) validate() (err error) {
-	return nil
+	if *s.Filepath != "" { // optional
+		_, err := filepath.Abs(*s.Filepath)
+		if err != nil {
+			return fmt.Errorf("filepath is not valid: %w", err)
+		}
+	}
 }
 
 func (s *StorageSettings) copy() (copied StorageSettings) {
 	return StorageSettings{
-		Filepath: s.Filepath,
+		Filepath: gosettings.CopyPointer(s.Filepath),
 	}
 }
 
-// overrideWith overrides fields of the receiver
-// settings object with any field set in the other
-// settings.
 func (s *StorageSettings) overrideWith(other StorageSettings) {
 	s.Filepath = gosettings.OverrideWithPointer(s.Filepath, other.Filepath)
 }
 
 func (s *StorageSettings) setDefaults() {
-	s.Filepath = gosettings.DefaultPointer(s.Filepath, constants.ServersData)
+	const defaultFilepath = "/gluetun/servers.json"
+	s.Filepath = gosettings.DefaultPointer(s.Filepath, defaultFilepath)
 }
 
 func (s StorageSettings) String() string {
@@ -39,7 +42,10 @@ func (s StorageSettings) String() string {
 }
 
 func (s StorageSettings) toLinesNode() (node *gotree.Node) {
-	node = gotree.New("Storage settings:")
+	if *s.Filepath == "" {
+		return gotree.New("Storage settings: disabled")
+	}
+	node = gotree.New("Storage settings:")		
 	node.Appendf("Filepath: %s", *s.Filepath)
 	return node
 }
--- a/internal/configuration/settings/storage.go
+++ b/internal/configuration/settings/storage.go
@@ -1,0 +1,50 @@
+package settings
+
+import (
+	"github.com/qdm12/gluetun/internal/constants"
+	"github.com/qdm12/gosettings"
+	"github.com/qdm12/gosettings/reader"
+	"github.com/qdm12/gotree"
+)
+
+// StorageSettings contains settings to configure the storage.
+type StorageSettings struct {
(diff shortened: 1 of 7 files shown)


internal/configuration/settings/storage.go
```diff
@@ -0,0 +1,50 @@
...
+
+// overrideWith overrides fields of the receiver
+// settings object with any field set in the other
+// settings.
+func (s *StorageSettings) overrideWith(other StorageSettings) {
+	s.Filepath = gosettings.OverrideWithPointer(s.Filepath, other.Filepath)
+}
+
+func (s *StorageSettings) setDefaults() {
+	s.Filepath = gosettings.DefaultPointer(s.Filepath, constants.ServersData)
```
qdm12, line 34: nit we can probably set the constant here and remove `constants.ServersData`:

```suggestion
	const defaultFilepath = "/gluetun/servers.json"
	s.Filepath = gosettings.DefaultPointer(s.Filepath, defaultFilepath)
```

internal/configuration/settings/storage.go
```diff
@@ -0,0 +1,50 @@
...
+
+func (s *StorageSettings) copy() (copied StorageSettings) {
+	return StorageSettings{
+		Filepath: s.Filepath,
+	}
+}
+
+// overrideWith overrides fields of the receiver
+// settings object with any field set in the other
+// settings.
```
qdm12, line 28: nit a bit unneeded 🤔

```suggestion
```

internal/configuration/settings/storage.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"github.com/qdm12/gluetun/internal/constants"
+	"github.com/qdm12/gosettings"
+	"github.com/qdm12/gosettings/reader"
+	"github.com/qdm12/gotree"
+)
+
+// StorageSettings contains settings to configure the storage.
+type StorageSettings struct {
```
qdm12, line 11: nit rename to just `Storage`, since we're already in the settings package.

internal/configuration/settings/storage.go
```diff
@@ -0,0 +1,50 @@
...
+	s.Filepath = gosettings.DefaultPointer(s.Filepath, constants.ServersData)
+}
+
+func (s StorageSettings) String() string {
+	return s.toLinesNode().String()
+}
+
+func (s StorageSettings) toLinesNode() (node *gotree.Node) {
+	node = gotree.New("Storage settings:")
+	node.Appendf("Filepath: %s", *s.Filepath)
```
qdm12, line 43:
```suggestion
	if *s.Filepath == "" {
		return gotree.New("Storage settings: disabled")
	}
	node = gotree.New("Storage settings:")		
	node.Appendf("Filepath: %s", *s.Filepath)
```

internal/configuration/settings/storage.go
```diff
@@ -0,0 +1,50 @@
...
+	return node
+}
+
+func (s *StorageSettings) read(r *reader.Reader) (err error) {
+	s.Filepath = r.Get("STORAGE_FILEPATH", reader.AcceptEmpty(true))
```
qdm12, line 48: Please add `STORAGE_FILEPATH=/gluetun/servers.json` to the Dockerfile in the `ENV` section, thanks!

internal/configuration/settings/storage.go
```diff
@@ -0,0 +1,50 @@
...
+	Filepath *string
+}
+
+func (s StorageSettings) validate() (err error) {
+	return nil
+}
+
+func (s *StorageSettings) copy() (copied StorageSettings) {
+	return StorageSettings{
+		Filepath: s.Filepath,
```
qdm12, line 22:
```suggestion
		Filepath: gosettings.CopyPointer(s.Filepath),
```

internal/configuration/settings/storage.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+// StorageSettings contains settings to configure the storage.
+type StorageSettings struct {
+	// Filepath is the path to the servers.json file. An empty string disables on-disk storage.
+	Filepath *string
+}
+
+func (s StorageSettings) validate() (err error) {
+	return nil
```
qdm12, line 17:
```suggestion
	if *s.Filepath != "" { // optional
		_, err := filepath.Abs(*s.Filepath)
		if err != nil {
			return fmt.Errorf("filepath is not valid: %w", err)
		}
	}
```

internal/configuration/settings/storage.go
jfroy: There is an existing `Storage` interface in the `settings` package that conflicts. I didn't want to make this PR much larger by trying to tackle that at the same time.

internal/configuration/settings/storage.go
jfroy: I left it out because when the setting is unset it will use the default. That's also why I had to use a pointer -- not set means use default, empty string means disable writing. In any case, I can certainly add the setting with its default value in the dockerfile.

internal/configuration/settings/storage.go
qdm12: Indeed, my bad! Let's rename the `Storage` interface to `FilterChoicesGetter` , you have the 🟢 light! Thanks for the initial thought of keeping the PR smaller, it's appreciated!

internal/configuration/settings/storage.go
qdm12: > I left it out because when the setting is unset it will use the default.

Indeed, but this is for documentation purposes for the Docker image, to show what's the default.

qdm12: My bad, I just re-read this, I previously misunderstood. I think `reader.AcceptEmpty(true)` does solve it indeed, good catch! My apologies for not pointing you in that direction!

qdm12 (dismissed):
Awesome 💯 🎖️ 👍 !

qdm12:
Just two more things to do:

1. Rebase your branch on the master branch. It looks like it has conflicts, I'm not too sure why though since nothing changed too much recently 🤔 Hopefully not too much of a hassle, otherwise let me know!
2. Can you create a PR adding a storage.md file in https://github.com/qdm12/gluetun-wiki/tree/main/setup/options describing that new environment variable 🙏 Thanks!!

jfroy:
> 1. Rebase your branch on the master branch. It looks like it has conflicts, I'm not too sure why though since nothing changed too much recently 🤔 Hopefully not too much of a hassle, otherwise let me know!

The rename of the interface conflicts with the add-on of the `Warner Warner` parameter. Easy to merge.

qdm12 (approved):

qdm12:
Awesome, thanks for your contribution 💯 🥳 ! Merged 😉 !

PR #2423, opened 2024-08-18
feat(control server authentication): add basic http auth

### LINKED ISSUE #2238: Feature Request: control server authentication
### Is this urgent?

No

### Host OS

Fedora CoreOS

### CPU arch

x86_64

### VPN service provider

Custom

docker run

Running version v3.38.0 built on 2024-03-25T15:53:33.983Z (commit b3ceece)

### What's the problem 🤔

The internal service exposes its complete configuration to unauthenticated local services and users. It is also possible to change the state of the VPN by restarting the VPN (and maybe other things).

A basic protection using an Authorization Header would be good to ensure no thirdparty that might know of the presence of gluetun can gain the VPN credentials.
Otherwise it's nice to have the API.

```plain text
# curl http://localhost:8000/v1/vpn/settings

{"type":"wireguard","provider":{"name":"*******","server_selection":{"vpn":"wireguard","target_ip":"0.0.0.0","countries":["********"],"categories":null,"regions":[],"cities":null,"isps":null,"names":null,"numbers":null,"hostnames":null,"owned_only":false,"free_only":false,"premium_only":false,"stream_only":false,"multi_hop_only":false,"port_forward_only":false,"openvpn":{"config_file_path":"","tcp":false,"custom_port":0,"pia_encryption_preset":""},"wireguard":{"endpoint_ip":"0.0.0.0","endpoint_port":0,"public_key":""}},"port_forwarding":{"enabled":false,"provider":"","status_file_path":"/tmp/gluetun/forwarded_port","listening_port":0}},"openvpn":{"version":"2.5","user":"","password":"","config_file_path":"","ciphers":null,"auth":"","cert":"","key":"","encrypted_key":"","key_passphrase":"","pia_encryption_preset":"","mssfix":0,"interface":"tun0","process_user":"nonrootuser","verbosity":1,"flags":null},"wireguard":{"private_key":"*************************","pre_shared_key":"","addresses":["10.5.0.2/32"],"allowed_ips":["0.0.0.0/0","::/0"],"interface":"tun0","mtu":1400,"implementation":"auto"}}

---
curl -X PUT --data '{ "status": "stopped"}' http://localhost:8000/v1/vpn/status
{"outcome":"stopped"}
```

### Share your configuration

_No response_

joejose97:
-  Add basic http auth to vpn/settings  & openvpn/settings routes.
  - Get username and password from environment variables.
  - Sets a default username and password.

Partly fixes #2238 

@qdm12 , please provide feedback.
Do we need a default username and password.
I have added basic auth only for the /settings route so that breakage with existing users are minimal.

Edit:
Original PR is updated to use the new auth middleware. Above points are no longer valid.
See #2238 , #2434

### DIFF
--- a/internal/server/middlewares/auth/basic.go
+++ b/internal/server/middlewares/auth/basic.go
@@ -1,0 +1,51 @@
+package auth
+
+import (
+	"crypto/sha256"
+	"crypto/subtle"
+	"net/http"
+	"strings"
+)
+
+type basicAuthMethod struct {
+	userpasshash [32]byte
+}
+
+func newBasicAuthMethod(user string, pass string) *basicAuthMethod {
+	var sb strings.Builder
+
+	sb.WriteString(user)
+	sb.WriteString(pass)
+	var matchbytes = sha256.Sum256([]byte(sb.String()))
+	return &basicAuthMethod{
+		userpasshash: matchbytes,
+	}
+}
+
+// equal returns true if another auth checker is equal.
+// This is used to deduplicate checkers for a particular route.
+func (a *basicAuthMethod) equal(other authorizationChecker) bool {
+	otherBasicMethod, ok := other.(*basicAuthMethod)
+	if !ok {
+		return false
+	}
+	return a.userpasshash == otherBasicMethod.userpasshash
+}
+
+func (a *basicAuthMethod) isAuthorized(r *http.Request) bool {
+	authsuccess := false
+	var inpSb strings.Builder
+	// Get Inputs from http request
+	username, password, ok := r.BasicAuth()
+	if ok {
+		inpSb.WriteString(username)
+		inpSb.WriteString(password)
+		inputhash := sha256.Sum256([]byte(inpSb.String()))
+		authsuccess = (subtle.ConstantTimeCompare(a.userpasshash[:], inputhash[:]) == 1)
+	}
+	if authsuccess {
+		return true
+	}
+
+	return false
+}
--- a/internal/server/middlewares/auth/basic.go
+++ b/internal/server/middlewares/auth/basic.go
@@ -4,21 +4,15 @@ import (
 	"crypto/sha256"
 	"crypto/subtle"
 	"net/http"
-	"strings"
 )
 
 type basicAuthMethod struct {
-	userpasshash [32]byte
+	authDigest [32]byte
 }
 
-func newBasicAuthMethod(user string, pass string) *basicAuthMethod {
-	var sb strings.Builder
-
-	sb.WriteString(user)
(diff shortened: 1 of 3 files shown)


qdm12:
Please see and reply to https://github.com/qdm12/gluetun/issues/2238#issuecomment-2295450851 🙏 
My apologies for not doing this earlier, as @ezekieldas pointed out, this turns out to be rather urgent to fix/do something about it (changed the priority of the issue to urgent)!

qdm12:
Can you point your PR towards the branch `auth-middleware-2238` and inspire from #2437 to implement basic auth with the new 'auth code infrastructure' in place? 🙏 Thanks!

joejose97:
I've updated to use the auth middleware.
Sorry for the delay.

internal/server/middlewares/auth/basic.go
```diff
@@ -0,0 +1,36 @@
...
+	return a.authDigest == otherBasicMethod.authDigest
+}
+
+func (a *basicAuthMethod) isAuthorized(_ http.ResponseWriter, request *http.Request) bool {
+	username, password, ok := request.BasicAuth()
+	if !ok {
+		return false
+	}
+	requestAuthDigest := sha256.Sum256([]byte(username + password))
+	return subtle.ConstantTimeCompare(a.authDigest[:], requestAuthDigest[:]) == 1
```
qdm12, line 35: Why do we use `subtle.ConstantTimeCompare`? Couldn't we just use

```suggestion
	return bytes.Equal(a.authDigest[:], requestAuthDigest[:])
```

qdm12 (approved):
> I've updated to use the auth middleware

Awesome 💯 ❤️ 
FYI I rebased your branch + pushed a small commit to simplify code a bit

> Sorry for the delay.

Please, no pressure at all 😸

internal/server/middlewares/auth/basic.go
joejose97: We can, constant time compare just protects against timing based attacks.
ie Brute forcing can be made easier with the timing information if we use simple equal.
See https://ropesec.com/articles/timing-attacks/

Thanks for the changes. This is my first time using golang.😂

qdm12: Oh awesome, thanks for the teaching moment! 💯

qdm12: Actually with a hash digest, I doubt it matters much. It does matter a hell lot for #2437 so that's great. keeping it just in case in this PR.

joejose97: Oh yeah that's true.

PR #2490, opened 2024-09-23
fix(cyberghost): update cyberghost servers list

y0hnn:
Cyberghost servers list is not up to date. For example, impossible to connect to hosts in Vietnam.
But it seems the Cyberghost API used is not available anymore. For now, I just updated the list with a small script doing a `dig` on the hostnames, and putting every `A` records inside it.
You can check it:
```
const fs = require('fs');
const dns = require('dns');
const path = require('path');

// Read the JSON file
const filePath = path.join(__dirname, 'cg.json');
const data = JSON.parse(fs.readFileSync(filePath, 'utf8'));

// Function to perform DNS lookup and update IPs
const updateIPs = async (hostname) => {
    return new Promise((resolve, reject) => {
        dns.resolve(hostname, 'A', (err, addresses) => {
            if (err) {
                console.error(`Failed to resolve ${hostname}:`, err);
                return resolve([]); // Resolve with an empty array on error
            }
            resolve(addresses);
        });
    });
};

// Function to process each server entry
const processServers = async () => {
    for (let server of data.cyberghost.servers) {
        console.log(`Processing ${server.hostname} for ${server.country}...`);
        const updatedIPs = await updateIPs(server.hostname);
        if (updatedIPs.length > 0) {
            server.ips = updatedIPs;
            console.log(`Updated IPs for ${server.hostname}:`, updatedIPs);
        } else {
            console.log(`No IPs found for ${server.hostname}.`);
        }
    }

    // Write the updated JSON back to the file
    fs.writeFileSync(filePath, JSON.stringify(data, null, 2), 'utf8');
    console.log(`Updated JSON saved to ${filePath}`);
};

// Start the process
processServers().catch(error => {
    console.error('Error processing servers:', error);
});
```


qdm12:
> But it seems the Cyberghost API used is not available anymore.

Not sure what API you are talking about?

For now, Gluetun goes through all possible combinations of groupid and country code, and tries to resolve each.
See https://github.com/qdm12/gluetun/blob/7ebbaf4351404b0ae4ed3ff67985637385338094/internal/provider/cyberghost/updater/hosttoserver.go#L13

The existing code does pretty much what you wrote in javascript, just run it 😉 
See [the wiki](https://github.com/qdm12/gluetun-wiki/blob/main/setup/servers.md#update-using-the-command-line)

PR #2502, opened 2024-10-01
feat(publicip): add Cloudflare support

jjlin:
Note that the Cloudflare API does not seem to support returning info for arbitrary IP addresses (only the caller's public IP), so it can't be used with calls to FetchMultiInfo, but since this is only used in the context of a couple of VPN providers currently, it may not be a significant limitation.

### DIFF
--- a/internal/publicip/api/cloudflare.go
+++ b/internal/publicip/api/cloudflare.go
@@ -1,0 +1,91 @@
+package api
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/netip"
+	"strings"
+
+	"github.com/qdm12/gluetun/internal/constants"
+	"github.com/qdm12/gluetun/internal/models"
+)
+
+type cloudflare struct {
+	client *http.Client
+}
+
+func newCloudflare(client *http.Client) *cloudflare {
+	return &cloudflare{
+		client: client,
+	}
+}
+
+// FetchInfo obtains information on the public IP address of the machine,
+// and returns an error if the `ip` argument is set since the Cloudflare API
+// can only be used to provide details about the current machine public IP.
+func (c *cloudflare) FetchInfo(ctx context.Context, ip netip.Addr) (
+	result models.PublicIP, err error) {
+	url := "https://speed.cloudflare.com/meta"
+	if ip.IsValid() {
+		return result, fmt.Errorf("%w: cloudflare cannot provide information on the arbitrary IP address %s",
+			ErrServiceLimited, ip)
+	}
+
+	request, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+	if err != nil {
+		return result, err
+	}
+
+	response, err := c.client.Do(request)
+	if err != nil {
+		return result, err
+	}
+	defer response.Body.Close()
+
+	switch response.StatusCode {
+	case http.StatusOK:
+	case http.StatusTooManyRequests:
+		return result, fmt.Errorf("%w from %s: %d %s",
+			ErrTooManyRequests, url, response.StatusCode, response.Status)
+	default:
+		return result, fmt.Errorf("%w from %s: %d %s",
+			ErrBadHTTPStatus, url, response.StatusCode, response.Status)
+	}
+
+	decoder := json.NewDecoder(response.Body)
+	var data struct {
+		Hostname       string     `json:"hostname,omitempty"`
+		ClientIP       netip.Addr `json:"clientIp,omitempty"`
+		ASOrganization string     `json:"asOrganization,omitempty"`
+		Country        string     `json:"country,omitempty"`
+		City           string     `json:"city,omitempty"`
+		Region         string     `json:"region,omitempty"`
+		PostalCode     string     `json:"postalCode,omitempty"`
+		Latitude       string     `json:"latitude,omitempty"`
+		Longitude      string     `json:"longitude,omitempty"`
+	}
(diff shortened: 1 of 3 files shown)


internal/publicip/api/cloudflare.go
```diff
@@ -0,0 +1,96 @@
...
+	return &cloudflare{
+		client: client,
+	}
+}
+
+// FetchInfo obtains information on the IP address provided using the
+// Cloudflare API. If the IP is the zero value, the public IP address
+// of the machine is used as the IP. Since the Cloudflare API can only
+// be used to provide details about the public IP, providing a non-zero
+// IP results in an error indicating that the IP is not supported.
```
qdm12, line 29:
```suggestion
// FetchInfo obtains information on the public IP address of the machine,
// and returns an error if the `ip` argument is set since the Cloudflare API 
// can only be used to provide details about the current machine public IP.
```

internal/publicip/api/errors.go
```diff
@@ -6,4 +6,5 @@ var (
 	ErrTokenNotValid   = errors.New("token is not valid")
 	ErrTooManyRequests = errors.New("too many requests sent for this month")
 	ErrBadHTTPStatus   = errors.New("bad HTTP status received")
+	ErrUnsupportedIP   = errors.New("service does not support arbitrary IP addresses")
```
qdm12, line 9: nit if I may, perhaps this is more appropriate:

```go
ErrServiceLimited = errors.New("service is limited")
```

and then wrap it in the cloudflare code with `fmt.Errorf("%w: cloudflare cannot provide information on the arbitrary IP address %s", ErrServiceLimited, ip)` - that way that 'common' error `ErrServiceLimited` can also be used elsewhere in the future. It also feels a bit wrong that it's "unsupported ip", it would be supported if it would be our IP address, so it's more of a service limitation 🤔

internal/publicip/api/cloudflare.go
```diff
@@ -0,0 +1,96 @@
...
+
+	response, err := cf.client.Do(request)
+	if err != nil {
+		return result, err
+	}
+	defer response.Body.Close()
+
+	switch response.StatusCode {
+	case http.StatusOK:
+	case http.StatusTooManyRequests, http.StatusForbidden:
```
qdm12, line 50:
```suggestion
	case http.StatusTooManyRequests:
```

internal/publicip/api/cloudflare.go
```diff
@@ -0,0 +1,96 @@
...
+
+	decoder := json.NewDecoder(response.Body)
+	var data struct {
+		Hostname       string     `json:"hostname,omitempty"`
+		ClientIP       netip.Addr `json:"clientIp,omitempty"`
+		HTTPProtocol   string     `json:"httpProtocol,omitempty"`
+		ASN            uint32     `json:"asn,omitempty"`
+		ASOrganization string     `json:"asOrganization,omitempty"`
+		// Colo is the closest airport code
+		Colo       string `json:"colo,omitempty"`
```
qdm12, line 66: nit unused so let's remove

```suggestion
```

internal/publicip/api/cloudflare.go
```diff
@@ -0,0 +1,96 @@
...
+		return result, fmt.Errorf("%w from %s: %d %s",
+			ErrBadHTTPStatus, url, response.StatusCode, response.Status)
+	}
+
+	decoder := json.NewDecoder(response.Body)
+	var data struct {
+		Hostname       string     `json:"hostname,omitempty"`
+		ClientIP       netip.Addr `json:"clientIp,omitempty"`
+		HTTPProtocol   string     `json:"httpProtocol,omitempty"`
+		ASN            uint32     `json:"asn,omitempty"`
```
qdm12, line 63: nit unused so let's remove

```suggestion
```

qdm12 (commented):
Nice, still a useful addition I think! As long as it's not the default, it should work fine 😉

internal/publicip/api/cloudflare.go
jjlin: done

internal/publicip/api/cloudflare.go
jjlin: done

internal/publicip/api/cloudflare.go
jjlin: done

internal/publicip/api/cloudflare.go
jjlin: done

internal/publicip/api/cloudflare.go
```diff
@@ -0,0 +1,96 @@
...
+		client: client,
+	}
+}
+
+// FetchInfo obtains information on the IP address provided using the
+// Cloudflare API. If the IP is the zero value, the public IP address
+// of the machine is used as the IP. Since the Cloudflare API can only
+// be used to provide details about the public IP, providing a non-zero
+// IP results in an error indicating that the IP is not supported.
+func (cf *cloudflare) FetchInfo(ctx context.Context, ip netip.Addr) (
```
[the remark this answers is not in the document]
jjlin: I think `cf` is the more customary abbreviation for Cloudflare, and I'm not aware of any widespread convention to strictly use single letters for receiver names (just "short" names, e.g. https://google.github.io/styleguide/go/decisions.html#receiver-names), but if this is your local convention, I'm happy to go with it.

qdm12 (approved):

qdm12:
Awesome, thanks!! 💯

Issue #2506, opened 2024-10-04
Bug: No connection with Wireguard on IPv6-only host


arseneyr:
### Is this urgent?

No

### Host OS

Amazon Linux 2023

### CPU arch

aarch64

### VPN service provider

Mullvad

### What are you using to run the container

docker run

### What is the version of Gluetun

Running version latest built on 2024-09-29T18:12:41.313Z (commit 7ebbaf4)

### What's the problem 🤔

I'm getting no Wireguard connection when running a Docker container on an IPv6-only host. I've narrowed it down to the following two issues:

1. Gluetun is selecting IPv4 addresses from its server list and fails to connect until it randomly selects an IPv6 address. It would be nice to have a flag for Gluetun so that Wireguard (and maybe OpenVPN too?) will only try servers via their IPv6 address.

2. There's a bug in allowing NDP multicast packets: 

    https://github.com/qdm12/gluetun/blob/7ebbaf4351404b0ae4ed3ff67985637385338094/internal/firewall/iptables.go#L191

    The destination subnet should be `ff02::1:ff00:0/104`. I have confirmed this change restores Wireguard connectivity (when Gluetun selects an IPv6 address as mentioned above). Even with this fix, it seems like the `ip6tables` rules are still blocking some NDP packets from the link-local address:

    ```
    Oct  3 23:31:05  IN= OUT=eth0 MAC= SRC=fe80::42:acff:fe11:2 DST=fc00::1 LEN=72 TC=0 HOPLIMIT=255 
    FLOWLBL=0 PROTO=ICMPv6 TYPE=135 CODE=0 MARK=0x0 
    Oct  3 23:31:06  IN= OUT=eth0 MAC= SRC=fe80::42:acff:fe11:2 DST=fc00::1 LEN=72 TC=0 HOPLIMIT=255 
    FLOWLBL=0 PROTO=ICMPv6 TYPE=135 CODE=0 MARK=0x0 
    Oct  3 23:31:07  IN= OUT=eth0 MAC= SRC=fe80::42:acff:fe11:2 DST=fc00::1 LEN=72 TC=0 HOPLIMIT=255 
    FLOWLBL=0 PROTO=ICMPv6 TYPE=135 CODE=0 MARK=0x0 
    Oct  3 23:31:11  IN= OUT=eth0 MAC= SRC=fe80::42:acff:fe11:2 DST=fe80::1 LEN=72 TC=0 HOPLIMIT=255 
[...]

Issue #1563, opened 2023-05-11
Feature request: Env variable to disable IPv6 auto-enable


buroa:
### What's the feature 🧐

Allow us to complete disable IPv6, instead of automatically checking if our system supports it.

https://github.com/qdm12/gluetun/blob/master/internal/netlink/ipv6.go#L9
https://github.com/qdm12/gluetun/blob/master/internal/provider/utils/wireguard.go#L12
https://github.com/qdm12/gluetun/blob/master/internal/wireguard/address.go#L14
https://github.com/qdm12/gluetun/blob/master/internal/configuration/sources/env/wireguard.go#L12

Something like a env variable that goes into the user settings. Looks like the wireguard is already checking that setting so as long as we can pass it as an override, this will work for anyone that has the same setup.

qdm12:
Why do you need this? What's the use case?

buroa:
@qdm12 If I use `mullvad` for example, it attempts to connect to a lot of IPv6-enabled wireguard servers -- but I don't actually have IPv6 enabled in my homelab -- so it shouldn't even attempt this.

qdm12:
Ok that's interesting, maybe a better way to address this would be to fix the IPv6 detection mechanism 😉 

The mechanism right now is https://github.com/qdm12/private-internet-access-docker/blob/3100cc1e5ee7d6523cc02e1c874a271713c13395/internal/netlink/ipv6.go#L9 and it basically goes through the routes of each links, and if one route is IPv6, it assumes IPv6 is supported. What's the output of

```
docker run -it --rm alpine ip route
docker run -it --rm alpine ip -6 route
```

buroa:
> ip -6 route

```
ip r && ip -6 route
default via 10.244.1.220 dev eth0 mtu 9000
10.244.1.220 dev eth0 scope link
fe80::/64 dev eth0 proto kernel metric 256 pref medium
```

qdm12:
@qriff interesting points, I am thinking to add an ipv6 dummy route to test for ipv6 support, instead of checking existing routes for IPv6 source or destination.

Would you have any suggestion of what `ip6 route add ...` you would use to not affect existing routing tables (not a routing expert to be honest) 🤔

samip5:
This also happens with airvpn as I have blocked IPv6 outbound, and it stil tries to connect via v6.

For reference:
```
# ip r && ip -6 r 
default via 10.244.4.77 dev eth0 mtu 1500 
[...]

PR #2523, opened 2024-10-14
feat(netlink): detect ipv6 support level

qdm12:
- 'supported' if one ipv6 route is found that is not loopback and not a default route
- 'internet' if one default ipv6 route is found

PR #2557, opened 2024-11-01
Allow skipping IPv6 server IPs via VPN_IPV6_SERVER

diamkil:
This code allows skipping IPv6 server IPs by setting VPN_IPV6_SERVER=off. This is useful when IPv6 support is enabled in docker without a routable IPv6, as most VPN providers give an IPv6 but not all ISPs give an IPv6.

This solves #1563 even if not automatic
A modified version of this code could solve #2506


qdm12:
Thanks! However, I prefer to have a reliable IPv6 detection mechanism without an additional option, see #2523 which is just waiting for testers.

Can you try image tag `:pr-2523` with `LOG_LEVEL=debug` and check how it does? More details on how to run it https://github.com/qdm12/gluetun/issues/2506#issuecomment-2411786415

qdm12:
Closed in favor of now merged #2523

PR #2564, opened 2024-11-03
DNS debug logs

qdm12:
For #2533 debugging purposes only.

- log out round trip duration for failed/timed out DNS queries
- log out all DNS requests handled
- log out all DNS requests handled successfully


qdm12:
No longer needed, see #2533 resolution.

PR #2574, opened 2024-11-08
fix(firewall): pass -n to iptables in path test

jfroy:
Performing reverse dns lookups during setup can take a long time, especially if they timeout because gluetun is restarting and the requests are blocked (and thus only return after a long timeout).


qdm12 (approved):
Totally great simple tiny and effective PR, congrats! 🎖️ !

PR #2611, opened 2024-12-01
feat(extras/scripts): update Transmission peer port

juanlufont:
The script `transmission-port-update.sh` updates the _peer port_ used by the torrent client Transmission using its API. The peer port is the port used and announced by Transmission to accept incoming connections. You can use this script to set a VPN forwarded port as the current listening port for Transmission.

The script uses only shell scripting and can run out of the box within the current `gluetun` docker image, it only relies on the command `wget`, which is already included by default in the image.


ku1ik:
I've used this script and it worked on the first try. Thank you!

extras/scripts/transmission-port-update.sh
```diff
@@ -0,0 +1,121 @@
...
+if [ -z "${PORT}" ]; then
+    echo "ERROR: No PORT provided!"
+    exit 1
+fi
+
+if [ "${_USECRED}" ]; then
+    # make sure username AND password were provided
+    if [ -z "${USERNAME}" ] || [ -z "${PASSWORD}" ]; then
+        echo "ERROR: Username or password not provided."
+        exit 1
```
qdm12, line 68: nit you can split this in two if conditions, to have a more precise error message, either username not provided or password not provided

extras/scripts/transmission-port-update.sh
```diff
@@ -0,0 +1,121 @@
...
+    else
```
qdm12, line 69: nit no need for the else since the if runs `exit 1` (aka early return)

extras/scripts/transmission-port-update.sh
```diff
@@ -0,0 +1,121 @@
...
+        "$RPC_URL" \
+        -O -
+)
+
+# check string returned by wget
+SUCCESS='{"arguments":{},"result":"success"}'
+if [ "$RES" = "$SUCCESS" ]; then
+    echo "Success! Peer port updated to ${PORT}"
+    exit 0
+else
```
qdm12, line 118: nit no need for the else given the early return with `exit 0` above

also - related to the change proposed - it would be nice to have the error path in an if condition, and the success path at indentation zero below it, so more things can be added after a success in the future without change.

extras/scripts/transmission-port-update.sh
```diff
@@ -0,0 +1,121 @@
+#!/bin/sh
```
qdm12, line 1: This script is meant to be run at the port forwarding setup completed event right?
If so, see [this page](https://github.com/qdm12/gluetun-wiki/blob/ee041f335fc0a2e27d4a3c59e37a8957f6dae23f/setup/advanced/vpn-port-forwarding.md#custom-port-forwarding-updown-command)

- we need to use `PORTS` instead of `PORT`, and process possibly multiple ports (or just take the first one from a csv of ports)
- let's copy your script in the Docker image. Add a `COPY extras/scripts/transmission-port-update.sh /somepathofyourchoosing` right below the `HEALTHCHECK ` line
- create a PR to the gluetun wiki on that port forwarding page mentioned above, to tell users about that script they can use - up to you, a new section might be appropriate

Thanks!

qdm12 (changes requested):

ku1ik:
Just reporting that I've been using this for over a month and it's been working flawlessly 👌

extras/scripts/transmission-port-update.sh
juanlufont: Thanks a lot for all the feedback, I will work on the script in the next days :+1:

extras/scripts/transmission-port-update.sh
juanlufont: Done

extras/scripts/transmission-port-update.sh
juanlufont: Done

extras/scripts/transmission-port-update.sh
juanlufont: Transmission expects as peer-port value a single integer. To support `{{PORTS}}` I am splitting by commas (if any) and keeping only the first value.

juanlufont: I have updated the Dockerfile

juanlufont: I created a PR for the Gluetun wiki: https://github.com/qdm12/gluetun-wiki/pull/132

extras/scripts/transmission-port-update.sh
```diff
@@ -0,0 +1,132 @@
...
+        echo "ERROR: Transmission RPC Username not provided."
+        exit 1
+    fi
+    if [ -z "${PASSWORD}" ]; then
+        echo "ERROR: Transmission RPC Password not provided."
+        exit 1
+    fi
+    # basic auth options, --auth-no-challenge avoids 409 Conflict
+    WGET_OPTS="
+            ${WGET_OPTS}
```
qdm12, line 82:
```suggestion
            --quiet
```

extras/scripts/transmission-port-update.sh
```diff
@@ -0,0 +1,132 @@
...
+        --header="X-Transmission-Session-Id: $SESSION_ID" \
+        --post-data="${PAYLOAD}" \
+        "$RPC_URL" \
+        -O -
+)
+
+# check string returned by wget
+SUCCESS='{"arguments":{},"result":"success"}'
+if [ "$RES" != "$SUCCESS" ]; then
+    echo "ERROR: Could not update Transmission peer-port"
```
qdm12, line 127:
```suggestion
    echo "ERROR: Could not update Transmission peer-port: ${RES}"
```

extras/scripts/transmission-port-update.sh
```diff
@@ -0,0 +1,132 @@
...
+    exit 1
+fi
+
+echo "Success! Transmission peer-port updated to ${PORT}"
+exit 0
```
qdm12, line 132: no need for the final exit 😉 (afaik)

```suggestion
```

qdm12 (commented):
Thanks for the review of the PR (and the wiki PR 👍 ) Just a few minor things to do if possible 🙏 thanks!

extras/scripts/transmission-port-update.sh
```diff
@@ -0,0 +1,132 @@
+#!/bin/sh
+
+# Description: This script updates the peer port for the Transmission torrent
+# client using its RPC API.
+# Author: Juan Luis Font
+
+# default values
+DEFAULT_URL="http://localhost:9091/transmission/rpc"
+WGET_OPTS="--quiet"
```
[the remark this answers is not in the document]
juanlufont: I have moved `--quiet` flag out of `WGET_OPTS` but I am keeping the variable to store there, if needed, the API credentials and some extra flags they require. Is this ok?

iridial:
I did some tests in k3s with gluetun `v3.40` running as `initContainer` for transmission, and using `VPN_PORT_FORWARDING_UP_COMMAND` as trigger to call the script using: `'/bin/sh -c "/scripts/transmission-port-update.sh --port {{PORTS}}"'`.
The transmission container takes more time to be available to receive the port forward update request, and this leads the initial call of the script to fail.
I suggest to include an initial check if transmission api service is available, and if not, retry once more after configurable timeout with `/bin/sleep <value>`.

sboger:
I'm just using this with great success...

`      - VPN_PORT_FORWARDING_UP_COMMAND=/bin/sh -c 'apk add transmission-remote && transmission-remote localhost -p {{PORTS}} && transmission-remote localhost -t all --reannounce'
`

extras/scripts/transmission-port-update.sh
```diff
@@ -0,0 +1,133 @@
...
+    echo "Options:"
+    echo "  -h, --help       Show this help message and exit."
+    echo "  -u, --user USER  Specify the Transmission RPC user name."
+    echo "                   (Omit if not required)"
+    echo "  -p, --pass PASS  Specify the Transmission RPC password."
+    echo "                   (Omit if not required)"
+    echo "  -P, --port PORT  Specify the Transmission Peer Port."
+    echo "                   If PORT is a comma-separated list of ports,"
+    echo "                   only use the first one."
+    echo "                   REQUIRED"
```
qdm12, line 25: Let's have this take multiple ports as a comma separated list (`{{PORTS}}` value injected by Gluetun).
If transmission supports only a single port, use `$(echo {{PORTS}} | cut -d, -f1)` to take only the first port in case there are multiple

qdm12 (changes requested):

qdm12:
> The transmission container takes more time to be available to receive the port forward update request, and this leads the initial call of the script to fail.

Let's add `--retry-connrefused -t 3` to the wget command

extras/scripts/transmission-port-update.sh
```diff
@@ -0,0 +1,133 @@
...
+# shellcheck disable=SC2086
+SESSION_ID=$(
+    wget \
+        --quiet \
+        ${WGET_OPTS} \
+        --server-response \
+        "$RPC_URL" 2>&1 |
+        grep 'X-Transmission-Session-Id:' |
+        awk '{print $2}'
+)
```
iridial, line 102: consider wrapping this in a loop, which fails after some retries:
```
i=1
while [ "$i" -le $MAX_RETRIES ]; do
    # get the X-Transmission-Session-Id
    # shellcheck disable=SC2086
    SESSION_ID=$(
        wget \
            --quiet \
            ${WGET_OPTS} \
            --server-response \
            "$RPC_URL" 2>&1 |
            grep 'X-Transmission-Session-Id:' |
            awk '{print $2}'
    )
    
    if [ -n "$SESSION_ID" ]; then
        echo "Transmission available. SESSION_ID: $SESSION_ID"
        break
    else
        echo "Attempt $i of $MAX_RETRIES failed, waiting for next attempt..."
        sleep 5
    fi

    i=$(( i + 1 ))
done
```

extras/scripts/transmission-port-update.sh
```diff
@@ -0,0 +1,133 @@
...
+
+# check string returned by wget
+SUCCESS='{"arguments":{},"result":"success"}'
+if [ "$RES" != "$SUCCESS" ]; then
+    echo "ERROR: Could not update Transmission peer-port: ${RES}"
+    exit 1
+fi
+
+echo "Success! Transmission peer-port updated to ${PORT}"
```
qdm12, line 132: nit

```suggestion
echo "Transmission peer-port updated to ${PORT}"
```

extras/scripts/transmission-port-update.sh
qdm12: Never mind, fcdba0a3cc92e34d93563850eeb069c277a24289 adds support for a `{{PORT}}` variable which is the first one.

PR #2700, opened 2025-02-13
feat(extras/scripts): update qBittorrent peer port with optional authentication

astappiev:
Inspired by https://github.com/qdm12/gluetun/pull/2611 and talks in https://github.com/qdm12/gluetun/discussions/1555 created a script to update peer-port of qBittorrent.

Upd: added option to authenticate.


astappiev:
The only downside of these scripts is that they don't cover a scenario when a client is started after gluetun.

astappiev:
I think it should be implemented as another script (as it is not client dependent), or even in the gluetun core.

extras/scripts/qbittorrent-port-update.sh
```diff
@@ -0,0 +1,124 @@
...
+
+build_default_url() {
+    port="${1:-$WEBUI_PORT}"
+    echo "http://127.0.0.1:${port}/api"
+}
+
+# default values
+WEBUI_PORT="8080"
+DEFAULT_URL=$(build_default_url)
+WGET_OPTS="--retry-connrefused --tries=5"
```
qdm12, line 16: Why is this needed?

extras/scripts/qbittorrent-port-update.sh
```diff
@@ -0,0 +1,124 @@
+#!/bin/sh
+
+# Description: This script updates the peer-port for the qBittorrent torrent client using its WebUI API.
+# Note: For this to work, "Bypass authentication for clients on localhost" should be enabled
+# in the WebUI settings (json key bypass_local_auth).
```
qdm12: Since username and password are supported, i think this can be removed?

extras/scripts/qbittorrent-port-update.sh
```diff
@@ -0,0 +1,124 @@
...
+    echo "                         (Omit if not required)"
+    echo "  -P, --port PORT        Specify the qBittorrent peer-port."
+    echo "                         REQUIRED"
+    echo "  -W, --webui-port PORT  Specify the qBittorrent WebUI Port."
+    echo "                         Default: ${WEBUI_PORT}"
+    echo "  -U, --url URL          Specify the qBittorrent API URL."
+    echo "                         DEFAULT: ${DEFAULT_URL}"
+    echo "                         Overrides --webui-port option."
+    echo "Example:"
+    echo "  $0 -u admin -p **** --port 40409"
```
qdm12: If I may propose, can we just keep the long form flags?
Maybe rather opinionated, but short flags are just easy to forget what they mean, and would make later debugging more difficult than explicit long form flags.

extras/scripts/qbittorrent-port-update.sh
```diff
@@ -0,0 +1,124 @@
+#!/bin/sh
+
+# Description: This script updates the peer-port for the qBittorrent torrent client using its WebUI API.
```
qdm12: See https://github.com/qdm12/gluetun-wiki/commit/55ffd925ac372b866035853e9dadb600b4b54067

We should have a "down" mode to change the port to something else.
Or, maybe better, have it set it to some random port i.e. `9876`, and then to the actual port we want, such that there is only an up command to set.

qdm12:
Also my apologies for the massive delay, I'm back in the coding world so this should be merged soon.

qdm12:
@AndaPlays 

> The only thing I would appreciate is if we could add an option for Nmap to check whether the port is actually open

Good idea, and I agree with @astappiev it could be in the Gluetun core. What nmap command would you usually run to check a port is open? I'll look at having this run on every port forwarding setup in Go.

extras/scripts/qbittorrent-port-update.sh
```diff
@@ -0,0 +1,124 @@
...
+    echo ""
+    echo "Update qBittorrent peer-port via API"
+    echo ""
+    echo "Options:"
+    echo "  -h, --help             Show this help message and exit."
+    echo "  -u, --user USER        Specify the qBittorrent username."
+    echo "                         (Omit if not required)"
+    echo "  -p, --pass PASS        Specify the qBittorrent password."
+    echo "                         (Omit if not required)"
+    echo "  -P, --port PORT        Specify the qBittorrent peer-port."
```
qdm12: this should accept a comma separated list of ports from Gluetun's `{{PORTS}}` to simplify usage of this script.
And then it should use `$(echo {{PORTS}} | cut -d, -f1)` to take only the first port just in case, since qbitorrent supports only one port.

qdm12 (changes requested):

extras/scripts/qbittorrent-port-update.sh
```diff
@@ -0,0 +1,124 @@
...
+        exit 1
+    fi
+
+    # set cookie for future requests
+    WGET_OPTS="${WGET_OPTS} --header=Cookie:$cookie"
+fi
+
+# update peer host via API
+wget ${WGET_OPTS} -qO- --post-data="json={\"random_port\":false}" "$PREF_URL/v2/app/setPreferences"
+wget ${WGET_OPTS} -qO- --post-data="json={\"listen_port\":$PORT}" "$PREF_URL/v2/app/setPreferences"
```
qdm12: You can combine these together, and also set upnp to false


```suggestion
wget ${WGET_OPTS} -qO- --post-data="json={\"random_port\":false,\"upnp\":false,\"listen_port\":$PORT}" "$PREF_URL/v2/app/setPreferences"
```

extras/scripts/qbittorrent-port-update.sh
astappiev: Not sure if I got it right, does it mean we should use it like this `bittorrent-port-update.sh --port {{PORTS}}`?

E.g. via `VPN_PORT_FORWARDING_UP_COMMAND=/bin/sh -c "/scripts/qbittorrent-port-update.sh --port {{PORTS}} --webui-port 9081"`

If so, it's already supported, this is exactly how I used it since beggining.

extras/scripts/qbittorrent-port-update.sh
astappiev: It might require some time until qbittorrent is up and listening when I start my compose stack

extras/scripts/qbittorrent-port-update.sh
astappiev: I'm not sure what's best here; the provider I'm using is always giving a unique port.
For now, I set it to `0` as a "safe" port to reset the setting. If we use any "random" port, there is a minor chance that it might match with an actual port.

AndaPlays:
> @AndaPlays
> 
> > The only thing I would appreciate is if we could add an option for Nmap to check whether the port is actually open
> 
> Good idea, and I agree with @astappiev it could be in the Gluetun core. What nmap command would you usually run to check a port is open? I'll look at having this run on every port forwarding setup in Go.

@qdm12  In my port manager I use nmap in a very simple way to verify that the forwarded TCP port is actually reachable from outside. The logic is here in my [script](https://github.com/AndaPlays/gluetun-qbittorrent-port-manager/blob/main/start.sh#L47). In short, I call nmap -Pn -p <port> <public_ip> (with a small retry loop) and parse the output to check whether the port is reported as open. -Pn is important to me because host discovery (ping etc.) is often blocked, and I only care whether the specific forwarded port is open vs. closed/filtered.

extras/scripts/qbittorrent-port-update.sh
qdm12: Correct, but {{PORTS}} can have multiple ports so for convenience and a shorter command, let's cut and use the first port within the script. The flag should be renamed to --ports and indicate that despite the name it only picks the first port

qdm12:
@AndaPlays for this nmap command to succeed, I suppose you do need a program listening on the forwarded port right? What I'm thinking is to, when setting up port forwarding:

1. start an ephemeral tcp server on the forwarded port / if the port is used already then don't that's fine too (udp is more complicated and not widely used by users so we'll skip that)
2. check public-ip:port with a tcp dial (same as the nmap command)
3. Report to debug logs it is open and finish the port forwarding setup process (I'm a maniac of logging problems only and avoid logging if things work correctly)

I'm going to mark our comments as off topic, would you please open an issue for this? Thanks!!

extras/scripts/qbittorrent-port-update.sh
astappiev: I think the other way around, let's keep it named as `--port` to indicate that only one of them is used.

The logic to split and keep the first was already there
https://github.com/astappiev/gluetun/blob/839e318c9c30676c8ecd9f042c524e87a1634f32/extras/scripts/qbittorrent-port-update.sh#L60

extras/scripts/qbittorrent-port-update.sh
qdm12: Can you add a comment for this? :wink:

extras/scripts/qbittorrent-port-update.sh
qdm12: setting it to 0 is a good idea, although that's not done here. Maybe we could do it right before calling the API to set to the desired port?

extras/scripts/qbittorrent-port-update.sh
qdm12: Actually.... 
Added support for `{{PORT}}` variable in fcdba0a3cc92e34d93563850eeb069c277a24289 which takes the first of the ports. No reason to have it supported in scripts!

monstermuffin:
Hey guys, I know it's not a native solution but this use case is exactly why I made [qSticky.](https://github.com/monstermuffin/qSticky/) Been working very well and it doesn't matter what order things come up as it's a middleman between the two applications.

extras/scripts/qbittorrent-port-update.sh
```diff
@@ -0,0 +1,131 @@
...
+        echo "ERROR: Could not authenticate with qBittorrent."
+        exit 1
+    fi
+
+    # set cookie for future requests
+    WGET_OPTS="${WGET_OPTS} --header=Cookie:$cookie"
+fi
+
+# update peer host via API, 0 is a dummy port, required due to https://github.com/qdm12/gluetun-wiki/pull/147
+wget ${WGET_OPTS} -qO- --post-data="json={\"random_port\":false,\"upnp\":false,\"listen_port\":0}" "$PREF_URL/v2/app/setPreferences"
```
qdm12: as per https://github.com/qdm12/gluetun/issues/2997#issuecomment-3566749335
let's add


```suggestion
wget ${WGET_OPTS} -qO- --post-data="json={\"random_port\":false,\"upnp\":false,\"listen_port\":0,\"current_network_interface\":\"\",\"current_interface_address\":\"\"}" "$PREF_URL/v2/app/setPreferences"
```

extras/scripts/qbittorrent-port-update.sh
```diff
@@ -0,0 +1,131 @@
...
+if [ $? -ne 0 ]; then
+    echo "ERROR: Could not update qBittorrent peer-port (first call failed)."
+    exit 1
+fi
+
+# second call to set the actual port
+wget ${WGET_OPTS} -qO- --post-data="json={\"listen_port\":$PORT}" "$PREF_URL/v2/app/setPreferences"
```
qdm12:
```suggestion
wget ${WGET_OPTS} -qO- --post-data="json={\"listen_port\":$PORT,\"current_network_interface\":\"$VPN_INTERFACE\",\"current_interface_address\":\"0.0.0.0\"}" "$PREF_URL/v2/app/setPreferences"
```

and I'll add a `{{VPN_INTERFACE}}` to the template variable available to the command.

extras/scripts/qbittorrent-port-update.sh
astappiev: I updated the example to use `--port {{PORT}}`, but left `PORT=$(echo "$2" | cut -d',' -f1)` in to keep it more foolproof

extras/scripts/qbittorrent-port-update.sh
astappiev: I have added that, but maybe let's use `lo` as a temporal interface; it feels safer.

extras/scripts/qbittorrent-port-update.sh
astappiev: Added, but I'm concerned by the `0.0.0.0` as the default address.
It means we restrict to never use IPv6 (I don't know if there are any IPv6 VPNs)

qdm12: There are (airvpn, mullvad) AFAIK; I copied this from a comment of someone; any idea if we can make it for all ipv4 and ipv6? What's the default? Maybe not touching that one would be a better idea.

extras/scripts/qbittorrent-port-update.sh
qdm12: Great idea! Updated the wiki with that suggestion.

extras/scripts/qbittorrent-port-update.sh
qdm12: New template variable `{{VPN_INTERFACE}}` is available, so we can use it for the current network interface json field.

extras/scripts/qbittorrent-port-update.sh
```diff
@@ -0,0 +1,145 @@
+#!/bin/sh
+
+# Description: This script updates the peer-port for the qBittorrent torrent client using its WebUI API.
+#
+# How to use:
+# 1. (Optional) Disable authentication for localhost clients in qBittorrent WebUI settings ("Bypass authentication for clients on localhost" or `bypass_local_auth` in json).
+# 2. Set the environment variable:
+# VPN_PORT_FORWARDING_UP_COMMAND=/bin/sh -c "/scripts/qbittorrent-port-update.sh --port {{PORT}} --webui-port 9081"
```
qdm12:
```suggestion
# VPN_PORT_FORWARDING_UP_COMMAND=/bin/sh -c "/scripts/qbittorrent-port-update.sh --port {{PORT}} --iface {{VPN_INTERFACE}} --webui-port 9081"
```

extras/scripts/qbittorrent-port-update.sh
astappiev: I have played around with this setting in qBitttorrent, so far there are the following options available:
- `""` (empty) - All addresses
- `0.0.0.0` - All IPv4 addresses
- `::` - All IPv6 addresses
- `10.2.0.2` - wireguard network, `docker inspect gluetun` finds this env variable `"WIREGUARD_ADDRESSES=10.2.0.2/32"`
image


Following this, the ideal solution would be to use `10.2.0.2`, but how to detect it automatically?

astappiev: Ah okay, I have found that I set this manually `WIREGUARD_ADDRESSES` in configuration. Which means, we can default to `""` and users can specify `--addr` argument with their env variable (which which the netmask have to be removed)

extras/scripts/qbittorrent-port-update.sh
```diff
@@ -0,0 +1,183 @@
...
+        exit 1
+    fi
+    if [ -z "${PASSWORD}" ]; then
+        echo "ERROR: qBittorrent password not provided"
+        exit 1
+    fi
+
+    cookie=$(wget ${WGET_OPTS} -qO- \
+        --header "Referer: ${WEBUI_URL}" \
+        --post-data "username=${USERNAME}&password=${PASSWORD}" \
```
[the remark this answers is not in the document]
qdm12: Ugliest most retarded fix ever, no thanks AI

extras/scripts/qbittorrent-port-update.sh
```diff
@@ -0,0 +1,183 @@
...
+        grep -i "set-cookie:" | \
+        sed 's/.*set-cookie: //I;s/;.*//')
+
+    if [ -z "${cookie}" ]; then
+        echo "ERROR: Failed to authenticate with qBittorrent. Check username/password or verify WebUI is accessible"
+        exit 1
+    fi
+
+    # set cookie for future requests
+    WGET_OPTS="${WGET_OPTS} --header=Cookie:$cookie"
```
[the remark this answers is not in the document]
qdm12: Is this still a problem now that is uses a cookie jar?

extras/scripts/qbittorrent-port-update.sh
```diff
@@ -0,0 +1,183 @@
...
+        shift 2
+        ;;
+    --url)
+        if [ $# -lt 2 ] || [ -z "$2" ]; then
+            echo "Error: --url requires a non-empty argument."
+            usage
+            exit 1
+        fi
+        WEBUI_URL="$2"
+        shift 2
```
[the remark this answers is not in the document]
qdm12: @copilot apply changes based on [this feedback](https://github.com/qdm12/gluetun/pull/2700#discussion_r2879764069)

PR #2775, opened 2025-04-11
hostname updated for expressvpn servers

stauffenberg2020:
ExpressVPN has updated the server hostnames.


qdm12:
This is done in https://github.com/qdm12/gluetun/pull/2888 thanks for taking the time to contribute still though! 👍

PR #2808, opened 2025-05-19
Fix FastestVPN NL based servers.json

FlashyDuck:
FastestVPN changed IP's NL based servers.


qdm12:
Please change the updater code in https://github.com/qdm12/gluetun/tree/master/internal/provider/fastestvpn/updater if the automated updater doesn't work as expected. In case you didn't know, you can update servers automatically and should not be changing servers.json manually.

qdm12:
Closing this due to inactivity, and assuming the auto updater does still work for fastestvpn.

PR #2821, opened 2025-06-06
Fix: Re-apply custom firewall post-rules after VPN reconnection and rule updates

atasmohammadi:
### Problem
When using custom firewall post-rules in `/iptables/post-rules.txt`, these rules are only applied during the initial firewall setup but are not re-applied when:
- VPN connection changes or reconnects
- Firewall ports are updated
- Outbound subnets are modified

This causes custom rules to be lost during normal operation, requiring manual re-application of rules (e.g., via SSH or exec into the container).

### Solution
This PR adds a helper method `applyUserPostRules()` and ensures it's called after any significant firewall rule changes in:
- `SetVPNConnection()` in vpn.go
- `SetAllowedPort()` in ports.go
- `SetOutboundSubnets()` in outboundsubnets.go

The changes are minimal and follow the existing code patterns.

### Testing
I've tested this fix with a custom Gluetun setup using WireGuard and the following scenario:
1. Created custom post-rules in `/iptables/post-rules.txt` to allow specific traffic
2. Observed that without this fix, rules were lost after VPN reconnection
3. With this fix, rules persist through VPN reconnections and firewall updates
4. Verified with `iptables -L` that rules remain in place after reconnection events

### Use Case
This is particularly important for users who:
- Need to allow specific traffic through the VPN tunnel (tun0)
- Run other containers in network_mode: "container:gluetun"
- Need persistent custom firewall rules for services like BitTorrent clients

### Changes
- Added `applyUserPostRules()` helper method in firewall.go
- Modified `SetVPNConnection()` to re-apply post-rules after VPN changes
- Modified `SetAllowedPort()` to re-apply post-rules after port changes
- Modified `SetOutboundSubnets()` to re-apply post-rules after subnet changes
- Added improved logging in `runUserPostRules()` (optional)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,91 +1,71 @@
 name: CI
+
 on:
-  release:
-    types:
-      - published
   push:
     branches:
-      - master
-    paths:
-      - .github/workflows/ci.yml
-      - cmd/**
-      - internal/**
-      - pkg/**
-      - .dockerignore
-      - .golangci.yml
-      - Dockerfile
-      - go.mod
-      - go.sum
+      - main
+      - develop
+      - release/*
+      - beta/*
+    tags:
+      - v*
   pull_request:
-    paths:
-      - .github/workflows/ci.yml
-      - cmd/**
-      - internal/**
-      - pkg/**
-      - .dockerignore
-      - .golangci.yml
-      - Dockerfile
-      - go.mod
-      - go.sum
+    branches:
+      - main
+      - develop
+      - release/*
+      - beta/*
+  release:
+    types: [published]
+  workflow_dispatch:
 
 jobs:
   verify:
     runs-on: ubuntu-latest
-    permissions:
-      actions: read
-      contents: read
-    env:
-      DOCKER_BUILDKIT: "1"
     steps:
       - uses: actions/checkout@v4
-
-      - uses: reviewdog/action-misspell@v1
+      - uses: actions/setup-go@v5
         with:
-          locale: "US"
-          level: error
-          exclude: |
-            ./internal/storage/servers.json
-            *.md
-
-      - name: Linting
-        run: docker build --target lint .
-
-      - name: Mocks check
-        run: docker build --target mocks .
-
-      - name: Build test image
-        run: docker build --target test -t test-container .
-
-      - name: Run tests in test container
-        run: |
-          touch coverage.txt
-          docker run --rm --device /dev/net/tun \
-          -v "$(pwd)/coverage.txt:/tmp/gobuild/coverage.txt" \
-          test-container
-
-      - name: Build final image
-        run: docker build -t final-image .
+          go-version-file: go.mod
+          cache: true
+      - name: Verify
+        run: make verify
 
   codeql:
+    name: CodeQL
     runs-on: ubuntu-latest
     permissions:
       actions: read
       contents: read
       security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: ["go"]
+
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-go@v5
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      # Initializes the CodeQL tools for scanning.
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v3
         with:
-          go-version: "^1.23"
-      - uses: github/codeql-action/init@v3
-        with:
-          languages: go
-      - uses: github/codeql-action/autobuild@v3
-      - uses: github/codeql-action/analyze@v3
+          languages: ${{ matrix.language }}
+          # If you wish to specify custom queries, you can do so here or in a config file.
+          # By default, queries listed here will override any specified in a config file.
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -147,4 +130,4 @@ jobs:
             COMMIT=${{ steps.shortcommit.outputs.value }}
             VERSION=${{ fromJSON(steps.meta.outputs.json).labels['org.opencontainers.image.version'] }}
           tags: ${{ steps.meta.outputs.tags }}
-          push: true
+          push: ${{ github.event_name != 'pull_request' }}
```
qdm12, line 133: please revert this file for the time being, this is out of scope.

.github/workflows/release.yml
```diff
@@ -0,0 +1,37 @@
...
+      - name: Create Release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ github.ref }}
+          release_name: Release ${{ github.ref }}
+          body: ${{ steps.changelog.outputs.changelog }}
+          draft: false
+          prerelease: false
```
qdm12, line 37: same, revert this. Happy to get another PR eventually though.

internal/firewall/iptables.go
```diff
@@ -283,6 +283,7 @@ func (c *Config) runUserPostRules(ctx context.Context, filepath string, remove b
 			_ = c.runIptablesInstruction(ctx, flipRule(rule))
 		}
 	}()
+
```
qdm12, line 286: nit unneeded newline change here

```suggestion
```

internal/firewall/enable.go
```diff
@@ -144,13 +119,49 @@ func (c *Config) enable(ctx context.Context) (err error) {
...
+// To use in defered call when enabling the firewall.
+func (c *Config) fallbackToDisabled(ctx context.Context) {
+	if ctx.Err() != nil {
+		return
+	}
+	if err := c.disable(ctx); err != nil {
+		c.logger.Error("failed reversing firewall changes: " + err.Error())
+	}
+}
+
```
qdm12, line 164: move these blocks where they previously were to see diffs more clearly.

qdm12 (changes requested):
Thanks for the PR! Can you please move things around as suggested in comments, in order to have an easier PR review? Thanks!

internal/firewall/iptables.go
```diff
@@ -329,6 +314,21 @@ func (c *Config) RunUserPostRules(ctx context.Context, filepath string, remove b
 			rule = flipRule(rule)
 		}
 
+		// Check if this rule was already applied (avoid duplicates)
+		if !remove {
```
qdm12, line 318: Why do we track duplicates? That should be up to the user to ensure they don't have duplicate iptables commands

internal/firewall/ports.go
```diff
@@ -35,13 +35,18 @@ func (c *Config) SetAllowedPort(ctx context.Context, port uint16, intf string) (
...
 			port, intf, err)
 	}
 	netInterfaces[intf] = struct{}{}
 	c.allowedInputPorts[port] = netInterfaces
 
+	// Apply post-rules only once after adding the port, and only if not already applied
+	if err := c.applyPostRulesOnce(ctx); err != nil {
+		c.logger.Warn("failed to apply post-rules after adding port: " + err.Error())
+	}
+
```
qdm12, line 49: How is this different from just running the post rules once at firewall enabling? Especially given `applyPostRulesOnce` is not even aware of the newly allowed port.

internal/firewall/enable.go
```diff
@@ -41,62 +41,39 @@ func (c *Config) SetEnabled(ctx context.Context, enabled bool) (err error) {
...
+		if touched && err != nil {
 			c.fallbackToDisabled(ctx)
 		}
 	}()
 
-	const remove = false
+	// Clear any previously applied post-rules
+	if err = c.clearAppliedPostRules(ctx); err != nil {
+		c.logger.Warn("failed to clear previous post-rules: " + err.Error())
+	}
```
qdm12, line 67: this is not needed, given the firewall is only enabled once on a clear set of rules. Plus disable does remove all rules anyway, in case the firewall would get disabled then re-enabled.

internal/firewall/enable.go
```diff
@@ -144,13 +119,49 @@ func (c *Config) enable(ctx context.Context) (err error) {
...
 	}
 
 	return nil
 }
 
+func (c *Config) disable(ctx context.Context) (err error) {
+	// Clear applied post-rules when disabling
+	if err = c.clearAppliedPostRules(ctx); err != nil {
+		c.logger.Warn("failed to clear post-rules during disable: " + err.Error())
+	}
```
qdm12, line 134: not needed because of existing `clearAllRules`

qdm12 (changes requested):
I don't see the point of this pull request, please elaborate on how this helps

Also

> Observed that without this fix, rules were lost after VPN reconnection

That is not true, rules are persisted already as far as I'm aware.

Plus modifications to `SetVPNConnection` and `SetOutboundSubnets` are missing.

Issue #2837, opened 2025-07-03
Bug: Indian PIA IP's not updated


meghadeep-com:
### Is this urgent?

None

### Host OS

Fedora 42 (Server Edition)

### CPU arch

x86_64

### VPN service provider

Private Internet Access

### What are you using to run the container

Podman

### What is the version of Gluetun

Running version latest built on 2025-01-22T08:30:14.628Z (commit 13532c8)

### What's the problem 🤔

Hardcoded IP's for PIA's Indian region haven't been updated. 

Quick short term fix would be to update the IP's.

Dirty and medium term fix would be allowing the container to run if servers.json is read-only, or to allow IP addresses in the SERVER_HOSTNAME environment variable, since the users would then be able to update the IP's manually.

Long term fix would be using DNS queries to build the servers.json file on the fly for a chosen provider in the environment variable. 

### Share your logs (at least 10 lines)

```plain text
2025-07-03T10:37:48Z INFO [healthcheck] program has been unhealthy for 11s: restarting VPN (healthcheck error: dialing: dial tcp4: lookup cloudflare.com on 1.1.1.1:53: write udp 192.168.6.25:44916->1.1.1.1:53: write: operation not permitted)
2025-07-03T10:37:48Z INFO [healthcheck] 👉 See https://github.com/qdm12/gluetun-wiki/blob/main/faq/healthcheck.md
2025-07-03T10:37:48Z INFO [healthcheck] DO NOT OPEN AN ISSUE UNLESS YOU READ AND TRIED EACH POSSIBLE SOLUTION
2025-07-03T10:37:48Z INFO [vpn] stopping
2025-07-03T10:37:48Z INFO [vpn] starting
2025-07-03T10:37:48Z INFO [firewall] allowing VPN connection...
2025-07-03T10:37:48Z INFO [openvpn] OpenVPN 2.6.11 x86_64-alpine-linux-musl [SSL (OpenSSL)] [LZO] [LZ4] [EPOLL] [MH/PKTINFO] [AEAD]
2025-07-03T10:37:48Z INFO [openvpn] library versions: OpenSSL 3.3.2 3 Sep 2024, LZO 2.10
2025-07-03T10:37:48Z INFO [openvpn] TCP/UDP: Preserving recently used remote address: [AF_INET]23.26.221.31:1197
2025-07-03T10:37:48Z INFO [openvpn] UDPv4 link local: (not bound)
[...]

PR #2838, opened 2025-07-03
Update servers.json for PIA India region

meghadeep-com:
The IP's changed between when I opened the [issue](https://github.com/qdm12/gluetun/issues/2837) to creating this pull request. It does seem that hardcoding IP's will be unsustainable in the near future.


hitboxr:
> It does seem that hardcoding IP's will be unsustainable in the near future.

The servers are not hardcoded, gluetun includes a built-in periodic update mechanism that is disabled by default: https://github.com/qdm12/gluetun-wiki/blob/main/setup/servers.md#update-the-vpn-servers-list

Setting the environment variable `UPDATER_PERIOD=24h` will direct gluetun to pull an updated server list directly from providers every 24h.

You can also pull an updated server list manually, e.g. `docker run --rm -v .yourpath>:/gluetun qmcgaw/gluetun update -enduser -providers 'private internet access'` but note that this is not pulled through the VPN and will leak traffic (unless you tunnel it through a separate gluetun instance).

meghadeep-com:
Gotchu! This works, thanks! 

This should be in the example for PIA, or should be the default behavior, TBH.

Issue #2270, opened 2024-05-07
Bug: health check flips between unhealthy and healthy with no logs


giorgiooriani:
### Is this urgent?

No

### Host OS

Synology DSM 7.2

### CPU arch

x86_64

### VPN service provider

Windscribe

### What are you using to run the container

docker-compose

### What is the version of Gluetun

Running version latest built on 2024-05-04T16:22:29.394Z (commit ef6874f)

### What's the problem 🤔

Mentioned this in issue #2154 but thought it would be better to open a new issue. 

The container keeps flipping between healthy and unhealthy when I am downloading a torrent at a high bandwidth (80-100Mbit/s). The logs don't seem to update at all beyond the initial healthcheck which shows healthy, even in debug mode so I can't figure out why. 

It doesn't seem to be an interruption of the VPN since in the background the torrent downloads fine, except for when the VPN is restarted due to the healthcheck. My internet connection is fine and I don't have a firewall setup, so I am not sure what is causing this behaviour.

### Share your logs (at least 10 lines)

```plain text
========================================
========================================
=============== gluetun ================
========================================
=========== Made with ❤️ by ============
======= https://github.com/qdm12 =======
========================================
========================================
Running version latest built on 2024-05-04T16:22:29.394Z (commit ef6874f)
🔧 Need help? https://github.com/qdm12/gluetun/discussions/new
🐛 Bug? https://github.com/qdm12/gluetun/issues/new
✨ New feature? https://github.com/qdm12/gluetun/issues/new
☕ Discussion? https://github.com/qdm12/gluetun/discussions/new
💻 Email? quentin.mcgaw@gmail.com
💰 Help me? https://www.paypal.me/qmcgaw https://github.com/sponsors/qdm12
2024-05-07T13:00:09+02:00 INFO [routing] default route found: interface eth0, gateway 10.0.5.1, assigned IP 10.0.5.15 and family v4
2024-05-07T13:00:09+02:00 INFO [routing] local ethernet link found: eth0
2024-05-07T13:00:09+02:00 INFO [routing] local ipnet found: 10.0.5.0/24
2024-05-07T13:00:10+02:00 INFO [firewall] enabling...
2024-05-07T13:00:10+02:00 DEBUG [firewall] iptables-legacy --policy INPUT DROP
2024-05-07T13:00:10+02:00 DEBUG [firewall] iptables-legacy --policy OUTPUT DROP
[...]

PR #2877, opened 2025-08-16
Increase healthcheck dial timeout for improved stability

ls0t:
Increases the healthcheck timeouts to improve stability. With this, I haven't seen any disconnects from ProtonVPN using default settings after days, instead of constant ~15-30m reconnects.

* The healthcheck timeouts are now 30s for the first connection, and up to 60s for a second attempt
* Adjusted the initial (first failure) timeout to 60s, to match with the longer healthcheck timeout
* Changed the `unhealthy` log to INFO, because otherwise the log message just shows `healthy!` over and over again without context that the healthcheck was found unhealthy, and why

Docker image available at `ghcr.io/ls0t/gluetun:latest`

Should fix https://github.com/qdm12/gluetun/issues/2154, https://github.com/qdm12/gluetun/issues/2270

### DIFF
--- a/internal/configuration/settings/healthywait.go
+++ b/internal/configuration/settings/healthywait.go
@@ -40,7 +40,7 @@ func (h *HealthyWait) overrideWith(other HealthyWait) {
 }
 
 func (h *HealthyWait) setDefaults() {
-	const initialDurationDefault = 6 * time.Second
+	const initialDurationDefault = 60 * time.Second
 	const additionDurationDefault = 5 * time.Second
 	h.Initial = gosettings.DefaultPointer(h.Initial, initialDurationDefault)
 	h.Addition = gosettings.DefaultPointer(h.Addition, additionDurationDefault)
--- a/internal/healthcheck/health.go
+++ b/internal/healthcheck/health.go
@@ -15,13 +15,10 @@ func (s *Server) runHealthcheckLoop(ctx context.Context, done chan<- struct{}) {
 
 	timeoutIndex := 0
 	healthcheckTimeouts := []time.Duration{
-		2 * time.Second,
-		4 * time.Second,
-		6 * time.Second,
-		8 * time.Second,
+		30 * time.Second,
 		// This can be useful when the connection is under stress
 		// See https://github.com/qdm12/gluetun/issues/2270
-		10 * time.Second,
+		60 * time.Second,
 	}
 	s.vpn.healthyTimer = time.NewTimer(s.vpn.healthyWait)
 
@@ -43,7 +40,7 @@ func (s *Server) runHealthcheckLoop(ctx context.Context, done chan<- struct{}) {
 			s.vpn.healthyTimer.Stop()
 			s.vpn.healthyWait = *s.config.VPN.Initial
 		case previousErr == nil && err != nil: // First failure
-			s.logger.Debug("unhealthy: " + err.Error())
+			s.logger.Info("unhealthy: " + err.Error())
 			s.vpn.healthyTimer.Stop()
 			s.vpn.healthyTimer = time.NewTimer(s.vpn.healthyWait)
 		case previousErr != nil && err != nil: // Nth failure


qdm12:
Would changing the values mentioned in https://github.com/qdm12/gluetun-wiki/blob/main/setup/options/healthcheck.md for example `HEALTH_VPN_DURATION_INITIAL=60s` solve your case? I don't think there is a need to change the healthcheck timeout, that's why it's a constant, but change the amount of time different subsystems (i.e. vpn loop) wants to wait for the healthcheck to stay unhealthy before restarting.

Now the VPN timeouts are intentionally set as such, so that it detects connection drops fast. We don't want to wait 60s for it to restart if the connection doesn't work, especially on the first try.

qdm12:
> Changed the unhealthy log to INFO, because otherwise the log message just shows healthy! over and over again without context that the healthcheck was found unhealthy, and why

I get what you mean, although I previously changed it to DEBUG because I had way too many issues opened from users not understanding what it was, and that it was confusing. Let's keep it a the debug level. However, the "healthy!" might be misleading that can be true, we could change it something like "connection seems to be restored!" 🤔

ls0t:
Thank you for the quick response!

Increasing `HEALTH_VPN_DURATION_INITIAL` does help prevent frequent VPN reconnects, though it doesn't prevent the healthcheck failure itself. I describe the problem in more detail in [this comment](https://github.com/qdm12/gluetun/issues/2154#issuecomment-3173205950), but it seems clear that the initial 2s (1st attempt) + 4s (2nd attempt) is not enough to prevent false positive healthcheck failures. If the connection is loaded enough, the current healthcheck timeouts will cause reconnects on a perfectly healthy VPN.

I have verified this reconnect despite a healthy VPN by running pings and commenting out `onUnhealthyVpn()`. I suspect what is happening is that many VPN providers have a hard cap on concurrent TCP connections to protect their infra. So ICMP and existing connections work fine, but new connections need to wait in line for an old one to expire. Hence the false positive healthcheck failure.

One nice thing about this change is the `unhealthy` log doesn't occur very often. In fact I don't think I've seen it once since increasing to 30s. Changed back per your request though.

I truly believe this fix will solve many support issues for you. Connection stability is night and day difference after this change in my environment.

qdm12:
Very interesting read of your comment, thanks for that! A few explanations on the current mechanism (also yeah the wiki needs to be updated!):

1. The TLS handshake was added later in order to check large packets (TLS) can go through, since we did (maybe still do) have some issues when it comes to packet segmentation and packet sizes (mtu etc.). So the healthcheck would say "YAY HEALTHY!" although no TLS could be established, so not really helpful. I would like to keep it there. We also don't use ICMP since we want to check the "fullest fatest" TLS can go through, and also that DNS is working 😉 
2. For the first VPN server connection: for me, that 6 seconds timeout always work whatever the VPN server. I want to have fast feedback on that one, for example to know if the Wireguard key is wrong on a first run. We could increase it to say 2s+3s+4s (healthcheck timeouts) => 9+1s (vpn timeout)
3. For maintaining the VPN connection, if a TLS handshake cannot go through within 10seconds, IMO the connection doesn't work. Who is going to wait 10s for an https request 😄 At least from a usability standpoint, that would be my definition of a non working connection. Now if you do funny things and overload the connection, then you can bump the available VPN timeouts.
4. I'm not sure I understand this:

    > I suspect what is happening is that many VPN providers have a hard cap on concurrent TCP connections to protect their infra.

     The healthcheck doesn't run concurrent TCP connections right?
5. Finally, since the VPN restart/timeouts is somehow decoupled from the healthcheck failing, I think it's fine if the healthcheck goes in failing state from time to time, before returning to healthy state.

Just to be clear, I'm more than happy to improve this, but let's consider all the aspects of it, especially since this healthcheck logic has been re-worked many many many times over the years 😸 Right now the healthcheck focuses on end-to-end usability (10s timeout for TLS, including DNS).

ls0t:
> The TLS handshake was added later in order to check large packets (TLS) can go through

Gotcha, I can definitely see where you're coming from regarding working TLS connections. I agree it makes sense to check these periodically.

> The healthcheck doesn't run concurrent TCP connections right?

The healthcheck doesn't run concurrent TCP connections, but my assumption is that most people use gluetun for torrent clients, and these create a ton of TCP connections, with qbittorrent defaulting to 500 connections, and some guides advising increasing the connections further. So while the vpn is available, there may be an inability to establish new TCP connections because the provider's NAT tables are full. I get what you mean about 10s+ TLS connections when browsing, but torrent clients create network conditions that directly compete with low-latency establishment of new TCP connections. Not to mention some connections have higher latency by nature, say a mobile or starlink connection, that will hit the 10s threshold that much faster.

I ran another experiment, where I went back to the latest gluetun release with default settings and lowered the connections on my qbit from 500 to 100. It hasn't reconnected yet after 30 minutes, will run it overnight to see (edit: it remained stable overnight). According to netstat, I was peaking around 400 connections when qbit was configured for 500. With qbit configured for 100, I sit around 250, though once in a while it goes back up to 400 anyway. My impression is qbit's max connection setting is more like a guideline, and the more you seed, the more you're going to get connections no matter what you do. It also feels like a lot to ask everyone to tweak this setting on all their software, rather than gluetun being robust in a variety of situations.

> for me, that 6 seconds timeout always work whatever the VPN server.

Yes it works initially, most of the time, but it doesn't continue to work over time when the torrent client is running. Besides the VPN provider, the reliability may also depend on the speed of your internet connection.

> We could increase it to say 2s+3s+4s (healthcheck timeouts) => 9+1s (vpn timeout)

IMO there are really two types of concerns: 1/ is the connection up, and 2/ is the connection configured properly. Since wireguard won't tell us directly, (brainstorming here) taking your idea a bit further, in the long term we could break out the healthcheck into two orthogonal checks:

1. ICMP for connectivity. Pinging the PtP tunnel endpoint would tell us if the other side of the connection is responding. Because we can run pings continuously, like once per second, we'll get very quick feedback if the connection goes down. ICMP won't count against our TCP state table limits. Plus we'll be able to measure if there is any appreciable packet loss, helping identify unstable connectivity.
2. TLS checks for configuration. If we can't establish TLS or DNS after startup, its likely there is a MTU or other configuration problem. This should be done at startup and repeat periodically, but it doesn't need to be done that frequently.

> I think it's fine if the healthcheck goes in failing state from time to time

Agreed that healthchecks can and will fail. IMO the tolerances right now are just too low for connection loaded by a torrent client (TCP connections and bandwidth being the presumed bottlenecks).

> Right now the healthcheck focuses on end-to-end usability (10s timeout for TLS, including DNS)

I completely get this, but I will say in practice that while I see some DNS failures, they're not that common. Maybe one every couple of hours. I don't know how the current client works, but it could be made more robust by keeping existing DoT connections open. Some quick testing suggests cloudflare will allow idle for about 1m45s and quad9 for about 55s.

One other thing - another benefit I noticed to the increased timeout is that my port forwards aren't changing all the time. When I use the official ProtonVPN client on my desktop, my seaport tool will keep the same port reservation for weeks at a time. But with the official gluetun release, even if the VPN doesn't reconnect, my port forwards change several times per day. With this PR, the port forwards stay stable like when I'm using the official client. This is a big benefit that prevents torrent clients from reconnecting and makes a difference in the user's experience since you're not constantly losing connections to seeders that don't port forward. I don't quite know why the port forwards no longer reset, but its another favorable data point.

ls0t:
Just a friendly ping on this issue. My torrents are still running better with the change, but overall the torrent client is not performing well in gluetun. Specifically there are some torrents with a 256mb block size that I just can't download when using gluetun despite plenty of peers. I believe gluetun is somehow causing frequent termination of TCP connections, and its very unlikely for my client to be able to download 256mb in one connection before it terminates. As a result, I have a 250gb torrent for which my client has attempted to download 750gb. Despite continuing to transfer data, the progress meter bounces between 89 and 90%, moving both directions.

When I switch my torrent client to use the official protonvpn client instead of gluetun, my download makes progress right away.

qdm12:
We can definitely have a ICMP smallest packet exchange healthcheck run every ~5s, and a more involved (TLS+DNS) check run at start (after ICMP ideally) and then every 5 minutes. The only worry I have would be if ICMP gets blocked by some VPN servers, but it's worth a try. What should we ICMP? The VPN server I suppose? I'll get started on that soon! (EDIT: started working on it now)

qdm12:
What I'm thinking of now, is to run health checks more from within the VPN loop, instead of independently of it:

1. VPN tunnel is connected (success openvpn or silent wireguard)
2. TCP+TLS+DNS check with 4s timeout. On failure, abort and restart the VPN
3. Start a service (aka long running goroutine) which performs:
    - every 5s, ICMP check (3s timeout). On failure, restart the VPN
    - every 5 minutes, TCP+TLS+DNS check (15s timeout). On failure, restart the VPN
4. Setup DNS over TLS, get public ip info etc. - the DNS should revert to plaintext in case it fails to setup
5. When VPN stops (crash or other), stop the service `4.` which would then be restarted by VPN up

qdm12:
This is implemented in https://github.com/qdm12/gluetun/pull/2923 which you can try with image tag `:pr-2923`. Closing this in favor of that new PR, although the description refers to the discussions over here.

ls0t:
Wow, thanks for all the progress on this. I stepped away for a few days and came back to a flurry of activity. 

In terms of where to ping, I usually ping the other end of the PtP tunnel. So if the wireguard interface has IP 10.2.0.2 the typical VPN-provider-side of the tunnel would be 10.2.0.1. This is the most direct route to the vpn provider, but doesn't say if you can reach the internet. 1.1.1.1 works too. I'll take a look at the PR!

qdm12:
So interestingly enough, trying to icmp ping the VPN local gateway (10.2.0.1 I believe) doesn't work. I have to ping the server public IP address for it to work 🤔 (or, again, 1.1.1.1, which does work well). Right now there is an option HEALTH_ICMP_TARGET_IP which defaults to the VPN public server IP address you are connected to.

However, I have one user facing a "sendto: operation not supported" for sending ICMP packets, whatever we tried to do permission-wise. And I feel he won't be the only one facing this if we merge this to master. Therefore, I'm thinking of [doing an UDP ping](https://github.com/qdm12/gluetun/pull/2923#issuecomment-3383064598). Now I could have this as a fallback if sending ICMP packets isn't allowed, but, ideally I would like to have only one way to simplify debugging and code, so just send UDP packet and listen for ICMP. I then thought about what you said on opening and closing TCP connections. Since UDP is connection-less, and that we target a UDP port not listening on anything in that UDP ping case, I think it would be fine to send UDP packets every ~10s right?

ls0t:
Hmm, it might depend on the VPN provider. I have no problem pinging 10.2.0.1 on Proton:

```
# ping 10.2.0.1 
PING 10.2.0.1 (10.2.0.1) 56(84) bytes of data.
64 bytes from 10.2.0.1: icmp_seq=1 ttl=64 time=48.2 ms
```

From a TCP connections limit, yes the UDP ping should not be impacted by a full TCP state. The challenge with UDP ping is that not all networks will return destination unreachable. It is standard IP stack behavior, but networks can filter them out for security reasons, because they make enumerating UDP services faster. I don't know the prevalence of this behavior, however if someone is impacted there is likely nothing they can do if they aren't the administrator of the network. Whereas at least with the ICMP permission issue, the user could change their stack somehow to overcome the problem.

Thinking outside the box, a plaintext DNS lookup might be a good "UDP ping". The nice part about this is its just a standard application-level connection, no ICMP or special packet handling required. Of course we'd just lookup a common website like google instead of anywhere the user is going.

PR #2883, opened 2025-08-28
Use go v1.25

shwoop:


qdm12:
Thanks for the contribution!
Unfortunately Gluetun is a pain to upgrade its dependencies to support **everyone**. I'm still working on having fixes on top of v3.40.0 and then we'll see about upgrading Alpine and Go.
For context:
- upgrading Alpine causes Openvpn/openssl/various-other-things to be upgraded, and this can break many setups (usually not mine 🤣)
- upgrading Go can **ALSO** break things. Actually just recently, bumping from 1.23 to 1.22 leaded to headaches when it comes to TLS since they now use quantum resistant algorithms, which have larger signatures, which don't work with the MTU we had configured.... as usual, on many setups - not mine.

So these are hard to test out for everyone, and because I'm only catching up (I was away for a few months), I'll close this for now. At the very least, these two bumps should be split in two, and separated by a few days to allow users to report issues.

PR #2884, opened 2025-08-28
chore(dev): add gpg for sign commits

shwoop:
To perform commit signing, the devcontainer only requires that gpg be installed.

This commit was signed from within the devcontainer.

https://code.visualstudio.com/remote/advancedcontainers/sharing-git-credentials?originUrl=%2Fremote%2Fadvancedcontainers%2Fsharing-git-credentials#_sharing-gpg-keys


.devcontainer/Dockerfile
```diff
@@ -1,2 +1,2 @@
 FROM ghcr.io/qdm12/godevcontainer:v0.21-alpine
-RUN apk add wireguard-tools htop openssl
+RUN apk add wireguard-tools htop openssl gnupg
```
qdm12, line 2: Nice, actually what would be better would be to have installed in https://github.com/qdm12/basedevcontainer/ (on both Alpine.Dockerfile and debian.Dockerfile), since this is used as a base for godevcontainer which is used here.
Would you mind opening a PR over there instead? 🙏 If not, let me know and I'll add it on that repository, no problem as well 😉

qdm12:
Side note, I'm actually signing commits with ssh and that doesn't need gpg 😉

.devcontainer/Dockerfile
shwoop: No problem.  I saw the base image repo but thought it was a bit presumptious of me to raise the PR there.

shwoop:
Moved upstream https://github.com/qdm12/basedevcontainer/pull/34

PR #2901, opened 2025-09-20
Update dependencies

programmerPhysicist:
Update a number of dependencies, using some branches from dependabot.


qdm12:
Thanks for the PR 💪 unfortunately I noticed it just now, and all are upgraded at this point.
Another note, it's better to upgrade dependency one per commit, so that it can later be dissected in case of issues.
I had some niche issues for just a few users, when upgrading Go dependencies and even Go version, due to Gluetun being rather low level in some aspects, so there is always a risk of breaking things when upgrading anything!
Closing this as unneeded, but I do appreciate the effort! 🎖️

PR #2888, opened 2025-09-02
feat(expressvpn): update hardcoded servers data

Lobstrosity:
I pulled down all the OVPN files from ExpressVPN’s manual configuration page and mapped the country, city (optional), and hostname from each.

PR #2881, opened 2025-08-27
Update servers list

shwoop:
Resolves #2876
Resolves #2831

PR #2918, opened 2025-09-30
Introduce pull request template

shwoop:
This template intends to highlight a number of facts about the codebase:
* do not edit servers.json.
* update the wiki if needed.

While also promoting best practices:
* there are tests for changes.
* the changes have been "deployed" and connected to one or more vpn providers.

The irony that this PR doesn't link back to an issue is not lost on me.

The main aim of the template is to promote the cli server config updating functionality and avoid PRs such as:
https://github.com/qdm12/gluetun/pull/2881
https://github.com/qdm12/gluetun/pull/2838
~~https://github.com/qdm12/gluetun/pull/2888~~
https://github.com/qdm12/gluetun/pull/2808


Lobstrosity:
Side question… what’s the reasoning for including the generated `servers.json` in the codebase at all? (As opposed to generating/updating it at runtime for the specified provider if it’s missing/stale.) cc @qdm12

Also (regarding #2888), I’m not sure that local update/merge even works for ExpressVPN. As far as I can tell, it’s the only provider that uses [a hard-coded list of servers](https://github.com/qdm12/gluetun/blob/master/internal/provider/expressvpn/updater/hardcoded.go). I wanted to use a server that isn’t in that list and providing a local file with just that server wouldn’t sync with the existing list.

.github/pull_request_template.md
```diff
@@ -0,0 +1,38 @@
+# Description
+
+<!-- Please describe the reason for the changes being proposed. -->
+
+# Issue
+
+<!-- Please link to the issue this change relates to. -->
```
qdm12:
```suggestion
<!-- Please link to the issue(s) this change relates to. -->
```

shwoop:
> Also (regarding https://github.com/qdm12/gluetun/pull/2888), I’m not sure that local update/merge even works for ExpressVPN.

That's a good point, I hadn't looked at the expressvpn provider, in the case of https://github.com/qdm12/gluetun/pull/2888, the servers.json has not been manually edited so would still be valid.

Maybe some language along the lines of "I did not manually edit the servers.json file" would fit better.

>what’s the reasoning for including the generated servers.json in the codebase at all? (As opposed to generating/updating it at runtime for the specified provider if it’s missing/stale.)

After opening this PR I created a [ticket](https://github.com/qdm12/gluetun/issues/2919) possible waysto avoid issues/PRs around stale hardcoded servers.json.

.github/pull_request_template.md
```diff
@@ -0,0 +1,38 @@
...
+
+# Assertions
+
+* [ ] I am aware that we do not accept manual changes to the servers.json file <!-- If this is your goal, please consult https://github.com/qdm12/gluetun-wiki/blob/main/setup/servers.md#update-using-the-command-line -->
+* [ ] I am aware that any changes to settings should be reflected in the [wiki](https://github.com/qdm12/gluetun-wiki/)
+* [ ] Test are included or these changes are covered by existing tests
```
qdm12: let's remove this for now, it's not always needed, and will ask for it if it feels needed. I prefer production code contributions versus higher-bar-tested-contributions (eventhough yeah it would be great). I feel like I first need to improve testing coverage of gluetun before asking this.

```suggestion
```

.github/pull_request_template.md
```diff
@@ -0,0 +1,38 @@
...
+    * [ ] ProtonVPN
+    * [ ] PureVPN
+    * [ ] SlickVPN
+    * [ ] Surfshark
+    * [ ] TorGuard
+    * [ ] VPNSecure.me
+    * [ ] VPNUnlimited
+    * [ ] VyprVPN
+    * [ ] WeVPN
+    * [ ] Windscribe
```
qdm12: Let's also remove this, because it's annoying to maintain yet another list of providers, whenever adding a provider

Anyway, this should be the CI responsibility through integration (not really any, something I need to work on!) or unit tests I think.

```suggestion
```

qdm12 (approved):
Thanks, much appreciated and much needed indeed!

qdm12:
> Side question… what’s the reasoning for including the generated `servers.json` in the codebase at all? (As opposed to generating/updating it at runtime for the specified provider if it’s missing/stale.)

Because Gluetun doesn't (and should not) have internet access before starting, so it needs servers data before starting. Hence the servers.json in the repository, since it's built-in the Go binary program. Basically as long as there at least a few servers still working for your provider, you can then update the servers automatically periodically (check the wiki, there is an env variable for that, it's off by default - maybe worth turning on by default I suppose...)

> Also (regarding #2888), I’m not sure that local update/merge even works for ExpressVPN. As far as I can tell, it’s the only provider that uses [a hard-coded list of servers](https://github.com/qdm12/gluetun/blob/master/internal/provider/expressvpn/updater/hardcoded.go). I wanted to use a server that isn’t in that list and providing a local file with just that server wouldn’t sync with the existing list.

Yeah it wouldn't really work, because updating servers locally only updates IP addresses (still useful eventually) of existing hardcoded servers. Sadly Expressvpn doesn't provide an API endpoint or file with all their servers listed, so this cannot be automated. One thing you could do is have your own servers.json bind mounted, modify it, and bump the timestamp for expressvpn in the json file, that will use your version. Anyone reading this and NOT using expressvpn, don't do this, instead update your servers programmatically!

PR #2940, opened 2025-10-12
fix(wg): can config endpoint with a host

mathieuruellan:
I want to set my wireguard endpoint with a hostname and not an IP address.
The IP Address should be re-evaluate not only on startup in case of the ip has changed
The test part is still ugly. 
Suggestions are welcome


qdm12:
No can do, network should not be permitted until the VPN is up. I'm working on a fix but it's complicated and will take some time to get finished. For now, you'll have to resolve the hostname yourself on the side.

PR #2970, opened 2025-11-05
feat(dns): resolve network-local names

qdm12:
WHY: To finally be able to resolve container names present in the same container network as Gluetun.
**WHAT I NEED FROM YOU** 🆘 : What is your current usage of `DNS_KEEP_NAMESERVER` or `SERVER_ADDRESS`?

NEXT STEPS:
1. Always enable the built-in DNS forwarder server (older `DOT` gone`) so that local names can always resolve
2. Remove `DNS_KEEP_NAMESERVER`. It should now be unneeded as far as I know.
3. Repurpose the `DNS_ADDRESS` such that:
    - if set to `127.0.0.1` (current default): ignore it
    - if it is without a port (current format): have the DNS forwarder server use it as the upstream resolver and force the upstream type to plain (=udp port 53)
    - otherwise (with a port specified, new format): have the DNS forwarder server use it as the upstream resolver and use the  upstream type as defined by `DNS_UPSTREAM_RESOLVER_TYPE`

---

On this PR:

You can run it with tag `:pr-2970`. No other configuration needed.

This is handled by a middleware built-in the local DNS forwarding server. Local names are queried to the private nameservers found in /etc/resolv.conf at container start (i.e. 127.0.0.11 for a Docker bridge network), and forwarded back to clients. That way you can have both DNS over TLS (with caching, filtering etc.) + local container names. Note non-local names AND local names not resolved are both then handled by the normal DNS over TLS pipeline (including filtering, caching etc.)

Small additional notes for this PR:

- having DOT=off or DNS_SERVER=off will cause the local resolution to no longer work, since it's built-in the local forwarder server

### DIFF
--- a/internal/dns/loop.go
+++ b/internal/dns/loop.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"net/netip"
 	"time"
 
 	"github.com/qdm12/dns/v2/pkg/middlewares/filter/mapfilter"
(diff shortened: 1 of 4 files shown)


PythonCoderAS:
For `DNS_KEEP_NAMESERVER`, I was attempting to fix an issue I was having with DNS requests timing out (I eventually traced the culprit to excessive outbound connections from the container behind gluetun).

qdm12:
This feels much needed, merging this for v3.41.0 let's yolo it

hitem:
WHAT I NEED FROM YOU 🆘 : What is your current usage of DNS_KEEP_NAMESERVER or SERVER_ADDRESS?
--------------------------
My use of 

      DOT: 'off'
      DNS_KEEP_NAMESERVER: 'on'
      DNS_ADDRESS: '192.168.1.155'  (obv not my custom dns, but this is my config basicly).
      
Is so i can use my own in house DNS server that uses DoH with cloudflared.
If i remove the DNS_KEEP_NAMESERVER i stop seeing requests coming through - i have multiple security filtering features and adblockers that i want to kick in before i ship it with DoH via cloudflared tunnel. I want to be able to also block specific domains or even whitelist. I do not have the ability to do so if i use gluten default settings or even set DNS_ADDRESS without the DNS_KEEP_NAMESERVER.  

Let me know if you need anything else, superappriciated of the software, always have been.
Keep up the great work.

Edit: My local DNS is its own device running on my network, so its no container that lives in same space as gluetun (gluetun runs on my machine specifically).

qdm12:
@hitem Thanks! Got it! Please try out image tag `pr-2988` (https://github.com/qdm12/gluetun/pull/2988) without changing your settings to see if it continues working? It should maintain retro-compatibility whilst removing a bunch of options. Coming post v3.41.0 release. Please comment on that #2988 since I won't monitor this one anymore! Thanks!

PR #2991, opened 2025-11-18
fix(purevpn/updater): parse country and city from hostname and merges with ip address information

mlapaj:
# Description

- The original updater fetches servers from the zip file and checks geo-location of each server IP address. This is not working correctly for virtual servers.
- Some VPN servers have incorrect geolocation data

# Assertions

### DIFF
--- a/internal/healthcheck/checker.go
+++ b/internal/healthcheck/checker.go
@@ -25,7 +25,7 @@ type Checker struct {
 	smallCheckType string
 	configMutex    sync.Mutex
 
-	icmpNotPermitted bool
+	icmpNotPermitted *bool
 
 	// Internal periodic service signals
 	stop context.CancelFunc
@@ -70,7 +70,7 @@ func (c *Checker) Start(ctx context.Context) (runError <-chan error, err error) 
 		panic("call Checker.SetConfig with non empty values before Checker.Start")
 	}
 
-	if c.icmpNotPermitted {
+	if c.icmpNotPermitted != nil && *c.icmpNotPermitted {
 		// restore forced check type to dns if icmp was found to be not permitted
 		c.smallCheckType = smallCheckDNS
 	}
@@ -154,11 +154,12 @@ func (c *Checker) smallPeriodicCheck(ctx context.Context) error {
 		}
 		ip := icmpTargetIPs[try%len(icmpTargetIPs)]
 		err := c.echoer.Echo(ctx, ip)
-		if errors.Is(err, icmp.ErrNotPermitted) {
-			c.icmpNotPermitted = true
+		if c.icmpNotPermitted == nil && errors.Is(err, icmp.ErrNotPermitted) {
+			c.icmpNotPermitted = new(bool)
+			*c.icmpNotPermitted = true
 			c.smallCheckType = smallCheckDNS
 			c.logger.Infof("%s; permanently falling back to %s checks",
-				smallCheckTypeToString(c.smallCheckType), err)
+				err, smallCheckTypeToString(c.smallCheckType))
 			return c.dnsClient.Check(ctx)
 		}
 		return err
--- a/internal/vpn/tunnelup.go
+++ b/internal/vpn/tunnelup.go
@@ -51,6 +51,11 @@ func (l *Loop) onTunnelUp(ctx, loopCtx context.Context, data tunnelUpData) {
 		l.logger.Info("👉 See https://github.com/qdm12/gluetun-wiki/blob/main/faq/healthcheck.md")
 	}
 
+	// Start collecting health errors asynchronously, since
+	// we should not wait for the code below to complete
+	// to start monitoring health and auto-healing.
+	go l.collectHealthErrors(ctx, loopCtx, healthErrCh)
+
 	if *l.dnsLooper.GetSettings().ServerEnabled {
 		_, _ = l.dnsLooper.ApplyStatus(ctx, constants.Running)
 	} else {
@@ -79,8 +84,6 @@ func (l *Loop) onTunnelUp(ctx, loopCtx context.Context, data tunnelUpData) {
 	if err != nil {
 		l.logger.Error(err.Error())
 	}
-
-	l.collectHealthErrors(ctx, loopCtx, healthErrCh)
 }
 
 func (l *Loop) collectHealthErrors(ctx, loopCtx context.Context, healthErrCh <-chan error) {


qdm12:
> I have updated them. We should consider to provide a way to update certificates and keys instead of hard-coding them.

Where did you get them from? Any publicly available source for them?

internal/provider/purevpn/openvpnconf.go
```diff
@@ -20,10 +20,11 @@ func (p *Provider) OpenVPNConfig(connection models.Connection,
...
...enFKsjuiAeFQ5CJ7SR11n/2x+Z8Rkt4TIZY0RfdEOm6cYc+fCzYJOdalUKiFJPVfB4SKAGILamLK6FCbsNR6sdVGQC/b1CSGjycqlQTC+XJqvBfx5yCjEE/Y0tutuz4cK/Bx6KEaO6fWJ2ZQKCAQEAhmB9Cm+7VklWSSbrPuvWaAy12xB2iVQKP2hWquddCDUE82IZVqouCpR7TrtaiKgiEpTNAIb+TbMhqqrkgRt0Ybh+c2OWNzcuK5VV0R5ccWqzLzoUQu6O4Da90qLQyEAXwiLP5TKCJMH7LujZVoJGGaKUJwOGTrZHOypj6fkEusmSIg8cpBJzM2h8dK+SfbWewYlhGDU54sKkZAH4bENptHAF9EhdU+0CkXKN7unm4JuOtxDMlrCE6S+YP8TUzmAgWsoztC+hXdKs20yNqo1rG9fsNyZaGrRBU3uMJ/2aeGD8XwSaNNcB6ryYYQ9SxgfkewEmOK0ichB+9lppTqwh3QKCAQBaQ5MO9AezfykUcMvKq2j1pQdhV+fH34ebFExdSALP6O/bg78WcfVtZDvR3F9ZCoqXHCSgy7uJPLgkUpMDJ3iFFiVh6IlZVzZbShCJkQLIglVlkh/iZwjv9pOjoHare332jRBSwpVwJIOs9bBydWqYs+jfQLKmVXvs25gNOWWyMgrjCHRnOPWExK93ZY+SlMbu8VsukW4F4gxsOVUu34jumqHP0QEBpDuUJR9cTXF5L+IkGvpgJeXZEhe4qX95XRAZ23KOpR5WP0ji+FH30SG21JqJUdP5tR9bPkgFP65pm5i4YeUmQ8pKAo/0j+EjWd9dgquC4V15AjxD1OOcwkupAoIBAQCDrswYi3ffzsnpqQ1MWrJnjkyB/fL6OuT+m+BLpn4JspLE0JYB/ykoqcv8SKNzDU0NBjozETmlwYVZpI6PAybVvE+ixkcsbOL1yJOI0dY5DcHK3CLI6/kaXtBxPBy7t8hEXvMA+wxO3da4UeDuJW26Y6kw+r5lT9trKrUosH9YvbnwYnW0EAGiSvYtyTq7Pk6r7Hxloeyk0ALeGHmyXUbMETXXGxuGb5LFnAdDTxr9td+moG0GcPvf0VmiB4K1B7/527q2WfKycS9/1nn4qAFPvykuzmQOM368ByDyr17SiYpoN19uDfqZa2caHmVSrhJKhKIBDGzuB/nUYJOnT4fX",//nolint:lll
```
qdm12: Is this a new requirement?

internal/provider/purevpn/updater/getocation.go
```diff
@@ -0,0 +1,302 @@
+package updater
+
+import (
+        "strings"
+)
+
+
+func getLocation(host string) (country string, city string){
+    var countryMap = map[string]string{
```
qdm12: use `constants.CountryCodes()` (from internal/constants)

internal/provider/purevpn/updater/getocation.go
```diff
@@ -0,0 +1,302 @@
...
+        "IL": "Chicago",
+        "UT": "Salt Lake City",
+        "TX": "Houston",
+        "WDC": "Washington DC",
+        "NJ": "Newark",
+        "VA": "Ashburn",
+        "CA": "Los Angeles",
+        "SA": "Seattle",
+        "GA": "Atlanta",
+
```
qdm12:
```suggestion
```

internal/provider/purevpn/updater/getocation.go
```diff
@@ -0,0 +1,302 @@
...
+        "EH": "Western Sahara",
+        "YE": "Yemen",
+        "ZM": "Zambia",
+        "ZW": "Zimbabwe",
+    }
+
+
+
+    // City code -> country name map
+    var cityMap = map[string]string{
```
qdm12: nit declare this variable globally so we don't re-create the map for every call to `getLocation` for each server

internal/provider/purevpn/updater/getocation.go
```diff
@@ -0,0 +1,302 @@
...
+
+        "PE": "Perth",
+        "SD": "Sydney",
+        "ME": "Melbourne",
+    }
+
+    code := ""
+    codeCity := ""
+    code = strings.ToUpper(host[0:2])
+    codeCity = strings.ToUpper(host[2:strings.Index(host,"2-")])
```
qdm12:
```suggestion
    code := strings.ToUpper(host[0:2])
    codeCity := strings.ToUpper(host[2:strings.Index(host,"2-")])
```

internal/provider/purevpn/updater/getocation.go
```diff
@@ -0,0 +1,302 @@
...
+
+    country = countryMap[code]
+    if country == "" {
+        country = code // fallback if not in map
```
qdm12: log a warning if this is not in map, so we know we should update the map

internal/provider/purevpn/updater/getocation.go
```diff
@@ -0,0 +1,302 @@
...
+    }
+
+    city = cityMap[codeCity]
+    if city == "" {
+        city = codeCity // fallback if not in map
```
qdm12: log a warning if this is not in map, so we know we should update the map

internal/provider/purevpn/updater/servers.go
```diff
@@ -80,22 +79,13 @@ func (u *Updater) FetchServers(ctx context.Context, minServers int) (
...
-		return nil, err
-	}
+    
 	for i := range servers {
-		servers[i].Country = ipsInfo[i].Country
-		servers[i].Region = ipsInfo[i].Region
-		servers[i].City = ipsInfo[i].City
+        country, city :=  getLocation(servers[i].Hostname)
+		servers[i].Country = country
+		servers[i].Region = ""
```
qdm12: ⚠️ changes needed: that's problematic, some users might be using `SERVER_REGIONS` as filters for their servers. If all servers data has no region, no server will be found. We can't do this, let's find a workaround.

qdm12: We could use the public ip data fetcher to get information, and cross check with the country and city from `getLocation`. If it's the same, we can use the region from the ip data fetcher. If not, we leave it to empty since it's likely invalid.

mlapaj:
> > I have updated them. We should consider to provide a way to update certificates and keys instead of hard-coding them.
> 
> Where did you get them from? Any publicly available source for them?

This was taken from the zip file - Recommended-CA2.zip. See certificates.

internal/provider/purevpn/openvpnconf.go
mlapaj: This is not new requirement. It comes from Recommended-CA2.zip. This file contains opevpn files and certificates as well

internal/provider/purevpn/updater/getocation.go
mlapaj: ok

internal/provider/purevpn/updater/servers.go
mlapaj: @qdm12 : Thanks for the review. I'll work on the code. This was rather a draft code / PoC.
I think we can use mixed approach - fetch IP data and if it is different, use data from getLocation.
Also I'll try to get server_region mapping for virtual addresses.

internal/provider/purevpn/openvpnconf.go
qdm12: I'm talking about the rsa key specifically 😉

qdm12:
A few changes done:

- openvpn config changes moved out of PR to have a focused PR. I did push those value updates to the master branch attributing you credits in the commit message
- added missing country codes and city codes in the mapping
- use country-city codes together to determine city name, because there could be `ukl` and `usl` for example (ugh lazies purevpn!)
- it tries to parse the country and city from the hostname. If it succeeds, it uses the region from the ip info data ONLY if the country and city more or less match what we parsed.
- if parsing the hostname fails, it uses the country and city from the ip info data
- on the master branch, I also pushed a change where the ip data fetcher now merges "cleverly" ip data from different sources, to have a more reliable country, city etc.

qdm12:
Also, when it comes to keys/certificates etc. these should not change. It's poor practice to just deprecate these without a every early warning as well. PureVPN is just meh so they changed all their keys/certs/etc. out of nowhere.

PR #3010, opened 2025-11-24
feat: Add new workflow to update the maintainers servers list

shwoop:
# Description

A new workflow has been created that will run the update servers action in maintainer mode, then raise/merge a PR with the changes.

Trivial checks are performed:
* the static servers list file has been altered
* no other files have been altered
* ~~the static servers list file is valid json~~

The workflow will run once every two months, on the first of the month, at 03:11UTC.
The workflow can also be triggered manually.

You can see some invocations on my fork here: https://github.com/shwoop/gluetun/actions/workflows/update-servers-list.yml
An example merged PR (since reverted): https://github.com/shwoop/gluetun/pull/16

# Issue

https://github.com/qdm12/gluetun/issues/2919

# Assertions

* [x] I am aware that we do not accept manual changes to the servers.json file 
* [x] I am aware that any changes to settings should be reflected in the [wiki](https://github.com/qdm12/gluetun-wiki/)

PR #3014, opened 2025-11-25
feat(windscribe): update servers data

LogMull:
# Description

Updated windscribe servers data using the cli tool.

# Issue

NA

# Assertions


qdm12:
@shwoop I thought this PR might make you smile

qdm12:
The point of the CLI tool is that you can use it locally 😉 Not really to make PRs. But thanks for the time invested still!
Closing this in favor of #3010

LogMull:
I understand, thanks for taking the time to look at it.  
The reason I submitted the PR was because our kube pods would never connect to the severs listed in the file, and because they could not connect to the VPN, the update was never able to run. 

I've since found another way around the issue we were having.

PR #3058, opened 2025-12-23
fix(publicip/api/cloudflare): add `Referer` header

jjlin:
# Description

The Cloudflare API now returns HTTP 403 without an appropriate `Referer` header

# Issue

None

# Assertions


qdm12 (approved):
Awesome, thanks!! 👍

Issue #2876, opened 2025-08-15
Bug: privado server list out of date and scrambled


eddiert:
### Is this urgent?

Yes

### Host OS

docker - truenas

### CPU arch

x86_64

### VPN service provider

Privado

### What are you using to run the container

docker-compose

### What is the version of Gluetun

Running version latest built on 2025-01-22T08:30:14.628Z (commit 13532c8)

### What's the problem 🤔

Privado server list seems to be out of date and wrong. For instance when I manually get a server config from the privado site for LAX it suggests lax-018. When I parse the privado servers.json I find only lax-013, lax-014, lax-015 none of which are resolvable through dns. Second I see many servers under city:los_angeles that I know shouldn't be there, like mex-011, mnl-006, nrt-011 etc.

### Share your logs (at least 10 lines)

```plain text
Sorry I don't have my logs as I have already worked around the problem by setting up a custom config. Here are some logs from my custom config
2025-08-15 06:22:38.716677+00:002025-08-15T06:22:38Z INFO [routing] default route found: interface eth0, gateway 172.16.6.1, assigned IP 172.16.6.2 and family v4
2025-08-15 06:22:38.716806+00:002025-08-15T06:22:38Z INFO [routing] default route found: interface eth0, gateway fdd0:0:0:3::1, assigned IP fdd0:0:0:3::2 and family v6
2025-08-15 06:22:38.716936+00:002025-08-15T06:22:38Z INFO [routing] adding route for 0.0.0.0/0
2025-08-15 06:22:38.717072+00:002025-08-15T06:22:38Z INFO [routing] adding route for ::/0
2025-08-15 06:22:38.717126+00:002025-08-15T06:22:38Z INFO [firewall] setting allowed subnets...
2025-08-15 06:22:38.717272+00:002025-08-15T06:22:38Z INFO [routing] default route found: interface eth0, gateway 172.16.6.1, assigned IP 172.16.6.2 and family v4
2025-08-15 06:22:38.717420+00:002025-08-15T06:22:38Z INFO [routing] default route found: interface eth0, gateway fdd0:0:0:3::1, assigned IP fdd0:0:0:3::2 and family v6
2025-08-15 06:22:38.717640+00:002025-08-15T06:22:38Z INFO TUN device is not available: open /dev/net/tun: no such file or directory; creating it...
2025-08-15 06:22:38.717925+00:002025-08-15T06:22:38Z INFO [http server] http server listening on [::]:8000
[...]

Issue #2118, opened 2024-02-18
Bug: Update the VPN servers list periodically does not work for Privado


moglifreitag:
### Is this urgent?

No

### CPU arch

None

### VPN service provider

Privado

### What are you using to run the container

docker-compose

### What is the version of Gluetun

Running version latest built on 2024-02-14T07:39:38.933Z (commit 423a5c3)

### What's the problem 🤔

There are new hostnames for Privado VPN server in "Switzerland" based on https://privadovpn.com/apps/ovpn_configs.zip
zrh-0(09-12).vpn.privado.io

Current servers.json includes
zrh-00(5-8).vpn.privado.io

Activated periodically update with UPDATER_PERIOD=5m and it runs without an error. But servers.json remains unchanged.
Also tried UPDATER_MIN_RATIO=0.1 (beside the default 0.8)

Manual update works with docker run --rm -v /yourpath:/gluetun qmcgaw/gluetun update -enduser -providers privado

### Share your logs (at least 10 lines)

```plain text
2024-02-18T00:34:10+01:00 INFO [dns] ready
2024-02-18T00:34:10+01:00 INFO [vpn] You are running on the bleeding edge of latest!
2024-02-18T00:34:10+01:00 INFO [ip getter] Public IP address is 98.98.131.240 (Turkey, Istanbul, Istanbul)
2024-02-18T00:39:02+01:00 INFO [updater] updating Privado servers...
2024-02-18T00:39:04+01:00 WARN [updater] reached the maximum number of consecutive failures: 2 failed attempts resolving prg-003.vpn.privado.io: lookup prg-003.vpn.privado.io on 127.0.0.1:53: dial udp: lookup 1.1.1.1:53: no such host
2024-02-18T00:39:04+01:00 WARN [updater] reached the maximum number of consecutive failures: 2 failed attempts resolving hkg-001.vpn.privado.io: lookup hkg-001.vpn.privado.io on 127.0.0.1:53: dial udp: lookup 1.1.1.1:53: no such host
2024-02-18T00:39:04+01:00 WARN [updater] reached the maximum number of consecutive failures: 2 failed attempts resolving sea-020.vpn.privado.io: lookup sea-020.vpn.privado.io on 127.0.0.1:53: dial udp: lookup 1.1.1.1:53: no such host
[...]

PR #3071, opened 2026-01-01
feat(privado): add URL override for authenticated server list access

kldzj:
# Description

Adds `PRIVADO_OVPN_URL` environment variable to allow overriding the default Privado config URL. This enables access to the full authenticated server list instead of the limited public list, by self-hosting the full list and accessing it through the updater.

# Issue

The default `https://privadovpn.com/apps/ovpn_configs.zip` endpoint only returns a subset of servers when unauthenticated.

#2876 #2118

# Assertions


qdm12:
Thanks @kldzj!
However privado implemented a json api to get their servers data, and gluetun now uses that to get their servers (no more zip files) so closing this in favor of https://github.com/qdm12/gluetun/commit/4b8dc8ded7219bc99710b9282cfe170a225acfc8

Issue #3093, opened 2026-01-16
Bug: WireGuard healthcheck fails on older kernels after v3.38.0 (iptables-nft incompatibility)


jarlbrak:
### Is this urgent?

None

### Host OS

Synology DSM 7.x (DS918+) - Linux kernel 4.4.180+

### CPU arch

x86_64

### VPN service provider

Custom

### What are you using to run the container

docker-compose

### What is the version of Gluetun

Latest (qmcgaw/gluetun:latest built 2025-12-29) - Issue occurs in versions after v3.38.0

## Summary

WireGuard connections fail on systems with older Linux kernels (specifically kernel 4.4.x on Synology NAS) when using gluetun versions newer than v3.38.0. The VPN fails healthchecks with "TLS handshake: context deadline exceeded" errors, even though the WireGuard setup completes without errors.

## Root Cause Analysis

The issue appears to be caused by the Alpine upgrade in v3.39.0 which changed the iptables backend from `iptables-legacy` to `iptables-nft`. Older kernels (like Synology's 4.4.180) may not fully support the nft backend, causing network rules to not function correctly even though no errors are reported.

## Evidence

- **v3.38.0 works perfectly** - VPN connects, healthcheck passes, traffic flows through tunnel
- - **v3.39.0+ fails** - WireGuard setup completes but no traffic passes through tunnel, healthcheck times out
When the issue occurs:
- Ping to VPN server from host works
- - WireGuard interface is created inside container
- - - But ping to 8.8.8.8 from inside container shows 100% packet loss
- - - - Healthcheck TLS handshake to github.com times out
## Workaround

Rolling back to v3.38.0 immediately resolves the issue.

## Suggested Fix

Consider adding an option to force `iptables-legacy` backend for systems with older kernels, or detect kernel version and automatically select the appropriate backend.

### Share your logs (at least 10 lines)

```plain text
2026-01-16T03:50:06Z INFO [routing] default route found: interface eth0, gateway 172.18.0.1, assigned IP 172.18.0.3 and family v4
2026-01-16T03:50:06Z INFO [routing] adding route for 0.0.0.0/0
2026-01-16T03:50:06Z INFO [firewall] setting allowed subnets...
2026-01-16T03:50:06Z INFO TUN device is not available: open /dev/net/tun: no such file or directory; creating it...
2026-01-16T03:50:06Z INFO [healthcheck] listening on 127.0.0.1:9999
[...]

PR #3106, opened 2026-01-24
feat(dns): add DOT_CONNECTION_POOLING environment variable

jarlbrak:
## Summary

- Add `DOT_CONNECTION_POOLING` environment variable to control DNS over TLS connection pooling behavior
- When set to `off` or `false`, a new TCP+TLS connection is created for each DNS query instead of reusing pooled connections
- Default is `on` (pooling enabled) for backwards compatibility

## Motivation

On systems with older Linux kernels (e.g., Synology NAS devices running kernel 4.4.x), the DNS over TLS connection pooling causes "connection reset by peer" errors and i/o timeouts.

Debug logs show:
```
[dns] getting tls connection for request IN A github.com.: creating connection: running TLS handshake with 1.1.1.1:853 (cloudflare-dns.com): read tcp 10.14.0.2:38266->1.1.1.1:853: read: connection reset by peer
[dns] exchanging over tls connection (2000ms) for request IN A speed.cloudflare.com.: read tcp 10.14.0.2:55048->1.0.0.1:853: i/o timeout
```

Testing with `DOT=off` (falling back to plaintext DNS) confirmed the issue is specific to DoT connection pooling.

This PR depends on: https://github.com/qdm12/dns/pull/152

Fixes: #3093

## Test plan

- [x] Verified code compiles
- [x] Verified `DOT=off` workaround works on affected system
- [ ] Integration test with `DOT_CONNECTION_POOLING=off` on affected Synology NAS

### DIFF
--- a/internal/configuration/settings/dns.go
+++ b/internal/configuration/settings/dns.go
@@ -53,6 +53,13 @@ type DNS struct {
 	// It defaults to false and cannot be nil in the
 	// internal state.
 	KeepNameserver *bool
+	// DoTPooling enables or disables connection pooling for DNS
+	// over TLS. When set to false, a new TCP+TLS connection is
+	// created for each DNS query. This can help on systems where
+	// TCP connection reuse causes issues (e.g., older Linux kernels).
+	// It defaults to true (pooling enabled) and cannot be nil in
+	// the internal state.
+	DoTPooling *bool `json:"dot_pooling"`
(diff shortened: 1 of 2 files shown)


qdm12:
This is likely not going to fix the situation so closing this for the time being. Happy to re-open and continue work here if this turns out to be needed.

PR #3107, opened 2026-01-24
chore(all): replace netlink library for more flexibility

qdm12:
# Description

- drop usage of github.com/vishvananda/netlink
- use github.com/mdlayher/genetlink and github.com/jsimonetti/rtnetlink
- these new libraries allow more control especially to tweak conntrack in the upcoming changes
- change netlinker Link methods to accept a `linkIndex uint32` instead of an entire `netlink.Link` struct.
- change MTU type from uint16 to uint32 and family type from int to uint8 to match netlink kernel messages
- move types from internal/netlink/types.go to their respective file in internal/netlink

# Issue

# Assertions

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,7 +59,7 @@ jobs:
       - name: Run tests in test container
         run: |
           touch coverage.txt
-          docker run --rm --device /dev/net/tun \
+          docker run --rm --cap-add=NET_ADMIN --device /dev/net/tun \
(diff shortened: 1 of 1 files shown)


qdm12:
Merging this since it looks like it's looking fine... and need that rtnetlink library to fix path mtu discovery on the latest image as well 🚀 !

MacNavy:
When running DOT my test box is timing out.  When using this image and DOH, it works fine.
Logs attached.
[PR-3107 Log 012826.txt](https://github.com/user-attachments/files/24921120/PR-3107.Log.012826.txt)

Issue #234, opened 2020-09-07
Feature request: SOCKS5 plaintext proxy in Go


qdm12:
1. What's the feature?

SOCKS5 plaintext proxy in Go

2. Why do you need this feature?

When you don't want to use a Shadowsocks client and don't care about encryption (i.e. in your LAN) or because your browser is limited to SOCKS5 and doesn't have a Shadowsocks extension. Also for low power devices where encryption can have a performance penalty.

3. Extra information?

This should be added in Go and should be relatively straight forward to implement.

alpe12:
Ouch, 2020.
Please.

qdm12:
@alpe12 you can plug in any socks server container for now (dante or other server implementing the socks protocol).

Right now a SOCKS5 proxy built-in is not a priority since there is already an http proxy and a shadowsocks server (encrypted socks5)

alpe12:
To anyone who might be interested. This is what I'm using now to get socks5:

```
version: "3"
services:
  gluetun:
    image: qmcgaw/gluetun
    container_name: gluetun
    cap_add:
      - NET_ADMIN
    ports:
      - 8888:8888/tcp # HTTP proxy
      - 1080:1080 #socks5 proxy
    volumes:
      - ./config:/gluetun
    environment:
      - VPNSP=nordvpn
      - VPN_TYPE=openvpn
      - OPENVPN_USER=>USER<
      - OPENVPN_PASSWORD=>PASSWORD<
      - REGION=Brazil
      - OPENVPN_PROTOCOL=udp
      - OPENVPN_FLAGS=--auth-nocache
      - TZ=America/Sao_Paulo
      - HTTPPROXY=on
  socks5:
    image: serjs/go-socks5-proxy
    depends_on:
      - gluetun
    network_mode: "service:gluetun"
```

timdonovanuk:
Very frustrating no inbuilt socks5. All clients support it.

x0st:
If anybody still needs this, I forked gluetun and added support for socks5.

https://github.com/x0st/gluetun

```
docker run -it --rm --cap-add=NET_ADMIN -e VPN_SERVICE_PROVIDER=... -e OPENVPN_USER=... -e OPENVPN_PASSWORD=... -e SOCKS5=on -e SOCKS5_PASSWORD=123 -p 1080:1080 gluetun
```

You can pass any username, only password will be checked.

```
curl -X GET https://github.com -x socks5://user:123@127.0.0.1:1080
```

[...]

PR #3133, opened 2026-02-11
feat: add socks5 proxy server

angelakis:
# Description

I would like to add simple socks5 functionality. I know shadowsocks exists already but not all tools can use it. It's a long requested [feature](https://github.com/qdm12/gluetun/issues/234). 

I am not a Go developer and I implemented this by adding a new dependency that maybe is not the preferred way considering you have implemented an ss-server to use here for shadowsocks. The [dependency I added](https://github.com/armon/go-socks5) is ancient but seems stable enough and with no obvious (to me) flaws. I can change it to something else if needed. But anyways, as I implemented the feature and I am actively using it, I thought of creating a PR.  

If it's going to be accepted I can create a PR for the wiki also.

# Issue

https://github.com/qdm12/gluetun/issues/234

# Assertions

* [X] I am aware that we do not accept manual changes to the servers.json file
* [X] I am aware that any changes to settings should be reflected in the [wiki](https://github.com/qdm12/gluetun-wiki/)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -196,6 +196,14 @@ ENV VPN_SERVICE_PROVIDER=pia \
     HTTPPROXY_PASSWORD= \
     HTTPPROXY_USER_SECRETFILE=/run/secrets/httpproxy_user \
     HTTPPROXY_PASSWORD_SECRETFILE=/run/secrets/httpproxy_password \
+    # SOCKS5
+    SOCKS5=off \
+    SOCKS5_LOG=off \
+    SOCKS5_LISTENING_ADDRESS=":1080" \
+    SOCKS5_USER= \
+    SOCKS5_PASSWORD= \
+    SOCKS5_USER_SECRETFILE=/run/secrets/socks5_user \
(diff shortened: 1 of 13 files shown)


qdm12:
Hi @angelakis thanks for the contribution!
I appreciate the work, unfortunately (my bad for the bad communication):

- as you guessed correctly I'm aiming at less dependencies (kind of ruined with protonvpn servers update authentication but oh well - i think i will change that soon)
- i already have a [(wip) socks5 branch](https://github.com/qdm12/gluetun/compare/master...socks5) 😄 re-inventing the wheel I suppose, but at least it's minimal and only-server-code, no external dependency. However it's lacking the newer loop code you've put so I will just hijack your branch, merge in the socks5 branch in yours so you get some credits for the grunt work done! I will re-check my socks5 implementation is done and works find, it has been a few years since I wrote this and forgot about it 🫨

Note to self:
- update readme, labels, wiki
- document new options in pr

qdm12:
closed in favor of no-dependency #3336

Issue #2365, opened 2024-07-24
Feature request: Amnezia WireGuard configuration support


marco570:
### What's the feature 🧐

Hi! Could you please add Amnezia WireGuard config support. 
Amnezia WireGuard config has additional properties
```
[Interface]
Address = ***
PrivateKey = ***
DNS = ***
MTU = ***
Jc = between 1-128
Jmin = must be less than Jmax
Jmax = less than 1280
S1 = 0
S2 = 0
H1 = 1
H2 = 2
H3 = 3
H4 = 4
```
https://zaneschepke.com/wgtunnel-docs/features.html#amneziawg-support

This feature was added to android wgtunnel app 
https://github.com/zaneschepke/wgtunnel/issues/149

marco570:
Can you add it (in high priority) for ProtonVPN first? 
Because Amnezia WireGuard config compatible with ProtonVPN.

codemistake:
I'm very interested in this functionality, I'll be waiting

avt00:
This is one of the only protocols available in countries with heavy censorship. It's not just for Proton, but for self-hosted solutions too. It would be great if it got higher priorities

qdm12:
@Zhurik did an amazing job with a PR for this https://github.com/qdm12/gluetun/pull/3150 as well as a wiki PR to add this to Gluetun 💯 🎖️ 

You can try it by building it with `docker build -t qmcgaw/gluetun https://github.com/Zhurik/gluetun.git#feature-add-amnezia-wg` and using `WIREGUARD_IMPLEMENTATION=amneziawg` and using this documentation: https://github.com/Zhurik/gluetun-wiki/blob/d7d8e8079e0aa8a524b8af344e072b948186c9af/setup/options/wireguard.md#amnezia-wg

Comments welcome on #3150 whilst the PR isn't merged yet.

DX37:
> [@Zhurik](https://github.com/Zhurik) did an amazing job with a PR for this [#3150](https://github.com/qdm12/gluetun/pull/3150) as well as a wiki PR to add this to Gluetun 💯 🎖️
> 
[...]

PR #3150, opened 2026-02-22
feat(wireguard): amneziawg implementation

Zhurik:
# Description

---
Edit by qdm12

Build it with docker + git at:

```
docker build -t qmcgaw/gluetun https://github.com/Zhurik/gluetun.git#feature-add-amnezia-wg
```

And then define parameters in your wireguard config or use new options such as `WIREGUARD_J1` (see Dockerfile)

---

For me this is basically the only working protocol and I saw some issues where people were asking for it

Maybe it would be correct to migrate all wireguard logic to [amnezia fork](https://github.com/amnezia-vpn/amneziawg-go). As far as I understand it's backwards compatible with vanilla wireguard but have some commits behind.

Anyways looking for suggestions!

# Issue

#2365 

# Assertions

### DIFF
--- a/internal/configuration/settings/amneziawg.go
+++ b/internal/configuration/settings/amneziawg.go
@@ -3,6 +3,8 @@ package settings
 import (
 	"errors"
 	"fmt"
+	"strconv"
+	"strings"
 
 	"github.com/qdm12/gosettings"
 	"github.com/qdm12/gosettings/reader"
@@ -171,19 +173,55 @@ var (
 	ErrJunkPacketBounds       = errors.New("junk packet minimum must be lower than or equal to maximum")
 	ErrJunkPacketMinMaxNotSet = errors.New("junk packet min and max must be set when junk packet count is set")
 	ErrJunkPacketCountNotSet  = errors.New("junk packet count must be set when junk packet min or max is set")
+	ErrHeaderRangeMalformed   = errors.New("header range is malformed")
 )
 
 func (s AmneziaWg) validate() error {
-	switch {
-	case *s.JunkPacketMax != 0 && *s.JunkPacketMin > *s.JunkPacketMax:
-		return fmt.Errorf("%w: jmin=%d and jmax=%d",
-			ErrJunkPacketBounds, s.JunkPacketMin, s.JunkPacketMax)
-	case *s.JunkPacketCount == 0 && (*s.JunkPacketMin != 0 || *s.JunkPacketMax != 0):
-		return fmt.Errorf("%w: jc=%d and jmin=%d and jmax=%d",
-			ErrJunkPacketCountNotSet, s.JunkPacketCount, *s.JunkPacketMin, *s.JunkPacketMax)
-	case *s.JunkPacketCount != 0 && (*s.JunkPacketMin == 0 || *s.JunkPacketMax == 0):
-		return fmt.Errorf("%w: jc=%d and jmin=%d and jmax=%d",
-			ErrJunkPacketMinMaxNotSet, s.JunkPacketCount, s.JunkPacketMin, s.JunkPacketMax)
+	if *s.JunkPacketCount == 0 {
+		if *s.JunkPacketMin != 0 || *s.JunkPacketMax != 0 {
+			return fmt.Errorf("%w: jc=%d and jmin=%d and jmax=%d",
+				ErrJunkPacketCountNotSet, s.JunkPacketCount, *s.JunkPacketMin, *s.JunkPacketMax)
+		}
+	} else {
+		if *s.JunkPacketMin == 0 || *s.JunkPacketMax == 0 {
+			return fmt.Errorf("%w: jc=%d and jmin=%d and jmax=%d",
+				ErrJunkPacketMinMaxNotSet, s.JunkPacketCount, *s.JunkPacketMin, *s.JunkPacketMax)
+		} else if *s.JunkPacketMin > *s.JunkPacketMax {
+			return fmt.Errorf("%w: jmin=%d and jmax=%d",
+				ErrJunkPacketBounds, *s.JunkPacketMin, *s.JunkPacketMax)
+		}
+	}
+
+	nameToHeaderRange := map[string]string{
+		"h1": *s.HeaderH1,
+		"h2": *s.HeaderH2,
+		"h3": *s.HeaderH3,
+		"h4": *s.HeaderH4,
+	}
+	for name, headerRange := range nameToHeaderRange {
+		if headerRange == "" {
+			continue
+		}
+		fields := strings.Split(headerRange, "-")
+		switch len(fields) {
+		case 1:
+			_, err := strconv.Atoi(fields[0])
+			if err != nil {
+				return fmt.Errorf("%w: %s value %s is not a number",
+					ErrHeaderRangeMalformed, name, headerRange)
+			}
+		case 2: //nolint:mnd
+			for _, field := range fields {
+				_, err := strconv.Atoi(field)
+				if err != nil {
+					return fmt.Errorf("%w: %s value %s is not a valid range",
+						ErrHeaderRangeMalformed, name, headerRange)
+				}
+			}
+		default:
+			return fmt.Errorf("%w: %s value %s must be in the form n or n-m",
+				ErrHeaderRangeMalformed, name, headerRange)
+		}
 	}
 
 	return nil


Zhurik:
Related PR in wiki - https://github.com/qdm12/gluetun-wiki/pull/179

go.mod
```diff
@@ -4,14 +4,14 @@ go 1.25.0
 
 require (
 	github.com/ProtonMail/go-srp v0.0.7
+	github.com/amnezia-vpn/amneziawg-go v0.0.0-20251219021448-449d7cffd4ad
```
qdm12: let's pin it to tag [v0.2.16](https://github.com/amnezia-vpn/amneziawg-go/releases/tag/v0.2.16)

go.mod
```diff
@@ -56,10 +56,11 @@ require (
 	github.com/prometheus/procfs v0.15.1 // indirect
 	github.com/qdm12/goservices v0.1.1-0.20251104135713-6bee97bd4978 // indirect
 	github.com/riobard/go-bloom v0.0.0-20200614022211-cdc8013cb5b3 // indirect
+	github.com/ti-mo/conntrack v0.6.0
```
qdm12: if you can rebase on the master branch, sorry I may had forced-push to override a bad commit recently, my bad!

Dockerfile
```diff
@@ -112,6 +112,23 @@ ENV VPN_SERVICE_PROVIDER=pia \
     WIREGUARD_ADDRESSES_SECRETFILE=/run/secrets/wireguard_addresses \
     WIREGUARD_MTU= \
     WIREGUARD_IMPLEMENTATION=auto \
+    # Wireguard AmneziaWG userspace obfuscation (requires WIREGUARD_IMPLEMENTATION=amneziawg)
+    WIREGUARD_JC= \
```
qdm12: Maybe let's prefix all these with `AMNEZIAWG` instead of `WIREGUARD`, since it's a different protocol?

.gitignore
```diff
@@ -1 +1,2 @@
 scratch.txt
+.DS_Store
```
qdm12, line 2:
```suggestion
```

internal/wireguard/settings_test.go
```diff
@@ -271,6 +272,28 @@ func Test_Settings_Check(t *testing.T) {
 			errWrapped: ErrImplementationInvalid,
 			errMessage: "invalid implementation: x",
 		},
+		"amnezia settings with non amnezia implementation": {
```
qdm12: nit I guess I should write my own linter, but change this to use underscores so that it's more easily searchable given the go test tooling converts spaces to underscores when running the subtests:


```suggestion
		"amnezia_settings_with_non_amnezia_implementation": {
```

Probably change the test case too or remove it, since it should be now outdated with other suggestions

internal/wireguard/run.go
```diff
@@ -6,10 +6,13 @@ import (
...
+	amneziaConn "github.com/amnezia-vpn/amneziawg-go/conn"
+	amneziaDevice "github.com/amnezia-vpn/amneziawg-go/device"
+	amneziaTun "github.com/amnezia-vpn/amneziawg-go/tun"
 	"github.com/qdm12/gluetun/internal/netlink"
-	"golang.zx2c4.com/wireguard/conn"
-	"golang.zx2c4.com/wireguard/device"
-	"golang.zx2c4.com/wireguard/tun"
+	wgConn "golang.zx2c4.com/wireguard/conn"
+	wgDevice "golang.zx2c4.com/wireguard/device"
+	wgTun "golang.zx2c4.com/wireguard/tun"
```
qdm12: nit use all lowercase for package aliases so i.e. `wgtun`

internal/wireguard/log.go
```diff
@@ -1,7 +1,8 @@
 package wireguard
 
 import (
-	"golang.zx2c4.com/wireguard/device"
+	amneziaDevice "github.com/amnezia-vpn/amneziawg-go/device"
+	wgDevice "golang.zx2c4.com/wireguard/device"
```
qdm12: use lowercase for package aliases

internal/amneziawg/settings.go
```diff
@@ -0,0 +1,127 @@
...
+	if s.InitPacketI4 != "" {
+		lines = append(lines, "i4="+s.InitPacketI4)
+	}
+	if s.InitPacketI5 != "" {
+		lines = append(lines, "i5="+s.InitPacketI5)
+	}
+
+	if len(lines) == 0 {
+		return ""
+	}
```
qdm12: nit unneeded

```suggestion
```

internal/amneziawg/settings.go
```diff
@@ -0,0 +1,127 @@
...
+			s.JunkPacketMin,
+			s.JunkPacketMax,
+		)
+	}
+
+	return nil
+}
+
+func (s Settings) UAPIConfig() string {
+	var lines []string
```
qdm12: let's use a `map[string]uint16` mapping parameter name (i.e. `jc`) to its value (i.e. `s.JunkPacketCount`), and then iterate with a for loop over it.

Same for the string types, using `map[string]string` below

internal/amneziawg/settings.go
```diff
@@ -0,0 +1,127 @@
...
+	InitPacketI4    string
+	InitPacketI5    string
+}
+
+func (s Settings) IsZero() bool {
+	return s == Settings{}
+}
+
+var (
+	ErrJunkPacketBounds = errors.New("junk packet minimum must be lower or equal than maximum")
```
qdm12:
```suggestion
	ErrJunkPacketBounds = errors.New("junk packet minimum must be lower than or equal to maximum")
```

internal/amneziawg/settings.go
```diff
@@ -0,0 +1,127 @@
...
+	if s.JunkPacketCount != 0 && (s.JunkPacketMin == 0 || s.JunkPacketMax == 0) {
+		return fmt.Errorf(
+			"%w: jc=%d and jmin=%d and jmax=%d",
+			ErrJunkPacketSize,
+			s.JunkPacketCount,
+			s.JunkPacketMin,
+			s.JunkPacketMax,
+		)
+	}
+
```
qdm12: nit changing to switch case + more condense for less lines


```suggestion
	switch {
	case s.JunkPacketMax != 0 && s.JunkPacketMin > s.JunkPacketMax:
			return fmt.Errorf("%w: jmin=%d and jmax=%d",
					ErrJunkPacketBounds, s.JunkPacketMin, s.JunkPacketMax)
	case s.JunkPacketCount == 0 && (s.JunkPacketMin != 0 || s.JunkPacketMax != 0):
			return fmt.Errorf("%w: jc=%d and jmin=%d and jmax=%d",
					ErrJunkPacketCount, s.JunkPacketCount, s.JunkPacketMin, s.JunkPacketMax)
	case s.JunkPacketCount != 0 && (s.JunkPacketMin == 0 || s.JunkPacketMax == 0):
			return fmt.Errorf("%w: jc=%d and jmin=%d and jmax=%d",
					ErrJunkPacketSize, s.JunkPacketCount, s.JunkPacketMin, s.JunkPacketMax)
	}

```

internal/amneziawg/settings.go
```diff
@@ -0,0 +1,127 @@
...
+	InitPacketI5    string
+}
+
+func (s Settings) IsZero() bool {
+	return s == Settings{}
+}
+
+var (
+	ErrJunkPacketBounds = errors.New("junk packet minimum must be lower or equal than maximum")
+	ErrJunkPacketSize   = errors.New("junk packet min and max must be set when junk packet count is set")
```
qdm12: rename to `ErrJunkPacketMinMaxNotSet`

internal/amneziawg/settings.go
```diff
@@ -0,0 +1,127 @@
...
+	ErrJunkPacketCount  = errors.New("junk packet count must be set when junk packet min or max is set")
```
qdm12: rename to `ErrJunkPacketCountNotSet`

internal/wireguard/config.go
```diff
@@ -24,6 +24,15 @@ func configureDevice(client *wgctrl.Client, settings Settings) (err error) {
 	return nil
 }
 
+func configureAmneziaDevice(device userspaceDevice, settings Settings) (err error) {
+	uapiConfig := settings.AmneziaWG.UAPIConfig()
+	err = device.IpcSet(uapiConfig)
+	if err != nil {
+		return fmt.Errorf("applying amneziawg userspace settings: %w", err)
```
qdm12: The caller should wrap the error from this function with `configuring Amnezia device: %w`, here we can wrap with something more precise:

```suggestion
		return fmt.Errorf("setting uapi config on ipc: %w", err)
```

internal/wireguard/run.go
```diff
@@ -29,6 +32,19 @@ var (
...
+	IpcHandle(net.Conn)
+	Wait() chan struct{}
+	Close()
+	IpcSet(string) error
+}
+
+var (
+	_ userspaceDevice = (*wgDevice.Device)(nil)
+	_ userspaceDevice = (*amneziaDevice.Device)(nil)
 )
```
qdm12: I don't think there is a need for that, compilation would fail elsewhere in this package if either one of the implementations does not implement the interface.

That is useful in case we have a package exposing an implementation but the interface is somewhere else.

internal/wireguard/settings.go
```diff
@@ -8,10 +8,34 @@ import (
...
+type WgImplementation string
+
+const (
+	WgAuto        WgImplementation = "auto"
+	WgUserspace   WgImplementation = "userspace"
+	WgKernelspace WgImplementation = "kernelspace"
+	WgAmnezia     WgImplementation = "amneziawg"
+)
+
+func ParseImplementation(s string) WgImplementation {
```
qdm12: I appreciate the minor improvement, but let's revert to have it with dumb strings as it was.
The reason it was a string is that it's an internal package, the implementation string is already validated in `internal/configuration/settings` and I don't want custom type aliases in `internal/configuration/settings` (annoying to encode & various other annoyances).

Plus it does feel a bit out of scope for this PR, but I'm being picky here 😄

internal/wireguard/settings.go
```diff
@@ -178,11 +206,20 @@ func (s *Settings) Check() (err error) {
...
-	case "auto", "kernelspace", "userspace":
+	case WgAuto, WgKernelspace, WgUserspace, WgAmnezia:
 	default:
 		return fmt.Errorf("%w: %s", ErrImplementationInvalid, s.Implementation)
 	}
 
+	if s.Implementation != WgAmnezia && !s.AmneziaWG.IsZero() {
+		return fmt.Errorf("%w", ErrImplementationAmneziaWG)
+	}
+
```
qdm12: we can have amneziawg settings set if the implementation is different (i.e. users may later switch implementation, but keep their amnezia settings set in memory)

```suggestion
```

and remove also `ErrImplementationAmneziaWG`

internal/wireguard/settings.go
```diff
@@ -267,8 +304,12 @@ func (s Settings) ToLines(settings ToLinesSettings) (lines []string) {
...
 	}
 
-	if s.Implementation != "auto" {
-		lines = append(lines, fieldPrefix+"Implementation: "+s.Implementation)
+	if s.Implementation != WgAuto {
+		lines = append(lines, fieldPrefix+"Implementation: "+string(s.Implementation))
+	}
+
+	if !s.AmneziaWG.IsZero() {
+		lines = append(lines, fieldPrefix+"AmneziaWG obfuscation: enabled")
```
qdm12: Would there be a point in having zero-ed settings with amneziawg?

---

If there is no point, let's revert this, I don't see what's the added value of having

```
Implementation: amneziawg
AmneziaWG obfuscation: enabled
```

versus just `Implementation: amneziawg`

---

If there is a point, let's change this to the following, such that we print out amnezia obfuscation details only if the implementation is amneziawg.

```suggestion
	switch s.Implementation {
	case "auto":
	case "amneziawg":
		lines = append(lines, fieldPrefix+"Implementation: "+s.Implementation)
		if !s.AmneziaWG.IsZero() {
			lines = append(lines, fieldPrefix+"AmneziaWG obfuscation: enabled")
		}
	default:
		lines = append(lines, fieldPrefix+"Implementation: "+s.Implementation)	
	}
```

internal/wireguard/run.go
```diff
@@ -252,10 +362,10 @@ func setupUserSpace(ctx context.Context,
 		return err
 	}
 
-	return link.Index, waitAndCleanup, nil
+	return link.Index, waitAndCleanup, device, nil
```
qdm12: nit sorry being very picky here, can you move the returned `device` as the 2nd argument? I like my cleanup functions to be the last non-error argument returned 😉 Thanks!!

internal/wireguard/run.go
```diff
@@ -175,43 +206,122 @@ func setupKernelSpace(ctx context.Context,
...
+		<-uapiAcceptErrorCh // wait for acceptAndHandle to exit
+
+		return err
+	}
+
+	return link.Index, waitAndCleanup, nil, nil
+}
+
+//nolint:dupl,ireturn
+func setupAmneziaUserSpace(ctx context.Context,
```
qdm12: Given `setupAmneziaUserSpace` is very similar to `setupUserSpace`, let's have a common `setupUserspaceGeneric` (better naming welcome if you think of anything) with signature

```go
func setupUserSpace(ctx context.Context,
	interfaceName string, netLinker NetLinker, mtu uint32,
	closers *closers, logger Logger,
	createTun func(name string, mtu int) (tunDevice, error),
    newDefaultBind func() bind,
    newDevice(device tunDevice, bind bind, logger Logger) device
) (
	linkIndex uint32, waitAndCleanup waitAndCleanupFunc, device userspaceDevice, err error,
)
```

You would need to have a wrapper for each CreateTun, for example:

```go
type tunDevice interface {
    wgtun.Device
}

func createTunAmnezia(name string, mtu int) (tunDevice, error) {
    return amneziatun.CreateTUN(name, mtu)
}

func createTunWireguard(name string, mtu int) (tunDevice, error) {
    return wgtun.CreateTUN(name, mtu)
}
```

Same thing for the `newDefaultBind` functions. For `newDevice`, similarly:

```go
type bind interface {
    wgconn.Bind
    amneziaconn.Bind // add this if needed eventually
}

type device interface {
    IpcHandle(socket net.Conn)
    Wait() chan struct{}
    Close()
}
```

And you would need a few functions to accept these new unexported local interfaces. You can also remove both `makeDeviceLoggerXYZ` functions.

internal/wireguard/run.go
```diff
@@ -150,8 +178,11 @@ func setupKernelSpace(ctx context.Context,
 		if link.VirtualType == "wireguard" && link.Name == interfaceName {
 			err = netLinker.LinkDel(link.Index)
 			if err != nil {
-				return 0, nil, fmt.Errorf("deleting previous Wireguard link %s: %w",
-					interfaceName, err)
+				return 0, nil, nil, fmt.Errorf(
+					"deleting previous Wireguard link %s: %w",
+					interfaceName,
+					err,
+				)
```
qdm12: nit

```suggestion
				return 0, nil, nil, fmt.Errorf("deleting previous Wireguard link %s: %w",
					interfaceName, err)
```

internal/wireguard/run.go
```diff
@@ -90,6 +109,14 @@ func (w *Wireguard) Run(ctx context.Context, waitError chan<- error, ready chan<
 		return
 	}
 
+	if device != nil {
+		err = configureAmneziaDevice(device, w.settings)
+		if err != nil {
+			waitError <- fmt.Errorf("%w: %s", ErrConfigure, err)
+			return
+		}
+	}
```
qdm12: Can't we stuff this in `setupAmneziaUserSpace`?

internal/wireguard/run.go
```diff
@@ -41,20 +57,23 @@ func (w *Wireguard) Run(ctx context.Context, waitError chan<- error, ready chan<
...
+	case WgUserspace:
+	case WgKernelspace:
 		if !kernelSupported {
 			waitError <- fmt.Errorf("%w", ErrKernelSupport)
 			return
 		}
 		setupFunction = setupKernelSpace
+	case WgAmnezia:
+		setupFunction = setupAmneziaUserSpace
+		w.logger.Info("Using amneziawg userspace implementation")
```
qdm12: nit no need for that, the point of the above logs is just to explain to the user what is picked in auto mode.

```suggestion
```

internal/server/vpn.go
```diff
@@ -94,7 +94,7 @@ func (h *vpnHandler) setStatus(w http.ResponseWriter, r *http.Request) {
 func (h *vpnHandler) getSettings(w http.ResponseWriter) {
 	settings := h.looper.GetSettings()
 	encoder := json.NewEncoder(w)
-	if err := encoder.Encode(settings); err != nil {
+	if err := encoder.Encode(settings); err != nil { //nolint:musttag
```
qdm12: revert this and add some json tags to the amnezia settings struct

internal/server/vpn.go
```diff
@@ -104,7 +104,7 @@ func (h *vpnHandler) getSettings(w http.ResponseWriter) {
 func (h *vpnHandler) patchSettings(w http.ResponseWriter, r *http.Request) {
 	var overrideSettings settings.VPN
 	decoder := json.NewDecoder(r.Body)
-	err := decoder.Decode(&overrideSettings)
+	err := decoder.Decode(&overrideSettings) //nolint:musttag
```
qdm12: revert this and add some json tags to the amnezia settings struct

internal/configuration/sources/files/wireguard_test.go
```diff
@@ -130,10 +139,15 @@ Address = 10.38.22.35/32
 			iniSection, err := iniFile.GetSection("Interface")
 			require.NoError(t, err)
 
-			privateKey, addresses := parseWireguardInterfaceSection(iniSection)
+			privateKey, addresses, jc, _, _, _, _, _, _, h1, _, _, _, i1, _, _, _, _ := parseWireguardInterfaceSection(
```
qdm12: maybe add a little amnezia struct to carry all these fields? that way we can also assert the entire struct versus an expected struct, to make sure we don't set wrong fields

internal/configuration/sources/files/wireguard.go
```diff
@@ -32,6 +32,22 @@ type WireguardConfig struct {
...
+	S4           *string
+	H1           *string
+	H2           *string
+	H3           *string
+	H4           *string
+	I1           *string
+	I2           *string
+	I3           *string
+	I4           *string
+	I5           *string
```
qdm12: Yeah let's add a little amnezia struct here to carry these fields

internal/configuration/settings/wireguard.go
```diff
@@ -245,6 +265,72 @@ func (w *Wireguard) read(r *reader.Reader) (err error) {
...
+
+	s4, err := r.Uint16Ptr("WIREGUARD_S4")
+	if err != nil {
+		return err
+	}
+	if s4 != nil {
+		w.AmneziaWG.PaddingS4 = *s4
+	}
+
+	w.AmneziaWG.HeaderH1 = r.String("WIREGUARD_H1", reader.ForceLowercase(false))
```
qdm12: nice catch on the `reader.ForceLowercase(false)`!

internal/configuration/settings/wireguard.go
```diff
@@ -245,6 +265,72 @@ func (w *Wireguard) read(r *reader.Reader) (err error) {
...
+		w.AmneziaWG.PaddingS3 = *s3
+	}
+
+	s4, err := r.Uint16Ptr("WIREGUARD_S4")
+	if err != nil {
+		return err
+	}
+	if s4 != nil {
+		w.AmneziaWG.PaddingS4 = *s4
+	}
```
qdm12: let's run this in a for loop using a `map[string]*uint16`, for example:

```go
amneziaUint16Params := map[string]*uint16{
    "WIREGUARD_JC": &w.AmneziaWG.JunkPacketCount,
}
```

internal/configuration/settings/wireguard.go
```diff
@@ -245,6 +265,72 @@ func (w *Wireguard) read(r *reader.Reader) (err error) {
 		reader.RetroKeys("WIREGUARD_INTERFACE"), reader.ForceLowercase(false))
 	w.Implementation = r.String("WIREGUARD_IMPLEMENTATION")
 
+	jc, err := r.Uint16Ptr("WIREGUARD_JC")
```
qdm12: Question: is `0` a valid value for `w.AmneziaWG.JunkPacketCount`? If it is, then you need to change its type from `uint16` to `*uint16`. Go-default invalid values are used to detect when a field is not set by the user. In this case, if `0` is a valid possible value, use `*uint16` so its Go-default of nil indicates it's not set. Likewise, if `0` is not a valid possible value, then stick to use `uint16` and not `*uint16`.

Same question for other fields

internal/configuration/settings/wireguard.go
```diff
@@ -167,6 +180,9 @@ func (w *Wireguard) overrideWith(other Wireguard) {
 	w.Interface = gosettings.OverrideWithComparable(w.Interface, other.Interface)
 	w.MTU = gosettings.OverrideWithComparable(w.MTU, other.MTU)
 	w.Implementation = gosettings.OverrideWithComparable(w.Implementation, other.Implementation)
+	if !other.AmneziaWG.IsZero() {
+		w.AmneziaWG = other.AmneziaWG
+	}
```
qdm12: Add an overrideWith method on the AmneziaWG struct, and use that here. Also check my other comment on the zero-value-validity thing below.

internal/configuration/settings/wireguard.go
```diff
@@ -154,6 +166,7 @@ func (w *Wireguard) copy() (copied Wireguard) {
 		Interface:                   w.Interface,
 		MTU:                         w.MTU,
 		Implementation:              w.Implementation,
+		AmneziaWG:                   w.AmneziaWG,
```
qdm12: add a copy method for this struct, since pointers won't deep copy by just doing that. See existing copy methods for example, there are a few helper functions to deep copy

internal/configuration/settings/wireguard.go
```diff
@@ -136,11 +139,20 @@ func (w Wireguard) validate(vpnProvider string, ipv6Supported bool) (err error)
...
+	validImplementations := []string{"auto", "userspace", "kernelspace", "amneziawg"}
 	if err := validate.IsOneOf(w.Implementation, validImplementations...); err != nil {
 		return fmt.Errorf("%w: %w", ErrWireguardImplementationNotValid, err)
 	}
 
+	if w.Implementation != "amneziawg" && !w.AmneziaWG.IsZero() {
+		return fmt.Errorf("%w", ErrWireguardAmneziaImplementation)
+	}
+
+	err = w.AmneziaWG.Validate()
```
qdm12: I think you can unexport Validate as validate

internal/configuration/settings/wireguard.go
```diff
@@ -136,11 +139,20 @@ func (w Wireguard) validate(vpnProvider string, ipv6Supported bool) (err error)
...
 
-	validImplementations := []string{"auto", "userspace", "kernelspace"}
+	validImplementations := []string{"auto", "userspace", "kernelspace", "amneziawg"}
 	if err := validate.IsOneOf(w.Implementation, validImplementations...); err != nil {
 		return fmt.Errorf("%w: %w", ErrWireguardImplementationNotValid, err)
 	}
 
+	if w.Implementation != "amneziawg" && !w.AmneziaWG.IsZero() {
+		return fmt.Errorf("%w", ErrWireguardAmneziaImplementation)
+	}
```
qdm12: let's remove this, let's allow the user to specify amnezia parameters and use a different implementation, so they could swap implementation live later with the control server


```suggestion
```

internal/configuration/settings/wireguard.go
```diff
@@ -46,6 +47,8 @@ type Wireguard struct {
 	// It defaults to "auto" and cannot be the empty string
 	// in the internal state.
 	Implementation string `json:"implementation"`
+	// AmneziaWG contains optional obfuscation parameters
```
qdm12: Are they optional? Should we default them to sane defaults? How does amnezia behave with all these parameters disabled?

internal/configuration/settings/wireguard.go
```diff
@@ -235,6 +251,10 @@ func (w Wireguard) toLinesNode() (node *gotree.Node) {
 		node.Appendf("Implementation: %s", w.Implementation)
 	}
 
+	if !w.AmneziaWG.IsZero() {
+		node.Append("AmneziaWG obfuscation settings: enabled")
+	}
+
```
qdm12: Is this obfuscation disabled if all parameters are not set for amnezia? Why would anyone use amnezia with zero-ed parameters?

Zhurik:
The only provider I know is [AmneziaWG](https://docs.amnezia.org/documentation/amnezia-wg/) themselves. I tried with selfhosted version, but more testing needed

qdm12:
Oooo interesting, maybe another PR to add them as provider would be cool. I've read apparently also Red Shield VPN and NymVPN support it, but they're not part of Gluetun yet. Out of scope for now 😉 
Although we should add a note in the wiki that this is for now only to use with the custom provider, given none support it outside of that, so users aren't tempted to use it and then complain it's broken 😄

qdm12:
> Maybe it would be correct to migrate all wireguard logic to [amnezia fork](https://github.com/amnezia-vpn/amneziawg-go).

Hmm that's a tough question. I'm very tempted, but also this means we depend on amnaziawg not being evil, miss some commits from wireguard-go, and risk a big issue in case they decide to diverge from wireguard-go in terms of retro-compatibility. I would say let's keep it separate as it is!

Zhurik:
> Although we should add a note in the wiki that this is for now only to use with the custom provider, given none support it outside of that, so users aren't tempted to use it and then complain it's broken 😄

Ok, I'll add that

internal/configuration/settings/wireguard.go
Zhurik: It behaves like traditional Wireguard, so yeah, I'll remove `optional`

.gitignore
Zhurik: Maybe add this? The more you have in the gitignore the less likely it is that trash will get through. I was writing code on Mac and this thing always slips in somehow. So why not add it?

Dockerfile
Zhurik: Ok, that's fair

internal/wireguard/settings.go
Zhurik: Yeah, there is actually no point

.gitignore
qdm12: I'm trying to keep it minimal since otherwise people start adding vscode, intellij, etc. git ignore lines.
But you're right, let's yolo it and add it! Please resolve this conversation 😉

Dockerfile
qdm12: Hm I wonder, this may conflict with the wireguard config file parsing, I think we must keep it to WIREGUARD_JC if we plan to extract JC from wireguard configs.

Zhurik: ok, i'll revert)

internal/wireguard/run.go
Zhurik: yeah, already moved

internal/wireguard/run.go
Zhurik: I changhed interface a bit but maybe even managed to refactor those

internal/configuration/settings/wireguard.go
Zhurik: There is no sense in using it without parameters, so I removed it

internal/configuration/settings/wireguard.go
Zhurik: I looked deeper into the docs and they can be 0, so changing to pointers

Zhurik:
Refactored some parts so it looks a bit better now

internal/amneziawg/settings.go
```diff
@@ -0,0 +1,111 @@
...
+	HeaderH3        string  `json:"header_h3"`
+	HeaderH4        string  `json:"header_h4"`
+	InitPacketI1    string  `json:"init_packet_i1"`
+	InitPacketI2    string  `json:"init_packet_i2"`
+	InitPacketI3    string  `json:"init_packet_i3"`
+	InitPacketI4    string  `json:"init_packet_i4"`
+	InitPacketI5    string  `json:"init_packet_i5"`
+}
+
+func (s Settings) Copy() (copied Settings) {
```
qdm12: unexport as `copy()`

internal/wireguard/settings.go
```diff
@@ -83,8 +87,7 @@ func (s *Settings) SetDefaults() {
 	}
 
 	if s.Implementation == "" {
-		const defaultImplementation = "auto"
-		s.Implementation = defaultImplementation
+		s.Implementation = "auto"
```
qdm12: revert to reduce git deltas




```suggestion
		const defaultImplementation = "auto"
		s.Implementation = defaultImplementation
```

internal/wireguard/run.go
```diff
@@ -55,6 +78,30 @@ func (w *Wireguard) Run(ctx context.Context, waitError chan<- error, ready chan<
 			return
 		}
 		setupFunction = setupKernelSpace
+	case "amneziawg":
+		userspaceBackend = userSpaceBackend{
+			createTun: func(name string, mtu int) (tunDevice, error) {
```
qdm12: Pretty awesome! I was thinking of doing something similar, congrats on figuring it out! 💯 !

internal/mod/info_linux.go
qdm12, line 117: Ehh we have a bit of a mixed history here (might be because of me), I took the liberty to force push to your branch to fix the history, and rebased it on the tip of the master branch. You can get the changes with

```sh
git checkout feature-add-amnezia-wg
git fetch origin
git reset --hard origin/feature-add-amnezia-wg
```

I also made a backup of the current branch in case you need it let me know!

internal/configuration/sources/secrets/reader.go
```diff
@@ -37,6 +37,7 @@ func New(warner Warner) (source *Source) {
 
 func (s *Source) String() string { return "secret files" }
 
+//nolint:gocyclo
```
qdm12: nit perhaps split aways the wireguard cases in their own function, so you would have the existing switch cases, and then a call to the function with a switch and the wireguard cases. Then we can remove that //nolint:gocyclo

internal/configuration/sources/files/wireguard_test.go
```diff
@@ -131,9 +140,13 @@ Address = 10.38.22.35/32
 			require.NoError(t, err)
 
 			privateKey, addresses := parseWireguardInterfaceSection(iniSection)
+			amneziaWgConfig := NewAmneziaWgConfigFromInerfaceIniSection(iniSection)
```
qdm12: typo `Inerface` i think

internal/configuration/sources/files/wireguard.go
```diff
@@ -65,9 +107,7 @@ func ParseWireguardConf(path string) (config WireguardConfig, err error) {
 	return config, nil
 }
 
-func parseWireguardInterfaceSection(interfaceSection *ini.Section) (
-	privateKey, addresses *string,
-) {
+func parseWireguardInterfaceSection(interfaceSection *ini.Section) (privateKey, addresses *string) {
```
qdm12: revert to reduce unneeded git diffs


```suggestion
func parseWireguardInterfaceSection(interfaceSection *ini.Section) (
	privateKey, addresses *string,
) {
```

internal/configuration/sources/files/wireguard.go
```diff
@@ -25,13 +25,54 @@ func (s *Source) lazyLoadWireguardConf() WireguardConfig {
...
+	H4   *string
+	I1   *string
+	I2   *string
+	I3   *string
+	I4   *string
+	I5   *string
+}
+
+func NewAmneziaWgConfigFromInerfaceIniSection(interfaceSection *ini.Section) *AmneziaWgConfig {
+	return &AmneziaWgConfig{
```
qdm12: just return a non-pointer struct, I don't think there is a point returning a pointer

internal/configuration/sources/files/wireguard.go
```diff
@@ -25,13 +25,54 @@ func (s *Source) lazyLoadWireguardConf() WireguardConfig {
...
+	H3   *string
+	H4   *string
+	I1   *string
+	I2   *string
+	I3   *string
+	I4   *string
+	I5   *string
+}
+
+func NewAmneziaWgConfigFromInerfaceIniSection(interfaceSection *ini.Section) *AmneziaWgConfig {
```
qdm12: nit rename to `parseWireguardAmneziaInterfaceSection` to match the existing format for `parseWireguardInterfaceSection`? 🤔 

Also this function does not need to be exported, so unexport it 😉

internal/configuration/sources/files/wireguard.go
```diff
@@ -25,13 +25,54 @@ func (s *Source) lazyLoadWireguardConf() WireguardConfig {
 	return s.cached.wireguardConf
 }
 
+type AmneziaWgConfig struct {
```
qdm12: this might not need to be exported, as long as its fields are exported to they can be accessed through `WireguardConfig`

internal/configuration/settings/wireguard.go
```diff
@@ -245,6 +255,72 @@ func (w *Wireguard) read(r *reader.Reader) (err error) {
 		reader.RetroKeys("WIREGUARD_INTERFACE"), reader.ForceLowercase(false))
 	w.Implementation = r.String("WIREGUARD_IMPLEMENTATION")
 
+	jc, err := r.Uint16Ptr("WIREGUARD_JC")
+	if err != nil {
+		return err
+	}
+	if jc != nil {
+		w.AmneziaWG.JunkPacketCount = jc
+	}
```
qdm12: simplify with


```suggestion
	w.AmneziaWG.JunkPacketCount, err = r.Uint16Ptr("WIREGUARD_JC")
	if err != nil {
		return err
	}
```

Same for the other calls below

internal/amneziawg/settings.go
```diff
@@ -0,0 +1,111 @@
...
+	copied.InitPacketI3 = s.InitPacketI3
+	copied.InitPacketI4 = s.InitPacketI4
+	copied.InitPacketI5 = s.InitPacketI5
+	return copied
+}
+
+func (s Settings) IsZero() bool {
+	return s == Settings{}
+}
+
```
qdm12: I don't think that's used anymore


```suggestion
```

internal/amneziawg/settings.go
```diff
@@ -0,0 +1,111 @@
+package amneziawg
```
qdm12: should we move this file in `internal/configuration/settings/amneziawg`? then we can probably unexport some methods such as `Copy()` as well.

internal/amneziawg/settings.go
```diff
@@ -0,0 +1,111 @@
...
+		"h4": s.HeaderH4,
+		"i1": s.InitPacketI1,
+		"i2": s.InitPacketI2,
+		"i3": s.InitPacketI3,
+		"i4": s.InitPacketI4,
+		"i5": s.InitPacketI5,
+	}
+
+	for key, val := range uintFields {
+		if val != nil && *val != 0 {
```
qdm12: if we don't do anything with them if they're `0`, I think it's safe to say `0` is not a valid value right, in which case we should change back from `*uint16` to `uint16` for these fields? Or should we still add `key=0` to the config?

internal/configuration/settings/wireguard.go
```diff
@@ -167,6 +176,7 @@ func (w *Wireguard) overrideWith(other Wireguard) {
 	w.Interface = gosettings.OverrideWithComparable(w.Interface, other.Interface)
 	w.MTU = gosettings.OverrideWithComparable(w.MTU, other.MTU)
 	w.Implementation = gosettings.OverrideWithComparable(w.Implementation, other.Implementation)
+	w.AmneziaWG = gosettings.OverrideWithComparable(w.AmneziaWG, other.AmneziaWG)
```
qdm12: you need to implement an overrideWith method, i.e.:

https://github.com/Zhurik/gluetun/blob/575561ff3b6cfd324f98396d54d810968155f3d3/internal/configuration/settings/wireguard.go#L169

internal/configuration/settings/wireguard.go
```diff
@@ -167,6 +176,7 @@ func (w *Wireguard) overrideWith(other Wireguard) {
...
 }
 
 func (w *Wireguard) setDefaults(vpnProvider string) {
```
qdm12, line 181: we should set defaults to the amnesiawg config (maybe `0`?). The invalid value (`0` or `nil` for uint16 fields) cannot stay "invalid" after the setDefaults call

internal/configuration/settings/wireguard.go
qdm12, line 250: If the implementation is selected as amneziawg, please create an amnezia tree node for its set/non-zero parameters 😉

internal/wireguard/common.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+type userSpaceBackend struct {
+	createTun    func(string, int) (tunDevice, error)
+	createBind   func() bind
+	createDevice func(tunDevice, bind, Logger) userspaceDevice
+	preStart     func(userspaceDevice, Settings) error
+}
+
+func setupUserSpaceCommon(ctx context.Context,
```
qdm12: can you move back that function and `acceptAndHandle` at the bottom of internal/wireguard/run.go as to reduce git diffs? Thanks! 👍

internal/wireguard/run.go
```diff
@@ -39,7 +42,27 @@ func (w *Wireguard) Run(ctx context.Context, waitError chan<- error, ready chan<
 		return
 	}
 
-	setupFunction := setupUserSpace
+	userspaceBackend := userSpaceBackend{
```
qdm12: nit maybe we can move these 2 userSpaceBackend definitions to a `userspace.go` file, or to `userspace.go` and `amnezia.go` files?

internal/wireguard/run.go
```diff
@@ -55,6 +78,30 @@ func (w *Wireguard) Run(ctx context.Context, waitError chan<- error, ready chan<
...
+					Verbosef: logger.Debugf,
+					Errorf:   logger.Errorf,
+				}
+				device := amneziadevice.NewDevice(wgamneziaTun, wgamneziaBind, &wgamneziaLogger)
+				return device
+			},
+			preStart: func(ud userspaceDevice, s Settings) error {
+				uapiConfig := s.AmneziaWG.UAPIConfig()
+				err = ud.IpcSet(uapiConfig)
+				return err
```
qdm12: nit just being careful, this uses `err` from the outer scope, a bit risky!


```suggestion
				return ud.IpcSet(uapiConfig)
```

qdm12:
ℹ️ apologies but I think the git history was all messed up, possibly due to me force pushing like an idiot on the master branch...

I took the liberty to force push to your branch to fix the history, and rebased it on the tip of the master branch. You can get the changes with

```sh
git checkout feature-add-amnezia-wg
git fetch origin
git reset --hard origin/feature-add-amnezia-wg
```

I also made a backup of the current branch in case you need it let me know!

internal/amneziawg/settings.go
Zhurik: From the [docs](https://github.com/amnezia-vpn/amneziawg-go?tab=readme-ov-file#configuration):

> If there is no value specified (for any param), AWG treats it as 0

So there is basically no difference if we specify them as 0 or don't specify them at all

Zhurik: ok, I'll revert them back to uint16

internal/configuration/settings/wireguard.go
Zhurik: I'll set them as zero

internal/configuration/sources/secrets/reader.go
Zhurik: I tried to refactor this but it becomes worse(

Zhurik:
- split amnezia into their own structs with their own set of methods
- left validation in one place
- and other multiple fixes

internal/configuration/settings/wireguard.go
qdm12: (sorry to reconsider this again) there is a use case for the zero value `0`/`""`: for example someone may want to set J1=1 but leave J2=0.

---

if the default is `0` and is a valid value, we need these fields as pointers to uint16 so their "Go default" is `nil`, to distinguish between the default value (`0`) and the unset value (`nil`).
Same for the string values, if their default is `""`, then we want them as string pointers too.

Then you need to:

- explicitely set these fields to their default (0, "") in a new function `func (s *AmneziaWg) setDefaults()` that call from `func (w *Wireguard) setDefaults(vpnProvider string)` - you can use `gosettings.DefaultPointer`
- copy these fields using in the `func (s AmneziaWg) copy() (copied AmneziaWg)` using `gosettings.CopyPointer(x)`
- override these fields in the `func (s *AmneziaWg) overrideWith(other AmneziaWg)` using `gosettings.OverrideWithPointer(a, b)`
- in `func (s *AmneziaWg) toLinesNode` and `func (s *AmneziaWg) validate`, you don't need to add nil checks, you can assume all fields are set (to their defaults or user specified values) when calling these functions, so directly dereference them i.e. `*s.JunckPacketCount`
- in `func (s *AmneziaWg) toLinesNode()`, remove the `f.val != 0` and `f.val != ""` checks; both are valid default values and should be logged out.

internal/configuration/settings/wireguard.go
```diff
@@ -245,6 +258,51 @@ func (w *Wireguard) read(r *reader.Reader) (err error) {
...
+
+	w.AmneziaWG.HeaderH1 = r.String("WIREGUARD_H1", reader.ForceLowercase(false))
+	w.AmneziaWG.HeaderH2 = r.String("WIREGUARD_H2", reader.ForceLowercase(false))
+	w.AmneziaWG.HeaderH3 = r.String("WIREGUARD_H3", reader.ForceLowercase(false))
+	w.AmneziaWG.HeaderH4 = r.String("WIREGUARD_H4", reader.ForceLowercase(false))
+	w.AmneziaWG.InitPacketI1 = r.String("WIREGUARD_I1", reader.ForceLowercase(false))
+	w.AmneziaWG.InitPacketI2 = r.String("WIREGUARD_I2", reader.ForceLowercase(false))
+	w.AmneziaWG.InitPacketI3 = r.String("WIREGUARD_I3", reader.ForceLowercase(false))
+	w.AmneziaWG.InitPacketI4 = r.String("WIREGUARD_I4", reader.ForceLowercase(false))
+	w.AmneziaWG.InitPacketI5 = r.String("WIREGUARD_I5", reader.ForceLowercase(false))
```
qdm12: move these to amneziawg.go as its own `func (s *AmneziaWg) read(r *reader.Reader) (err error)` which would then be called from `func (w *Wireguard) read(r *reader.Reader) (err error)`. You can also for loop these with a `map[string]*string` and `map[string]*uint16` I think

Dockerfile
```diff
@@ -112,6 +112,23 @@ ENV VPN_SERVICE_PROVIDER=pia \
...
+    WIREGUARD_S4= \
+    WIREGUARD_H1= \
+    WIREGUARD_H2= \
+    WIREGUARD_H3= \
+    WIREGUARD_H4= \
+    WIREGUARD_I1= \
+    WIREGUARD_I2= \
+    WIREGUARD_I3= \
+    WIREGUARD_I4= \
+    WIREGUARD_I5= \
```
qdm12: Sorry to re-iterate on that, re-reading the code I think it's fine to have these changed to `AMENZIAWG_XYZ` as long as `internal/configuration/sources/files/reader.go` string cases are changed, please update to use `AMNEZIAWG_` prefixes for the options 🙏 Thanks!

internal/wireguard/amnezia_settings.go
```diff
@@ -0,0 +1,62 @@
...
+	HeaderH4        string
+	InitPacketI1    string
+	InitPacketI2    string
+	InitPacketI3    string
+	InitPacketI4    string
+	InitPacketI5    string
+}
+
+func (s AmneziaSettings) uapiConfig() string {
+	var lines []string
```
qdm12: nit minor note that the linter may complain about: move this below uintField and stringFields and pre-allocate with `lines := make([]string, 0, len(uintFields)+len(stringFields))`

internal/wireguard/amnezia_settings.go
```diff
@@ -0,0 +1,62 @@
...
+		"i2": s.InitPacketI2,
+		"i3": s.InitPacketI3,
+		"i4": s.InitPacketI4,
+		"i5": s.InitPacketI5,
+	}
+
+	for key, val := range uintFields {
+		if val != 0 {
+			lines = append(lines, fmt.Sprintf("%s=%d", key, val))
+		}
```
qdm12: nit If we leave this to zero, does it still work fine? If so, let's write it out even if it's zero. Same for the strings below

internal/configuration/sources/secrets/reader.go
qdm12: Have

```go
	// ...
	case "wireguard_endpoint_port":
		return strPtrToStringIsSet(s.lazyLoadWireguardConf().EndpointPort)
	}

	value, isSet, matched := s.getAmneziaWg(key)
	if matched {
		return value, isSet
	}
```

and a method

```go
func (s *Source) getAmneziaWg(key string) (value string, isSet, matched bool) {
	switch key {
	case "wireguard_jc":
		value, isSet = strPtrToStringIsSet(s.lazyLoadWireguardConf().AmneziaParams.Jc)
	case "wireguard_jmin":
		value, isSet = strPtrToStringIsSet(s.lazyLoadWireguardConf().AmneziaParams.Jmin)
	case "wireguard_jmax":
		value, isSet = strPtrToStringIsSet(s.lazyLoadWireguardConf().AmneziaParams.Jmax)
	case "wireguard_s1":
		value, isSet = strPtrToStringIsSet(s.lazyLoadWireguardConf().AmneziaParams.S1)
	case "wireguard_s2":
		value, isSet = strPtrToStringIsSet(s.lazyLoadWireguardConf().AmneziaParams.S2)
	case "wireguard_s3":
		value, isSet = strPtrToStringIsSet(s.lazyLoadWireguardConf().AmneziaParams.S3)
	case "wireguard_s4":
		value, isSet = strPtrToStringIsSet(s.lazyLoadWireguardConf().AmneziaParams.S4)
	case "wireguard_h1":
		value, isSet = strPtrToStringIsSet(s.lazyLoadWireguardConf().AmneziaParams.H1)
	case "wireguard_h2":
		value, isSet = strPtrToStringIsSet(s.lazyLoadWireguardConf().AmneziaParams.H2)
	case "wireguard_h3":
		value, isSet = strPtrToStringIsSet(s.lazyLoadWireguardConf().AmneziaParams.H3)
	case "wireguard_h4":
		value, isSet = strPtrToStringIsSet(s.lazyLoadWireguardConf().AmneziaParams.H4)
	case "wireguard_i1":
		value, isSet = strPtrToStringIsSet(s.lazyLoadWireguardConf().AmneziaParams.I1)
	case "wireguard_i2":
		value, isSet = strPtrToStringIsSet(s.lazyLoadWireguardConf().AmneziaParams.I2)
	case "wireguard_i3":
		value, isSet = strPtrToStringIsSet(s.lazyLoadWireguardConf().AmneziaParams.I3)
	case "wireguard_i4":
		value, isSet = strPtrToStringIsSet(s.lazyLoadWireguardConf().AmneziaParams.I4)
	case "wireguard_i5":
		value, isSet = strPtrToStringIsSet(s.lazyLoadWireguardConf().AmneziaParams.I5)
	default:
		return "", false, false
	}
	return value, isSet, true
}
```

then you can remove `//nolint:gocyclo`

internal/configuration/settings/wireguard.go
Zhurik: Ok, I'll add those values, but won't add them in UAPIconfig

internal/wireguard/amnezia_settings.go
Zhurik: I haven't tested that, I'll check. Ok

Zhurik:
Sorry for delay, fixed code according to comments

internal/configuration/settings/amneziawg.go
```diff
@@ -0,0 +1,186 @@
...
+		"AMNEZIAWG_S3":   &s.PaddingS3,
+		"AMNEZIAWG_S4":   &s.PaddingS4,
+	}
+	for key, dst := range uint16Fields {
+		v, err := r.Uint16Ptr(key)
+		if err != nil {
+			return err
+		}
+		*dst = v
+	}
```
qdm12: nit simplify a tiny bit with / not a big fan of pointers-to-pointers 😄 


```suggestion
	uint16Fields := map[string]*uint16{
		"AMNEZIAWG_JC":   s.JunkPacketCount,
		"AMNEZIAWG_JMIN": s.JunkPacketMin,
		"AMNEZIAWG_JMAX": s.JunkPacketMax,
		"AMNEZIAWG_S1":   s.PaddingS1,
		"AMNEZIAWG_S2":   s.PaddingS2,
		"AMNEZIAWG_S3":   s.PaddingS3,
		"AMNEZIAWG_S4":   s.PaddingS4,
	}
	for key, dst := range uint16Fields {
		v, err := r.Uint16Ptr(key)
		if err != nil {
			return err
		} else if v != nil {
			*dst = *v
		}
	}
```

internal/configuration/settings/amneziawg.go
```diff
@@ -0,0 +1,186 @@
...
+		"AMNEZIAWG_I1": &s.InitPacketI1,
+		"AMNEZIAWG_I2": &s.InitPacketI2,
+		"AMNEZIAWG_I3": &s.InitPacketI3,
+		"AMNEZIAWG_I4": &s.InitPacketI4,
+		"AMNEZIAWG_I5": &s.InitPacketI5,
+	}
+	opt := reader.ForceLowercase(false)
+	for key, dst := range stringFields {
+		*dst = r.Get(key, opt) // *string (nil если не задано)
+	}
```
qdm12:
```suggestion
	stringFields := map[string]*string{
		"AMNEZIAWG_H1": s.HeaderH1,
		"AMNEZIAWG_H2": s.HeaderH2,
		"AMNEZIAWG_H3": s.HeaderH3,
		"AMNEZIAWG_H4": s.HeaderH4,
		"AMNEZIAWG_I1": s.InitPacketI1,
		"AMNEZIAWG_I2": s.InitPacketI2,
		"AMNEZIAWG_I3": s.InitPacketI3,
		"AMNEZIAWG_I4": s.InitPacketI4,
		"AMNEZIAWG_I5": s.InitPacketI5,
	}
	opt := reader.ForceLowercase(false)
	for key, dst := range stringFields {
		v := r.Get(key, opt)
		if v != nil {
			*dst = *v
		}
	}
```

internal/configuration/sources/files/wireguard_test.go
```diff
@@ -131,9 +140,13 @@ Address = 10.38.22.35/32
 			require.NoError(t, err)
 
 			privateKey, addresses := parseWireguardInterfaceSection(iniSection)
+			amneziaWgConfig := parseWireguardAmneziaInterfaceSection(iniSection)
 
 			assert.Equal(t, testCase.privateKey, privateKey)
 			assert.Equal(t, testCase.addresses, addresses)
+			assert.Equal(t, testCase.amneziaParams.Jc, amneziaWgConfig.Jc)
+			assert.Equal(t, testCase.amneziaParams.H1, amneziaWgConfig.H1)
+			assert.Equal(t, testCase.amneziaParams.I1, amneziaWgConfig.I1)
```
qdm12: nit I think this would do, so it also checks all fields of the amnezia struct

```suggestion
			assert.Equal(t, testCase.amneziaParams, amneziaWgConfig)
```

Dockerfile
```diff
@@ -18,7 +18,7 @@ ENV CGO_ENABLED=0
 COPY --from=golangci-lint /bin /go/bin/golangci-lint
 COPY --from=mockgen /bin /go/bin/mockgen
 WORKDIR /tmp/gobuild
-COPY go.mod go.sum ./
+COPY go.mod go.sum ./ß
```
qdm12:
```suggestion
COPY go.mod go.sum ./
```

internal/provider/utils/wireguard.go
```diff
@@ -16,6 +16,7 @@ func BuildWireguardSettings(connection models.Connection,
 	settings.PreSharedKey = *userSettings.PreSharedKey
 	settings.InterfaceName = userSettings.Interface
 	settings.Implementation = userSettings.Implementation
+	settings.AmneziaWG = builsAmneziaWgSettings(userSettings.AmneziaWG)
```
qdm12:
```suggestion
	settings.AmneziaWG = buildAmneziaWgSettings(userSettings.AmneziaWG)
```

internal/provider/utils/wireguard.go
```diff
@@ -55,3 +56,24 @@ func BuildWireguardSettings(connection models.Connection,
 
 	return settings
 }
+
+func builsAmneziaWgSettings(s settings.AmneziaWg) wireguard.AmneziaSettings {
```
qdm12:
```suggestion
func buildAmneziaWgSettings(s settings.AmneziaWg) wireguard.AmneziaSettings {
```

Dockerfile
```diff
@@ -112,6 +112,23 @@ ENV VPN_SERVICE_PROVIDER=pia \
...
     WIREGUARD_MTU= \
     WIREGUARD_IMPLEMENTATION=auto \
+    # Wireguard AmneziaWG userspace obfuscation (requires WIREGUARD_IMPLEMENTATION=amneziawg)
+    AMNEZIAWG_JC= \
+    AMNEZIAWG_JMIN= \
+    AMNEZIAWG_JMAX= \
+    AMNEZIAWG_S1= \
+    AMNEZIAWG_S2= \
+    AMNEZIAWG_S3= \
+    AMNEZIAWG_S4= \
```
qdm12:
```suggestion
    AMNEZIAWG_JC=0 \
    AMNEZIAWG_JMIN=0 \
    AMNEZIAWG_JMAX=0 \
    AMNEZIAWG_S1=0 \
    AMNEZIAWG_S2=0 \
    AMNEZIAWG_S3=0 \
    AMNEZIAWG_S4=0 \
```

qdm12:
I'm actually reworking the code, to have amneziawg as completely(almost) separate from wireguard; it seems amneziawg has a kernel module, so it feels kind of clunky to have it as a wireguard implementation; plus it's not really the same protocol, although it can behave as wireguard only, depending on its settings. I'll get back with a refactor today!

qdm12:
Refactored in b04529c3800cca6d3fcf7c981187e9d61d0d8998 - this breaks compatibility, but oh well just for 3 hours I doubt anyone has picked up using amneziawg in that short window (hopefully)!

Wiki updated ↔️ https://github.com/qdm12/gluetun-wiki/commit/476e96a6a6155832cddd6d80a66aadaa4e001482

DX37:
> Refactored in [b04529c](https://github.com/qdm12/gluetun/commit/b04529c3800cca6d3fcf7c981187e9d61d0d8998) - this breaks compatibility, but oh well just for 3 hours I doubt anyone has picked up using amneziawg in that short window (hopefully)!
> 
> Wiki updated ↔️ [qdm12/gluetun-wiki@476e96a](https://github.com/qdm12/gluetun-wiki/commit/476e96a6a6155832cddd6d80a66aadaa4e001482)

I am gettings errors like
`ERROR VPN settings: provider settings: server selection: Wireguard server selection settings: endpoint IP is not set`
Like docker-compose.yml wants me to describe WireGuard and Amnezia env vars simultaneously.

DX37:
This mixup is the closest config I can get right now
image
And it's still not working as intended
`ERROR [vpn] finding a VPN server: VPN type not supported for custom provider: amneziawg`

1a11:
@DX37 There's a discussion! So +1 #3239

PR #3164, opened 2026-02-26
feat(cryptostorm): add WireGuard and OpenVPN provider support

### LINKED ISSUE #3112: Feature request: cryptostorm.is VPN
### What's the feature 🧐

Hello,
Would it be possible to ensure compatibility with the excellent VPN provider: cryptostorm.is ?
Thank you.

_No response_

mpatton125:
## Description

- Add cryptostorm.is as a new VPN provider with both WireGuard and OpenVPN support
- Both protocols use port 443 (TCP/UDP), server data fetched from cryptostorm JSON API
- Includes port forwarding support via cryptostorm's internal forwarding server

## Issue

Fix #3112 

---

🤖 Generated with [Claude Code](https://claude.com/claude-code) <--- Full disclosure

### DIFF
--- a/internal/configuration/settings/portforward.go
+++ b/internal/configuration/settings/portforward.go
@@ -43,7 +43,8 @@ type PortForwarding struct {
 	// forwarded ports. The redirection is disabled if it is the slice [0],
 	// which is its default as well. If set and not [0], its length must match
 	// the PortsCount value, such that each forwarded port is redirected to
-	// the corresponding listening port.
+	// the corresponding listening port. For Cryptostorm, ListeningPorts[0]
+	// also specifies the port to request from the forwarding server (30000–65535).
 	ListeningPorts []uint16 `json:"listening_port"`
 	// PortsCount is the number of ports to forward. It is optional for ProtonVPN
 	// and be between 1 and 5. For other providers, it must be set to 1 if port
@@ -66,6 +67,7 @@ func (p PortForwarding) Validate(vpnProvider string) (err error) {
 		providerSelected = *p.Provider
 	}
 	validProviders := []string{
+		providers.Cryptostorm,
 		providers.Perfectprivacy,
 		providers.PrivateInternetAccess,
 		providers.Privatevpn,
@@ -143,14 +145,21 @@ func (p *PortForwarding) OverrideWith(other PortForwarding) {
 	p.Password = gosettings.OverrideWithComparable(p.Password, other.Password)
 }
 
-func (p *PortForwarding) setDefaults() {
+func (p *PortForwarding) setDefaults(vpnProvider string) {
 	p.Enabled = gosettings.DefaultPointer(p.Enabled, false)
 	p.Provider = gosettings.DefaultPointer(p.Provider, "")
 	p.Filepath = gosettings.DefaultPointer(p.Filepath, "/tmp/gluetun/forwarded_port")
 	p.UpCommand = gosettings.DefaultPointer(p.UpCommand, "")
 	p.DownCommand = gosettings.DefaultPointer(p.DownCommand, "")
 	p.ListeningPorts = gosettings.DefaultSlice(p.ListeningPorts, []uint16{0}) // disabled
-	p.PortsCount = gosettings.DefaultComparable(p.PortsCount, 1)
+	// For Cryptostorm, default PortsCount to the number of listening ports
+	// specified, since each listening port maps directly to a requested port.
+	defaultPortsCount := uint16(1)
+	if vpnProvider == providers.Cryptostorm &&
+		len(p.ListeningPorts) > 0 && p.ListeningPorts[0] != 0 {
+		defaultPortsCount = uint16(len(p.ListeningPorts))
+	}
+	p.PortsCount = gosettings.DefaultComparable(p.PortsCount, defaultPortsCount)
 }
 
 func (p PortForwarding) String() string {
--- a/internal/portforward/service/start.go
+++ b/internal/portforward/service/start.go
@@ -44,6 +44,10 @@ func (s *Service) Start(ctx context.Context) (runError <-chan error, err error) 
 		Username:       s.settings.Username,
 		Password:       s.settings.Password,
 		PortsCount:     s.settings.PortsCount,
+		// ListeningPorts is only used for Cryptostorm to request specific ports
+		// from the forwarding server; for other providers, it is only used to
+		// redirect forwarded ports to different local ports.
+		ListeningPorts: s.settings.ListeningPorts,
 	}
 	internalToExternalPorts, err := s.settings.PortForwarder.PortForward(ctx, obj)
 	if err != nil {
@@ -109,6 +113,9 @@ func (s *Service) onNewPorts(ctx context.Context, internalToExternalPorts map[ui
 		switch {
 		case userRedirectionEnabled: // precedence over auto redirection
 			destinationPort = s.settings.ListeningPorts[i]
+			if internalPort == destinationPort {
+				continue // no-op; avoid self-redirect in firewall
+			}
 		case port != internalPort: // auto redirection needed, source and destination ports differ
 			destinationPort = port
 		default:
--- a/internal/provider/cryptostorm/connection.go
+++ b/internal/provider/cryptostorm/connection.go
@@ -1,0 +1,15 @@
+package cryptostorm
+
+import (
+	"github.com/qdm12/gluetun/internal/configuration/settings"
+	"github.com/qdm12/gluetun/internal/models"
+	"github.com/qdm12/gluetun/internal/provider/utils"
+)
+
+func (p *Provider) GetConnection(selection settings.ServerSelection, ipv6Supported bool) (
+	connection models.Connection, err error,
+) {
+	defaults := utils.NewConnectionDefaults(443, 443, 443) //nolint:mnd
+	return utils.GetConnection(p.Name(),
+		p.storage, selection, defaults, ipv6Supported, p.connPicker)
+}
--- a/internal/provider/cryptostorm/openvpnconf.go
+++ b/internal/provider/cryptostorm/openvpnconf.go
@@ -1,0 +1,22 @@
+package cryptostorm
+
+import (
+	"github.com/qdm12/gluetun/internal/configuration/settings"
+	"github.com/qdm12/gluetun/internal/constants/openvpn"
+	"github.com/qdm12/gluetun/internal/models"
+	"github.com/qdm12/gluetun/internal/provider/utils"
+)
+
+func (p *Provider) OpenVPNConfig(connection models.Connection,
+	settings settings.OpenVPN, ipv6Supported bool,
+) (lines []string) {
+	providerSettings := utils.OpenVPNProviderSettings{
+		RemoteCertTLS:  true,
+		AuthUserPass:   true,
+		Ciphers:        []string{openvpn.AES256gcm},
+		VerifyX509Type: "name",
+		TLSCrypt:       "4875d729589689955012a2ee77f180ecb815c4a336c719c11241a058dafaae00806bbc21d5f1abad085341a3fca4b4f93949151c2979b4ee4390e8d9443acb0061d537f1e9157e45f542c3648f56330505f3eaff97ef82ee063b9d88bb9d5aa0060428455b51a2a4fd929d9af4b94adcb0a4acaa14ff62a9b0f4f9f0b3f01e71fc98a6c60e8584f4deb3de793a5a7bc27014c9369f9724bc810ef0d191b3020478eead725b3ae6aaef2e1030a197e417421f159ed54eb2629afcfb337cf9a0025bf1d5c0d820fffb219d0b4214043d2df27ed367b522945a5dadc748e2ca379e3971789dbdf609b3d9bfe866361b28e3c90589baa925157ad833093a5a7bede5", //nolint:lll
+		CAs:            []string{"MIICCzCCAW2gAwIBAgIUMRTTJ6nuPjmSxaRfbw5f+dZ9d/gwCgYIKoZIzj0EAwQwGTEXMBUGA1UEAwwOY3J5cHRvc3Rvcm0gQ0EwHhcNMTgwOTE3MjAwODU4WhcNMzgwOTE3MjAwODU4WjAZMRcwFQYDVQQDDA5jcnlwdG9zdG9ybSBDQTCBmzAQBgcqhkjOPQIBBgUrgQQAIwOBhgAEARKu20PBrr226TP6mQQGtzCqQqBKfGaA05Ml5nrGSV6wzBQDQga4/cPepGrE/tpzRX72KSfZD6nJfQLYen7kdc3PAEvWFBhCovq7e4L6xJ5qV5aMf89QjNhJ/xn//dlxE8Z6UfIx63dJX9q3EHNxateU84lDkbCrqckkckcZF4C1a9Ooo1AwTjAdBgNVHQ4EFgQUdaVDaoi48Yf2RugXqJ4yJ4Z4utgwHwYDVR0jBBgwFoAUdaVDaoi48Yf2RugXqJ4yJ4Z4utgwDAYDVR0TBAUwAwEB/zAKBggqhkjOPQQDBAOBiwAwgYcCQVcCw/8OVpNqltDYczqHmX4sMRsZTY0iIzl1rYY/0/ZPIvzjlMFnouHwb8asJZRMBNECq7u9PCbG3jdu6lYtcCm+AkIB3IYYKuXLKW7ucdttNODBqH2Rail+9oBWTV2ZFKVVwELlKadHx9UvAcpAaV1alkN80CgI2tad2/qVdpSIQpfVvTI="}, //nolint:lll
+	}
+	return utils.OpenVPNConfig(providerSettings, connection, settings, ipv6Supported)
+}
--- a/internal/provider/cryptostorm/portforward.go
+++ b/internal/provider/cryptostorm/portforward.go
@@ -14,7 +14,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/qdm12/gluetun/internal/provider/common"
 	"github.com/qdm12/gluetun/internal/provider/utils"
 )
 
@@ -81,8 +80,7 @@ func (p *Provider) PortForward(ctx context.Context, objects utils.PortForwardObj
 	}
 
 	if listeningPort == 0 {
-		return nil, fmt.Errorf("%w: set VPN_PORT_FORWARDING_LISTENING_PORTS to a value between 30000 and 65535",
-			common.ErrPortForwardNotSupported)
+		return nil, fmt.Errorf("port forwarding not supported: set VPN_PORT_FORWARDING_LISTENING_PORTS to a value between 30000 and 65535")
 	}
 
 	postBody := "port=" + strconv.FormatUint(uint64(listeningPort), 10)
@@ -108,8 +106,7 @@ func (p *Provider) PortForward(ctx context.Context, objects utils.PortForwardObj
 	}
 
 	if !requestedFound {
-		return nil, fmt.Errorf("%w: requested port %d not found in server response",
-			common.ErrPortForwardNotSupported, listeningPort)
+		return nil, fmt.Errorf("port forwarding not supported: requested port %d not found in server response", listeningPort)
 	}
 
 	p.forwardedPort = listeningPort
@@ -142,7 +139,7 @@ func (p *Provider) registerPort(ctx context.Context, client *http.Client,
 	defer response.Body.Close()
 
 	if response.StatusCode != http.StatusOK {
-		return false, fmt.Errorf("%w: %d %s", common.ErrHTTPStatusCodeNotOK,
+		return false, fmt.Errorf("HTTP status code not OK: %d %s",
 			response.StatusCode, response.Status)
 	}
 
@@ -210,10 +207,6 @@ func (p *Provider) KeepPortForward(ctx context.Context,
 	return ctx.Err()
 }
 
-func (p *Provider) deletePort(ctx context.Context, client *http.Client, port uint16) error {
-	return p.deletePortFromURL(ctx, client, portForwardURLs[0], port)
-}
-
 func (p *Provider) deletePortFromURL(ctx context.Context, client *http.Client, url string, port uint16) error {
 	body := strings.NewReader("delfwd=" + strconv.FormatUint(uint64(port), 10))
 	request, err := http.NewRequestWithContext(ctx, http.MethodPost, url, body)
@@ -227,7 +220,7 @@ func (p *Provider) deletePortFromURL(ctx context.Context, client *http.Client, u
 	}
 	_ = response.Body.Close()
 	if response.StatusCode != http.StatusOK {
-		return fmt.Errorf("%w: %d %s", common.ErrHTTPStatusCodeNotOK,
+		return fmt.Errorf("HTTP status code not OK: %d %s",
 			response.StatusCode, response.Status)
 	}
 	return nil
--- a/internal/provider/cryptostorm/portforward.go
+++ b/internal/provider/cryptostorm/portforward.go
@@ -156,8 +156,7 @@ func (p *Provider) registerPort(ctx context.Context, client *http.Client,
 		matches = regexForwardHTML.FindAllStringSubmatch(bodyStr, -1)
 	}
 	if len(matches) == 0 {
-		return false, fmt.Errorf("%w: no active port forwards found in response",
-			common.ErrPortForwardNotSupported)
+		return false, fmt.Errorf("port forwarding not supported: no active port forwards found in response")
 	}
 
 	// The server response lists all currently active forwards for this session,
--- a/internal/provider/cryptostorm/portforward.go
+++ b/internal/provider/cryptostorm/portforward.go
@@ -1,0 +1,272 @@
+package cryptostorm
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/fs"
+	"net/http"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/qdm12/gluetun/internal/provider/common"
+	"github.com/qdm12/gluetun/internal/provider/utils"
+)
+
+// portRangePattern matches the valid Cryptostorm port range 30000-65535.
+const portRangePattern = `([3-5]\d{4}|6[0-4]\d{3}|65[0-4]\d{2}|655[0-2]\d|6553[0-5])`
+
+// ipPattern matches either an IPv4 address or a bracketed IPv6 address.
+const ipPattern = `(?:\d+\.\d+\.\d+\.\d+|\[[0-9a-fA-F:]+\])`
+
+// regexForwardPlainText matches plain text responses (e.g. from curl):
+//
+//	37.120.234.253:55555 -> 10.10.123.139:55555
+//	[2400:a842:c46e:1::4a]:55555 -> [fd00:10:10::e0e0]:55555
+var regexForwardPlainText = regexp.MustCompile(
+	ipPattern + `:` + portRangePattern + `\s*->\s*` + ipPattern + `:\d+`)
+
+// regexForwardHTML matches the HTML response from the port forwarding page.
+// Each forwarded port has a hidden delete input:
+//
+//	<input type="hidden" name="delfwd" value="30000">
+var regexForwardHTML = regexp.MustCompile(
+	`name="delfwd"\s+value="` + portRangePattern + `"`)
+
+// portForwardURLs lists all port forwarding endpoints.
+// IPv4 and IPv6 must both be registered for dual-stack support.
+var portForwardURLs = []string{
+	"http://10.31.33.7/fwd",
+	"http://[2001:db8::7]/fwd",
+}
+
+// portForwardData is the data persisted to the port forward JSON file.
+type portForwardData struct {
+	Ports []uint16 `json:"ports"`
+}
+
+// PortForward registers a forwarded port with the Cryptostorm port forwarding server
+// and returns the active forwarded ports. The server returns plain text listing
+// current forwardings. We POST the desired port and parse the response.
+// Valid port range is 30000-65535.
+// See: https://cryptostorm.is/portfwd
+func (p *Provider) PortForward(ctx context.Context, objects utils.PortForwardObjects) (
+	internalToExternalPorts map[uint16]uint16, err error,
+) {
+	const timeout = 10 * time.Second
+	ctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+
+	// Determine the port to request:
+	// 1. Use VPN_PORT_FORWARDING_LISTENING_PORTS[0] if set (non-zero).
+	// 2. Otherwise try to read a previously persisted port.
+	// 3. Otherwise return an error (Cryptostorm does not auto-assign ports).
+	var listeningPort uint16
+	if len(objects.ListeningPorts) > 0 && objects.ListeningPorts[0] != 0 {
+		listeningPort = objects.ListeningPorts[0]
+	}
+	if listeningPort == 0 {
+		data, err := readPortForwardData(p.portForwardPath)
+		if err != nil {
+			return nil, fmt.Errorf("reading persisted port forward data: %w", err)
+		}
+		if len(data.Ports) > 0 {
+			listeningPort = data.Ports[0]
+		}
+	}
+
+	if listeningPort == 0 {
+		return nil, fmt.Errorf("%w: set VPN_PORT_FORWARDING_LISTENING_PORTS to a value between 30000 and 65535",
+			common.ErrPortForwardNotSupported)
+	}
+
+	postBody := "port=" + strconv.FormatUint(uint64(listeningPort), 10)
+
+	// Register the port with both IPv4 and IPv6 endpoints.
+	// The IPv6 endpoint may not be reachable if the tunnel has no IPv6,
+	// so we log a warning rather than failing hard.
+	requestedFound := false
+	for _, url := range portForwardURLs {
+		found, err := p.registerPort(ctx, objects.Client, url, postBody, listeningPort, objects)
+		if err != nil {
+			if url == portForwardURLs[0] {
+				// IPv4 failure is fatal.
+				return nil, err
+			}
+			// IPv6 failure is a warning only.
+			objects.Logger.Warn("IPv6 port forward registration failed (" + url + "): " + err.Error())
+			continue
+		}
+		if found {
+			requestedFound = true
+		}
+	}
+
+	if !requestedFound {
+		return nil, fmt.Errorf("%w: requested port %d not found in server response",
+			common.ErrPortForwardNotSupported, listeningPort)
+	}
+
+	p.forwardedPort = listeningPort
+	internalToExternalPorts = map[uint16]uint16{listeningPort: listeningPort}
+
+	// Persist so the next restart can reuse this port without re-requesting.
+	if err := writePortForwardData(p.portForwardPath, portForwardData{Ports: []uint16{listeningPort}}); err != nil {
+		return nil, fmt.Errorf("persisting port forward data: %w", err)
+	}
+
+	return internalToExternalPorts, nil
+}
+
+// registerPort POSTs the port to a single forwarding endpoint and returns
+// whether the requested port was confirmed in the response.
+func (p *Provider) registerPort(ctx context.Context, client *http.Client,
+	url, postBody string, listeningPort uint16, objects utils.PortForwardObjects,
+) (requestedFound bool, err error) {
+	request, err := http.NewRequestWithContext(ctx, http.MethodPost, url,
+		strings.NewReader(postBody))
+	if err != nil {
+		return false, fmt.Errorf("creating HTTP request: %w", err)
+	}
+	request.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+
+	response, err := client.Do(request)
+	if err != nil {
+		return false, fmt.Errorf("sending HTTP request: %w", err)
+	}
+	defer response.Body.Close()
+
+	if response.StatusCode != http.StatusOK {
+		return false, fmt.Errorf("%w: %d %s", common.ErrHTTPStatusCodeNotOK,
+			response.StatusCode, response.Status)
+	}
+
+	body, err := io.ReadAll(response.Body)
+	if err != nil {
+		return false, fmt.Errorf("reading response body: %w", err)
+	}
+
+	// Parse forwarded ports from the response. The server returns HTML to
+	// Go's HTTP client but plain text to curl, so we try both formats.
+	bodyStr := string(body)
+	matches := regexForwardPlainText.FindAllStringSubmatch(bodyStr, -1)
+	if len(matches) == 0 {
+		matches = regexForwardHTML.FindAllStringSubmatch(bodyStr, -1)
+	}
+	if len(matches) == 0 {
+		return false, fmt.Errorf("%w: no active port forwards found in response",
+			common.ErrPortForwardNotSupported)
+	}
+
+	// The server response lists all currently active forwards for this session,
+	// which may include stale ports from prior runs. Delete any that are not
+	// the one we requested, then verify ours is present.
+	const base, bitSize = 10, 16
+	for _, match := range matches {
+		portUint64, err := strconv.ParseUint(match[1], base, bitSize)
+		if err != nil {
+			return false, fmt.Errorf("parsing port number %q: %w", match[1], err)
+		}
+		port := uint16(portUint64)
+		if port == listeningPort {
+			requestedFound = true
+			continue
+		}
+		// Best-effort delete; log but don't fail if it errors.
+		if err := p.deletePortFromURL(ctx, client, url, port); err != nil {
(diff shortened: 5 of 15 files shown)


internal/provider/cryptostorm/updater/api.go
```diff
@@ -0,0 +1,55 @@
...
+	"fmt"
+	"net/http"
+)
+
+var (
+	ErrHTTPStatusCodeNotOK = errors.New("HTTP status code not OK")
+	ErrDecodeResponseBody  = errors.New("failed decoding response body")
+)
+
+// nodeData represents one entry in the Cryptostorm node list JSON.
```
qdm12: nit unneeded comment

```suggestion
```

internal/provider/cryptostorm/updater/api.go
```diff
@@ -0,0 +1,55 @@
...
+	WgPubKey string `json:"wg_pubkey"`
+	PortFwd  bool   `json:"port_forward"`
+}
+
+// fetchAPI retrieves the Cryptostorm node list.
+// Cryptostorm does not publish a formal JSON API; this function fetches
+// their publicly available node list. If the upstream format changes,
+// update the nodeData struct and parsing logic accordingly.
+func fetchAPI(ctx context.Context, client *http.Client) (data []nodeData, err error) {
+	const url = "https://cryptostorm.is/wireguard/nodes.json"
```
qdm12: there is no json API, this url is the same as `https://cryptostorm.is/wireguard`. This returns html, not json and this will fail

internal/provider/cryptostorm/connection.go
```diff
@@ -0,0 +1,15 @@
...
+import (
+	"github.com/qdm12/gluetun/internal/configuration/settings"
+	"github.com/qdm12/gluetun/internal/models"
+	"github.com/qdm12/gluetun/internal/provider/utils"
+)
+
+func (p *Provider) GetConnection(selection settings.ServerSelection, ipv6Supported bool) (
+	connection models.Connection, err error,
+) {
+	defaults := utils.NewConnectionDefaults(443, 443, 443) //nolint:mnd
```
qdm12, line 12: please double check openvpn udp tcp and wireguard all work with those 443 ports.

internal/provider/cryptostorm/openvpnconf.go
```diff
@@ -0,0 +1,24 @@
...
+		CAs:            []string{"MIICCzCCAW2gAwIBAgIUMRTTJ6nuPjmSxaRfbw5f+dZ9d/gwCgYIKoZIzj0EAwQwGTEXMBUGA1UEAwwOY3J5cHRvc3Rvcm0gQ0EwHhcNMTgwOTE3MjAwODU4WhcNMzgwOTE3MjAwODU4WjAZMRcwFQYDVQQDDA5jcnlwdG9zdG9ybSBDQTCBmzAQBgcqhkjOPQIBBgUrgQQAIwOBhgAEARKu20PBrr226TP6mQQGtzCqQqBKfGaA05Ml5nrGSV6wzBQDQga4/cPepGrE/tpzRX72KSfZD6nJfQLYen7kdc3PAEvWFBhCovq7e4L6xJ5qV5aMf89QjNhJ/xn//dlxE8Z6UfIx63dJX9q3EHNxateU84lDkbCrqckkckcZF4C1a9Ooo1AwTjAdBgNVHQ4EFgQUdaVDaoi48Yf2RugXqJ4yJ4Z4utgwHwYDVR0jBBgwFoAUdaVDaoi48Yf2RugXqJ4yJ4Z4utgwDAYDVR0TBAUwAwEB/zAKBggqhkjOPQQDBAOBiwAwgYcCQVcCw/8OVpNqltDYczqHmX4sMRsZTY0iIzl1rYY/0/ZPIvzjlMFnouHwb8asJZRMBNECq7u9PCbG3jdu6lYtcCm+AkIB3IYYKuXLKW7ucdttNODBqH2Rail+9oBWTV2ZFKVVwELlKadHx9UvAcpAaV1alkN80CgI2tad2/qVdpSIQpfVvTI="}, //nolint:lll
+		ExtraLines:     []string{"tls-version-min 1.2"},
```
qdm12: let's remove, this is unneeded for openvpn 2.6 and openvpn 2.5 is just limping around and will be removed shortly from gluetun anyway


```suggestion
```

internal/storage/servers.json
```diff
@@ -14753,4636 +13993,5710 @@
...
+        "hostname": "us4.vpn.airdns.org",
+        "tcp": true,
+        "udp": true,
+        "ips": [
+          "198.44.134.10"
+        ]
+      }
+    ]
+  },
+  "cryptostorm": {
```
qdm12: that was not obtained from the code in the updater, so it most likely pure hallucination; fix the updater and revert these changes.

Dockerfile
```diff
@@ -4,7 +4,7 @@ ARG GO_VERSION=1.25
 ARG XCPUTRANSLATE_VERSION=v0.9.0
 ARG GOLANGCI_LINT_VERSION=v2.4.0
 ARG MOCKGEN_VERSION=v1.6.0
-ARG BUILDPLATFORM=linux/amd64
+ARG BUILDPLATFORM
```
qdm12: revert this. this is so it works on older docker versions, and does not impact newer docker versions. no reason to change this, and it would be out of scope anyway... ugh artificial "intelligence"

internal/provider/cryptostorm/portforward.go
```diff
@@ -0,0 +1,94 @@
...
+	const timeout = 10 * time.Second
+	ctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+
+	postBody := ""
+	if objects.ListeningPort != 0 {
+		postBody = "port=" + strconv.FormatUint(uint64(objects.ListeningPort), 10)
+	}
+
+	pfURL := "http://" + cryptostormPFServer + "/fwd"
```
qdm12: change this to


```suggestion
	const portForwardURL = "http://10.31.33.7/fwd"
```

and remove the constant `cryptostormPFServer`. Always prefer local constants, unless the constant is used in different functions.

internal/provider/cryptostorm/portforward.go
```diff
@@ -0,0 +1,94 @@
...
+	"time"
+
+	"github.com/qdm12/gluetun/internal/provider/common"
+	"github.com/qdm12/gluetun/internal/provider/utils"
+)
+
+// regexActivePort matches forwarded ports listed in the Cryptostorm HTML response,
+// e.g. <p class="list-group-item-header">Port 55555</p>
+// Valid port range per Cryptostorm is 30000-65535.
+var regexActivePort = regexp.MustCompile(`list-group-item-header">Port ([0-9]+)<`)
```
qdm12: If the valid port range is 30000-65535, adapt the regex to match that range instead of just `[0-9]+`

internal/provider/cryptostorm/portforward.go
```diff
@@ -0,0 +1,94 @@
...
+// port forwarding server, reachable only from within the VPN tunnel.
+const cryptostormPFServer = "10.31.33.7"
+
+// PortForward registers a forwarded port with the Cryptostorm port forwarding server
+// and returns the active forwarded ports. The server always returns an HTML page;
+// we POST the desired port and then parse the current forwards list from the response.
+// If the port is already forwarded (e.g. from a previous session) it will appear in
+// the list regardless of whether the POST succeeded, so we treat that as success.
+// Valid port range is 30000–65535.
+// See: https://cryptostorm.is/port_forwarding
```
qdm12: 404 not found, possible hallucination? please fix

it's https://cryptostorm.is/portfwd

internal/provider/cryptostorm/portforward.go
qdm12: add a comment for IPv6 it's http://[2001:db8::7]/fwd - that will be used later.

qdm12: @mpatton125 can you check what `http://10.31.33.7/fwd` returns when you are connected to the VPN? To confirm what the AI wrote is correct

internal/provider/cryptostorm/portforward.go
```diff
@@ -0,0 +1,94 @@
...
+	// Parse all currently active port forwards from the HTML response.
+	matches := regexActivePort.FindAllStringSubmatch(string(body), -1)
+	if len(matches) == 0 {
+		return nil, fmt.Errorf("%w: no active port forwards found in response",
+			common.ErrPortForwardNotSupported)
+	}
+
+	const base, bitSize = 10, 16
+	for _, match := range matches {
+		portUint64, err := strconv.ParseUint(match[1], base, bitSize)
```
qdm12, line 167: can there be multiple ports? 🤔

internal/provider/utils/portforward.go
```diff
@@ -25,6 +25,9 @@ type PortForwardObjects struct {
 	Username string
 	// Password is used by Private Internet Access for port forwarding.
 	Password string
+	// ListeningPort is the port to request from the provider, used by Cryptostorm.
+	// A value of 0 means let the provider assign a port.
+	ListeningPort uint16
```
qdm12: @mpatton125 please confirm if `http://10.31.33.7/fwd` tells you you can specify a `port` field, and if that port field can be set to `0` to let cryptostorm assign you a random port.

If this is the case:

1. use the `VPN_PORT_FORWARDING_LISTENING_PORT` value to set this. If this one is not set, try to read it from `/gluetun/portforward/cryptostorm.json`. If this file does not exist, leave the ListeningPort field to 0 to let cryptostorm decide.
2. in all cases, persist the resulting port(s) to `/gluetun/portforward/cryptostorm.json`

And also:
- add a comment in internal/provider/privateinternetaccess/provider.go next to `const jsonPortForwardPath = "/gluetun/piaportforward.json"` saying `TODO v4: move to /gluetun/portforward/privateinternetaccess.json`
- update the comment for the settings field `VPN_PORT_FORWARDING_LISTENING_PORT` and make sure it runs a no-op in case the forwarded port is already the value specified in `VPN_PORT_FORWARDING_LISTENING_PORT`

qdm12 (changes requested):
- Big hallucination on the updating servers code. this needs to be fixed somehow, or this PR won't be accepted
- Port forwarding: I need a human to check the information returned from the port forwarding endpoint when connected
- When above is done, I need a human to verify openvpn tcp, udp, wireguard and port forwarding all work fine.

internal/provider/cryptostorm/updater/api.go
mpatton125: Now scraping endpoint data from https://cryptostorm.is/wireguard.

internal/provider/cryptostorm/connection.go
mpatton125: Definitely both work with 443 - provider says they accept OpenVPN and Wireguard connections on any port (1-65535).

internal/provider/cryptostorm/openvpnconf.go
mpatton125: Done.

internal/provider/cryptostorm/portforward.go
mpatton125: Yes this provider allows multiple port forwards, per endpoint.

internal/provider/utils/portforward.go
mpatton125: Yes you can specify a port at that URL, from 30000-65535.  No, it can't be 0 for a random port - user must specify.

internal/provider/cryptostorm/portforward.go
qdm12: Is there a maximum limit of ports forwarded??

Let's wait for #3208 to be merged, then this can use the new setting field `PortsCount`

internal/provider/utils/portforward.go
qdm12: Got it, let's wait for #3208 

note for future:

we'll then use `VPN_PORT_FORWARDING_LISTENING_PORTS=8000,8001` to specify ports you want. We should then change `func (p *PortForwarding) setDefaults()` to `func (p *PortForwarding) setDefaults(vpnProvider string)` and set, if the vpnProvider is cryptostorm only, `p.PortsCount = gosettings.DefaultComparable(p.PortsCount, len(p.ListeningPorts))`

internal/configuration/settings/portforward.go
```diff
@@ -39,7 +39,8 @@ type PortForwarding struct {
 	DownCommand *string `json:"down_command"`
 	// ListeningPort is the port traffic would be redirected to from the
 	// forwarded port. The redirection is disabled if it is set to 0, which
-	// is its default as well.
+	// is its default as well. For Cryptostorm, this also specifies the
+	// port to request for forwarding (must be between 30000 and 65535).
```
qdm12: (no action for now) to be updated after #3208 is merged and this branch gets updated with master.

internal/portforward/service/start.go
```diff
@@ -42,18 +41,45 @@ func (s *Service) Start(ctx context.Context) (runError <-chan error, err error)
 		CanPortForward: s.settings.CanPortForward,
 		Username:       s.settings.Username,
 		Password:       s.settings.Password,
+		ListeningPort:  s.settings.ListeningPort,
```
qdm12: (no action for now) to be updated after #3208 is merged - add comment stating this is only used for cryptostorm to request specific ports, for other providers, it's only used to redirect ports

internal/portforward/service/start.go
```diff
@@ -42,18 +41,45 @@ func (s *Service) Start(ctx context.Context) (runError <-chan error, err error)
...
+	s.logger.Info(portsToString(ports))
+
+	for _, port := range ports {
+		err = s.portAllower.SetAllowedPort(ctx, port, s.settings.Interface)
+		if err != nil {
+			return nil, fmt.Errorf("allowing port in firewall: %w", err)
+		}
 
-	err = s.onNewPorts(ctx, ports)
+		if s.settings.ListeningPort != 0 && port != s.settings.ListeningPort {
```
qdm12: (no action for now) to be updated after #3208 is merged - update to check by index `ports[i] != s.settings.ListeningPorts[i]` and also we should ensure both slices are sorted

internal/provider/cryptostorm/updater/api.go
```diff
@@ -0,0 +1,106 @@
...
+	"golang.org/x/net/html"
+)
+
+var ErrHTMLTableNotFound = errors.New("HTML server table not found")
+
+// nodeData represents one entry parsed from the Cryptostorm wireguard page.
+type nodeData struct {
+	Location string // e.g. "Canada - Montreal", "Austria", "US - Texas - Dallas"
+	Hostname string // e.g. "austria.cstorm.is"
+	WgPubKey string // WireGuard public key
```
qdm12: nit unneeded

```suggestion
	WgPubKey string
```

internal/provider/cryptostorm/updater/api.go
```diff
@@ -0,0 +1,106 @@
...
+func fetchNodes(ctx context.Context, client *http.Client) (
+	nodes []nodeData, warnings []string, err error) {
+	const url = "https://cryptostorm.is/wireguard"
+
+	rootNode, err := htmlutils.Fetch(ctx, client, url)
+	if err != nil {
+		return nil, nil, fmt.Errorf("fetching HTML: %w", err)
+	}
+
+	return parseHTML(rootNode)
```
qdm12:
```suggestion
	nodes, warnings, err = parseHTML(rootNode)
	if err != nil {
		return nil, nil, fmt.Errorf("parsing HTML: %w", err)
	}
	return nodes, warnings, nil
```

internal/provider/cryptostorm/updater/api.go
```diff
@@ -0,0 +1,106 @@
...
+	if tbody != nil {
+		rowParent = tbody
+	}
+	rows := htmlutils.DirectChildren(rowParent, htmlutils.MatchData("tr"))
+	for i, row := range rows {
+		if i == 0 {
+			// Skip header row.
+			continue
+		}
+
```
qdm12: nit

```suggestion
	for i := 1; i++; i < len(rows) { // skipping the header row
		row := rows[i]
```

internal/portforward/service/start.go
```diff
@@ -44,6 +44,7 @@ func (s *Service) Start(ctx context.Context) (runError <-chan error, err error)
 		Username:       s.settings.Username,
 		Password:       s.settings.Password,
 		PortsCount:     s.settings.PortsCount,
+		ListeningPorts: s.settings.ListeningPorts,
```
qdm12, line 50: this might be problematic in that it would be used to redirect the port to itself as well in the firewall code?

internal/provider/cryptostorm/updater/api.go
```diff
@@ -0,0 +1,106 @@
...
+	"fmt"
+	"net/http"
+	"strings"
+
+	htmlutils "github.com/qdm12/gluetun/internal/updater/html"
+	"golang.org/x/net/html"
+)
+
+var ErrHTMLTableNotFound = errors.New("HTML server table not found")
+
```
qdm12: my bad - can you rebase / merge the upstream master branch, and get rid of those sentinel errors. Unless you use `errors.Is` to check them, that was a silly linter decision from my part, so I did remove all sentinel errors except a few select that were actually useful. You can now just inline where it's being returned and not define it at global scope anymore

internal/provider/cryptostorm/updater/api.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	// html.Parse inserts <tbody> per the HTML5 spec, so <tr> elements
+	// are not direct children of <table>.
+	tbody := htmlutils.DirectChild(tableNode, htmlutils.MatchData("tbody"))
+	rowParent := tableNode
+	if tbody != nil {
+		rowParent = tbody
+	}
+	rows := htmlutils.DirectChildren(rowParent, htmlutils.MatchData("tr"))
```
qdm12, line 52:
```suggestion
	rows := htmlutils.DirectChildren(rowParent, htmlutils.MatchData("tr"))
	nodes = make([]nodeData, 0, len(rows)-1)
```

internal/provider/cryptostorm/updater/api.go
```diff
@@ -0,0 +1,106 @@
...
+		location = strings.TrimSpace(location)
+
+		hostname := strings.TrimSpace(textContent(cells[1]))
+		wgPubKey := strings.TrimSpace(textContent(cells[2]))
+
+		if hostname == "" {
+			warnings = append(warnings,
+				htmlutils.WrapWarning("empty hostname", row))
+			continue
+		}
```
qdm12:
```suggestion
		switch {
		case location == "":
			warnings = append(warnings,
				htmlutils.WrapWarning("empty location", row))
			continue
		case hostname == "":
			warnings = append(warnings,
				htmlutils.WrapWarning("empty hostname", row))
			continue
		}
```

internal/provider/cryptostorm/updater/hosttoserver.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"github.com/qdm12/gluetun/internal/models"
+)
+
+// hostToServer maps composite keys (hostname/protocol) to servers.
+// The hostname portion is used for DNS resolution.
+type hostToServer map[string]models.Server
+
+func (hts hostToServer) toUniqueHostsSlice() (hosts []string) {
+	seen := make(map[string]struct{})
```
qdm12:
```suggestion
	seen := make(map[string]struct{}, len(hts))
	hosts = make([]string, 0, len(seen))
```

internal/provider/cryptostorm/updater/hosttoserver.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+	return servers
+}
+
+// extractHost returns the hostname from a composite key like "host.example.com/wg".
+func extractHost(key string) string {
+	if idx := strings.LastIndex(key, "/"); idx >= 0 {
+		return key[:idx]
+	}
+	return key
```
qdm12, line 53: always assume the composite key is well formed, otherwise it's fine to panic / have empty host which would error somewhere else.


```suggestion
	return key[:strings.LastIndex(key, "/")]
```

internal/provider/cryptostorm/updater/servers.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	if err != nil {
+		return nil, fmt.Errorf("fetching nodes: %w", err)
+	}
+
+	hts := make(hostToServer)
+
+	for _, node := range nodes {
+		country, city := parseLocation(node.Location)
+
```
qdm12, line 28:
```suggestion
		if country == "" {
			return nil, fmt.Errorf("country is empty for node %v", node)
		}

```

internal/provider/cryptostorm/updater/servers.go
```diff
@@ -0,0 +1,119 @@
...
+		if node.WgPubKey != "" {
+			server := models.Server{
+				VPN:         vpn.Wireguard,
+				Country:     country,
+				City:        city,
+				Hostname:    node.Hostname,
+				WgPubKey:    node.WgPubKey,
+				PortForward: true,
+			}
+			hts[node.Hostname+"/wg"] = server
```
qdm12, line 39: the `/wg` does not produce a valid hostname, so change `hts` to be a host to serverS aka `map[string][]models.Server`, where each hostname has one or two servers (openvpn+wireguard). That way no need for `extractHost` anymore as well.

internal/provider/cryptostorm/updater/servers.go
```diff
@@ -0,0 +1,119 @@
...
+//   - "Sydney - Australia" (city - country, detected by known country names)
+func parseLocation(location string) (country, city string) {
+	parts := strings.Split(location, " - ")
+	switch len(parts) {
+	case 1:
+		return parts[0], ""
+	case 2:
+		// Check if the second part is a known country name, indicating
+		// a reversed "City - Country" format (e.g. "Sydney - Australia").
+		if isCountryName(parts[1]) && !isCountryName(parts[0]) {
```
qdm12, line 108: that check is a bit off given isCountryName is restricted to just reversed country names australia and japan, i think we can just have it as


```suggestion
		if isCountryName(parts[1]) {
```

unless you want to list all country names, but that's kind of risky and may break in the future.

internal/provider/cryptostorm/connection.go
```diff
@@ -0,0 +1,88 @@
...
+func (p *Provider) GetConnection(selection settings.ServerSelection, ipv6Supported bool) (
+	connection models.Connection, err error,
+) {
+	servers, err := p.storage.FilterServers(p.Name(), selection)
+	if err != nil {
+		return connection, fmt.Errorf("filtering servers: %w", err)
+	}
+
+	// Pick a random server from the filtered list.
+	server := servers[rand.New(p.randSource).Intn(len(servers))] //nolint:gosec
```
qdm12: rebase on upstream master, servers are now being cycled through

internal/provider/cryptostorm/connection.go
```diff
@@ -0,0 +1,88 @@
...
+		network = "ip"
+	}
+	ips, err := net.DefaultResolver.LookupNetIP(ctx, network, server.Hostname)
+	if err != nil {
+		return connection, fmt.Errorf("resolving %s: %w", server.Hostname, err)
+	}
+	if len(ips) == 0 {
+		return connection, fmt.Errorf("no IP addresses found for %s", server.Hostname)
+	}
+	ip := ips[rand.New(p.randSource).Intn(len(ips))] //nolint:gosec
```
qdm12: not happening, you need to use ips from the servers data, no network traffic before the vpn is up, and that includes dns traffic

internal/provider/cryptostorm/connection.go
```diff
@@ -0,0 +1,88 @@
...
+
+	// Determine protocol.
```
qdm12: unneeded



```suggestion
```

internal/provider/cryptostorm/connection.go
```diff
@@ -0,0 +1,88 @@
...
+	protocol := constants.UDP
+	if selection.VPN == vpn.OpenVPN && selection.OpenVPN.Protocol == constants.TCP {
+		protocol = constants.TCP
+	}
```
qdm12:
```suggestion
	if selection.VPN == vpn.OpenVPN {
		protocol = selection.OpenVPN.Protocol
```

internal/provider/cryptostorm/connection.go
```diff
@@ -0,0 +1,88 @@
...
+	default: // OpenVPN
+		if custom := *selection.OpenVPN.CustomPort; custom > 0 {
+			port = custom
+		}
+	}
+
+	// Allow explicit endpoint IP override.
+	switch selection.VPN {
+	case vpn.OpenVPN:
+		if t := selection.OpenVPN.EndpointIP; t.IsValid() && !t.IsUnspecified() {
```
qdm12: why `t`??? change to `ip`. same below

internal/provider/cryptostorm/portforward.go
qdm12: rebase on upstream master, which supports multiple ports. What's the number of ports limit?

internal/provider/cryptostorm/provider.go
```diff
@@ -0,0 +1,35 @@
...
+
+type Provider struct {
+	storage         common.Storage
+	randSource      rand.Source
+	portForwardPath string
+	forwardedPort   uint16 // set after a successful PortForward, used for teardown
+	common.Fetcher
+}
+
+func New(storage common.Storage, randSource rand.Source,
```
qdm12: rebase on upstream master, randSource is no longer needed.

internal/provider/cryptostorm/provider.go
```diff
@@ -0,0 +1,35 @@
...
+	client *http.Client, updaterWarner common.Warner,
+	parallelResolver common.ParallelResolver,
+) *Provider {
+	const jsonPortForwardPath = "/gluetun/portforward/cryptostorm.json"
```
qdm12, line 21: nit rename to `portForwardPath`

internal/provider/utils/portforward.go
```diff
@@ -27,6 +27,10 @@ type PortForwardObjects struct {
 	Password string
 	// PortsCount is used by ProtonVPN for port forwarding.
 	PortsCount uint16
+	// ListeningPorts are the user-specified ports. For Cryptostorm, ListeningPorts[0]
+	// is the port to request from the forwarding server (must be 30000–65535).
+	// A nil or [0] value means fall back to the persisted port.
```
qdm12, line 32: can there be multiple ports requested?

amarevite: From some preliminary testing, I don't think you can request multiple ports in one POST request, but you can have multiple ports open if you request them one by one. You can close them all at once though:
```bash
➜ curl --request POST --data 'port=80808,80809' http://10.31.33.7/fwd
INVALID_PARAM_RECEIVED
Add a port:
wget -qO- http://10.31.33.7/fwd --post-data="port=31337"
curl -sS http://10.31.33.7/fwd -d port=31337
Delete one port:
wget -qO- http://10.31.33.7/fwd --post-data="delfwd=31337"
curl -sS http://10.31.33.7/fwd -d delfwd=31337
Delete all ports:
wget -qO- http://10.31.33.7/fwd --post-data="delallfwd=1"
curl -sS http://10.31.33.7/fwd -d delallfwd=1
```
Leaving the parameter empty, setting it to 0, or trying `ports` instead of `port` are all invalid. It seems you have to explicitly request one specific port at a time.

I'd like to see this merged, so let me know if there's anything else you'd like me to test.

amarevite:
What work remains before this can be merged? Is it's just linting and git tree management or are code changes needed before it can be merged?

mpatton125:
I'm not sure what needs to be done for it to be merged. It works - I've been using it for months.

qdm12:
I'll review this post v3.42.0 release which I'm trying to get out soon - I'm still playing catchup with PRs these days, but I'll get to it !
You can use code from the fork/branch in the meantime.

Edit : I'll try to review it tomorrow fingers crossed !

internal/provider/cryptostorm/connection.go
qdm12: Really ? Can they accept both openvpn udp and wireguard (udp) on the same port ?? That seems odd 🤔 Or do you pick the port on their web ui for example ?

internal/provider/utils/portforward.go
qdm12: Interesting, something we can add later (or now up to you) for multiple ports

internal/configuration/settings/portforward.go
```diff
@@ -43,7 +43,8 @@ type PortForwarding struct {
 	// forwarded ports. The redirection is disabled if it is the slice [0],
 	// which is its default as well. If set and not [0], its length must match
 	// the PortsCount value, such that each forwarded port is redirected to
-	// the corresponding listening port.
+	// the corresponding listening port. For Cryptostorm, ListeningPorts[0]
+	// also specifies the port to request from the forwarding server (30000–65535).
```
qdm12, line 47: since cryptostorm appears to support multiple ports, I think it would be nice to support it here. That way that comment can be changed to:


```suggestion
	// the corresponding listening port. For VPN providers supporting requesting
	// specific ports (Cryptostorm), this is used to request ports.
```

internal/configuration/settings/portforward.go
```diff
@@ -142,14 +145,21 @@ func (p *PortForwarding) OverrideWith(other PortForwarding) {
...
 	p.ListeningPorts = gosettings.DefaultSlice(p.ListeningPorts, []uint16{0}) // disabled
-	p.PortsCount = gosettings.DefaultComparable(p.PortsCount, 1)
+	// For Cryptostorm, default PortsCount to the number of listening ports
+	// specified, since each listening port maps directly to a requested port.
+	defaultPortsCount := uint16(1)
+	if vpnProvider == providers.Cryptostorm &&
+		len(p.ListeningPorts) > 0 && p.ListeningPorts[0] != 0 {
+		defaultPortsCount = uint16(len(p.ListeningPorts))
+	}
+	p.PortsCount = gosettings.DefaultComparable(p.PortsCount, defaultPortsCount)
```
qdm12, line 162: since the Validate method already takes the vpn provider string as argument, let's just validate that the user sets the ports count to the same number of listening ports, and revert the changes to `setDefaults` here.
By the way, can a user request a random port instead of a specific port?

internal/portforward/service/start.go
```diff
@@ -109,6 +113,9 @@ func (s *Service) onNewPorts(ctx context.Context, internalToExternalPorts map[ui
 		switch {
 		case userRedirectionEnabled: // precedence over auto redirection
 			destinationPort = s.settings.ListeningPorts[i]
+			if internalPort == destinationPort {
+				continue // no-op; avoid self-redirect in firewall
+			}
```
qdm12, line 118: 👍

internal/provider/cryptostorm/updater/api.go
```diff
@@ -0,0 +1,106 @@
...
+	Location string // e.g. "Canada - Montreal", "Austria", "US - Texas - Dallas"
+	Hostname string // e.g. "austria.cstorm.is"
+	WgPubKey string
+}
+
+// fetchNodes retrieves and parses the Cryptostorm node list from their
+// wireguard page at https://cryptostorm.is/wireguard.
+func fetchNodes(ctx context.Context, client *http.Client) (
+	nodes []nodeData, warnings []string, err error) {
+	const url = "https://cryptostorm.is/wireguard"
```
qdm12, line 24: this returns servers for both wireguard and openvpn, correct?

internal/provider/cryptostorm/updater/hosttoserver.go
```diff
@@ -0,0 +1,54 @@
...
+	seen := make(map[string]struct{}, len(hts))
+	hosts = make([]string, 0, len(hts))
+	for key := range hts {
+		host := extractHost(key)
+		if _, ok := seen[host]; !ok {
+			seen[host] = struct{}{}
+			hosts = append(hosts, host)
+		}
+	}
+	return hosts
```
qdm12, line 24: nit a tiny bit shorter

```suggestion
	hosts := slices.Collect(maps.Keys(hts))
	for i, h := range hosts {
		hosts[i] = extractHost(h)
	}
	slices.Sort(hosts)
	return slices.Compact(hosts)
```

internal/provider/cryptostorm/updater/servers.go
```diff
@@ -0,0 +1,128 @@
...
+
+	"github.com/qdm12/gluetun/internal/constants/vpn"
+	"github.com/qdm12/gluetun/internal/models"
+	"github.com/qdm12/gluetun/internal/provider/common"
+)
+
+func (u *Updater) FetchServers(ctx context.Context, minServers int) (
+	servers []models.Server, err error,
+) {
+	nodes, _, err := fetchNodes(ctx, u.client)
```
qdm12, line 19: warning log the warnings returned

internal/provider/cryptostorm/updater/servers.go
qdm12: Bump on this. It's a nit, but makes everything a tiny bit simpler. You can even have map[string][2]models.Server where index 0 is wireguard and index 1 is openvpn.

internal/provider/cryptostorm/updater/servers.go
```diff
@@ -0,0 +1,128 @@
...
+	for _, host := range hosts {
+		addrs, resolveErr := net.DefaultResolver.LookupNetIP(ctx, "ip", host)
+		if resolveErr != nil {
+			// Non-fatal: skip hosts that fail to resolve
+			continue
+		}
+		hostToIPs[host] = addrs
+	}
+	return hostToIPs, nil
+}
```
qdm12, line 92: replace with a `parallelResolver common.ParallelResolver` (see other updater code for example)

internal/provider/cryptostorm/portforward.go
```diff
@@ -0,0 +1,264 @@
...
+//	<input type="hidden" name="delfwd" value="30000">
+var regexForwardHTML = regexp.MustCompile(
+	`name="delfwd"\s+value="` + portRangePattern + `"`)
+
+// portForwardURLs lists all port forwarding endpoints.
+// IPv4 and IPv6 must both be registered for dual-stack support.
+var portForwardURLs = []string{
+	"http://10.31.33.7/fwd",
+	"http://[2001:db8::7]/fwd",
+}
```
qdm12, line 45: nit just so it's "more" immutable - and then adapt callers to use `portForwardURLs()`


```suggestion
func portForwardURLs() []string {
	return []string{
		"http://10.31.33.7/fwd",
		"http://[2001:db8::7]/fwd",
	}
}
```

internal/provider/cryptostorm/portforward.go
```diff
@@ -0,0 +1,264 @@
...
+func (p *Provider) PortForward(ctx context.Context, objects utils.PortForwardObjects) (
+	internalToExternalPorts map[uint16]uint16, err error,
+) {
+	const timeout = 10 * time.Second
+	ctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+
+	// Determine the port to request:
+	// 1. Use VPN_PORT_FORWARDING_LISTENING_PORTS[0] if set (non-zero).
+	// 2. Otherwise try to read a previously persisted port.
```
qdm12, line 66: I don't really see the point of persisting the port then, the user likely would re-use the same `VPN_PORT_FORWARDING_LISTENING_PORTS` option across restarts

internal/provider/cryptostorm/portforward.go
```diff
@@ -0,0 +1,264 @@
...
+			return nil, fmt.Errorf("reading persisted port forward data: %w", err)
+		}
+		if len(data.Ports) > 0 {
+			listeningPort = data.Ports[0]
+		}
+	}
+
+	if listeningPort == 0 {
+		return nil, fmt.Errorf("port forwarding not supported: set VPN_PORT_FORWARDING_LISTENING_PORTS to a value between 30000 and 65535")
+	}
```
qdm12, line 84: this seems wrong?

Anyway likely remove if we only use `objects.ListeningPorts` from this point

internal/provider/cryptostorm/portforward.go
```diff
@@ -0,0 +1,264 @@
...
+
+	postBody := "port=" + strconv.FormatUint(uint64(listeningPort), 10)
+
+	// Register the port with both IPv4 and IPv6 endpoints.
+	// The IPv6 endpoint may not be reachable if the tunnel has no IPv6,
+	// so we log a warning rather than failing hard.
```
qdm12, line 90: you could inject the IPv6 support status to the `Provider` struct and use it here, to avoid trying a request and logging warnings when IPv6 isn't supported anyway.

internal/provider/cryptostorm/portforward.go
```diff
@@ -0,0 +1,264 @@
...
+		found, err := p.registerPort(ctx, objects.Client, url, postBody, listeningPort, objects)
+		if err != nil {
+			if url == portForwardURLs[0] {
+				// IPv4 failure is fatal.
+				return nil, err
+			}
+			// IPv6 failure is a warning only.
+			objects.Logger.Warn("IPv6 port forward registration failed (" + url + "): " + err.Error())
+			continue
+		}
```
qdm12, line 102: nit do it outside a for loop, this is a bit confusing that index 0 is ipv4 and index 1 is ipv6. Just have two code blocks following each-other, and ideally enable the ipv6 block in an if block if ipv6 is in fact supported.

internal/provider/cryptostorm/portforward.go
```diff
@@ -0,0 +1,264 @@
...
+
+	// Parse forwarded ports from the response. The server returns HTML to
+	// Go's HTTP client but plain text to curl, so we try both formats.
+	bodyStr := string(body)
+	matches := regexForwardPlainText.FindAllStringSubmatch(bodyStr, -1)
+	if len(matches) == 0 {
+		matches = regexForwardHTML.FindAllStringSubmatch(bodyStr, -1)
+	}
+	if len(matches) == 0 {
+		return false, fmt.Errorf("port forwarding not supported: no active port forwards found in response")
```
qdm12, line 159:
```suggestion
		return false, fmt.Errorf("port forwarding not supported: no active port forwards found in response: %s", bodyStr)
```

internal/provider/cryptostorm/portforward.go
```diff
@@ -0,0 +1,264 @@
...
+	if err := writePortForwardData(p.portForwardPath, portForwardData{Ports: []uint16{listeningPort}}); err != nil {
+		return nil, fmt.Errorf("persisting port forward data: %w", err)
+	}
+
+	return internalToExternalPorts, nil
+}
+
+// registerPort POSTs the port to a single forwarding endpoint and returns
+// whether the requested port was confirmed in the response.
+func (p *Provider) registerPort(ctx context.Context, client *http.Client,
```
qdm12, line 125: nit move this below KeepPortForward and above deletePortFromURL

internal/provider/cryptostorm/portforward.go
```diff
@@ -0,0 +1,264 @@
...
+
+func (p *Provider) KeepPortForward(ctx context.Context,
+	objects utils.PortForwardObjects,
+) (err error) {
+	// Cryptostorm port assignments persist for the session; no keepalive needed.
+	<-ctx.Done()
+
+	// Best-effort deregister on teardown. Use a fresh context since the
+	// original is already cancelled.
+	if p.forwardedPort != 0 {
```
qdm12, line 194: can `forwardedPort` be ever `0` when we call `.KeepPortForward`? I don't think so 🤔

internal/storage/hardcoded.go
```diff
@@ -20,7 +20,14 @@ func parseHardcodedServers() (allServers models.AllServers) {
...
 		if err != nil {
-			panic(fmt.Sprintf("reading embedded provider file %s for %s: %s", filename, provider, err))
+			// Provider not yet in the gluetun-servers module; start with an empty list.
+			// The updater can still populate servers at runtime.
+			const serversPath = "/gluetun/servers/"
+			allServers.ProviderToServers[provider] = models.Servers{
+				Version:  1,
+				Filepath: filepath.Join(serversPath, filename),
+			}
+			continue
```
qdm12, line 30: Remove this and in go.mod, add a `replace github.com/qdm12/gluetun-servers github.com/mpatton125/gluetun-servers#main` (or something of that nature) to get cryptostorm data from the PR? Typing from my phone but I'm curious if this works, since that would be useful for future VPN provider PRs!

qdm12 (changes requested):

internal/provider/cryptostorm/updater/api.go
amarevite: I think the answer to your question is yes. It's possible OpenVPN has more servers to choose from (see below), but all Wireguard servers listed there are also listed under OpenVPN.

It's not totally clear to me. The closest thing to a parallel OpenVPN page seems to be https://cryptostorm.is/configs/, which lists 45 locations. I only count 36 on the /wireguard page, but when I try to download (a few) unique OpenVPN configs, they're empty files. That implies they aren't supported/available at the moment, but they *do* have non-empty config files if you download them from https://cryptostorm.is/configs/ecc/ or https://cryptostorm.is/configs/rsa/ instead of from the generator.

I had to dig a bit to find any of this, so I assume cryptostorm prefers Wireguard as the default method to connect to their servers. From what I can tell, all Wireguard listed locations are ALSO valid OpenVPN locations (a few have only the state/country name and not the city, but there's only one in each case so they're feasibly the same), **so I think it's safe to use that list as a source of truth for both**.

qdm12: Got it, thanks for the investigation! Let's keep it to the wireguard+openvpn servers then, if someone wants more servers, they can extend the code to support them.

Issue #3188, opened 2026-03-04
Bug: ExpressVPN Cannot load CA certificate file


mjtbrady:
### Is this urgent?

None

### Host OS

OpneMediaVAult 7 (Debian 12)

### CPU arch

x86_64

### VPN service provider

ExpressVPN

### What are you using to run the container

docker-compose

### What is the version of Gluetun

(commit d8a3cc3)

### What's the problem 🤔

Just pulled the latest container to resolve the certificate issue identified in #3183, but it stills fails.

### Share your logs (at least 10 lines)

```plain text
2026-03-05T12:32:00+13:00 INFO [openvpn] OpenVPN 2.6.16 x86_64-alpine-linux-musl [SSL (OpenSSL)] [LZO] [LZ4] [EPOLL] [MH/PKTINFO] [AEAD]
2026-03-05T12:32:00+13:00 INFO [openvpn] library versions: OpenSSL 3.5.5 27 Jan 2026, LZO 2.10
2026-03-05T12:32:00+13:00 INFO [openvpn] OpenSSL: error:04800064:PEM routines::bad base64 decode:
2026-03-05T12:32:00+13:00 INFO [openvpn] Cannot load CA certificate file [[INLINE]] (no entries were read)
2026-03-05T12:32:00+13:00 INFO [openvpn] Exiting due to fatal error
2026-03-05T12:32:00+13:00 ERROR [vpn] exit status 1
2026-03-05T12:32:00+13:00 INFO [vpn] retrying in 1m0s
```

### Share your configuration

```yml

```

bfellner:
I am having the exact same issue.  I reverted to v3.41.0 and was able to connect.

malghana:
Same for me after i update, i hope this is an easy fix.

Also @qdm12 thanks for maintaining this project

malghana:
> I am having the exact same issue. I reverted to v3.41.0 and was able to connect.

Reverting back worked for me also, i'll stay on this until the issue is fixed, thanks for your reply

mjtbrady:
Reverting doesn't fix it for me and I don't understand why it would, because it doesn't it go back to using the old certificate?

I just get the TLS errors as per #3183.  Which is exactly what I would expect to see.

mjtbrady:
This error has also been noted in https://github.com/qdm12/gluetun/pull/3185#pullrequestreview-3892474756

dennisnienhuis:
I had some time today to look into it properly.

[...]

PR #3192, opened 2026-03-05
Fixed missing chars in cert

dennisnienhuis:
@qdm12 

# Description

The cert was invalid, was missing two characters.

# Issue

https://github.com/qdm12/gluetun/issues/3188

@qdm12  can you check linting for me, my linting does give other results than the repo. 

Also it seems like the https://raw.githubusercontent.com/qdm12/gluetun/master/internal/storage/servers.json is wrong for amsterdam?
```
 {
        "vpn": "openvpn",
        "country": "Netherlands",
        "city": "Amsterdam",
        "hostname": "netherlands-amsterdam-ca-version-2.expressnetw.com",
        "udp": true,
        "ips": [
          "173.239.199.138",
          "173.239.199.253"
        ]
      },
```

NS Lookup:
```
nslookup netherlands-amsterdam-ca-version-2.expressnetw.com


Non-authoritative answer:
Name:	netherlands-amsterdam-ca-version-2.expressnetw.com
Address: 173.239.199.80
Name:	netherlands-amsterdam-ca-version-2.expressnetw.com
Address: 173.239.199.183
```

@qdm12  i dont know how that list is updated, can you check this?

Also not sure if we need the old certificate which is valid untill april?

EDIT: Also the other expressvpn ip's are wrong. 

# Assertions


seibcat:
Yes, it seems that all expressvpn in server list are not updated, right?
In my case, even I try to manually update server list, it doesn’t work for expressvpn 

2026-03-05T17:18:48Z WARN reached the maximum number of consecutive failures: 2 failed attempts resolving algeria-ca-version-2.expressnetw.com: lookup algeria-ca-version-2.expressnetw.com: i/o timeout
2026-03-05T17:18:48Z WARN note: if running the update manually, you can use the flag -minratio to allow the update to succeed with less servers found
2026-03-05T17:18:48Z ERROR updating server information: getting expressvpn servers: not enough servers found: 0 and expected at least 136
2026-03-05T17:18:48Z INFO Shutdown successful

sometimes says 0 servers found and expected 136, other times 6 found and expected 12, etc

I tried dot=off and minratio 0.1 0.05, etc and it seems that maybe update a few (3 or 4) expressvpn servers, doesn’t work for most of all

PR #3208, opened 2026-03-08
feat(protonvpn): support up to 5 forwarded ports

qdm12:
# Description

- new `VPN_PORT_FORWARDING_PORTS_COUNT` option, `1` to `4` for protonvpn
- change `VPN_PORT_FORWARDING_LISTENING_PORT` to `VPN_PORT_FORWARDING_LISTENING_PORTS` to be able to redirect multiple ports

# Issue

#2381 

# Assertions


Lebowski89:
@qdm12 

Sounds like something that will get more accounts blocked. Mine was just disabled the other day out of the blue for a 'policy violation'. I emailed support asking for an explanation and they've just got back to me with this:

> Hello,

 
> Note that our automatic anti-abuse systems have noticed anomalies traffic as part of your VPN sessions, that could correspond to some bot-like traffic or some device infected with malware.Such type of traffic impacts negatively our infrastructure (and thus the experience for any other Proton VPN user), which was the reason for account suspension.
 
> https://proton.me/legal/terms 
 
> While we reinstate your account, we would like to advise you to avoid using any automated methods; disable port forwarding when not needed; keep your firewall properly configured and close unused ports and services, when using our VPN servers. In addition, we recommend running a malware scan on your device, especially when you are connected to a Proton VPN server.

> Best regards, 
Shaw
Proton Trust & Safety Customer Support Team

All I do is run Gluetun + VPN with qBittorrent and JD2. Only thing I did notice was the container has been going unhealthy after a while, which would lead to Gluetun trying to connect over and over and over again. Maybe that's why it blocked me?

I'd make a separate topic for this, but I don't have logs now and I've moved onto PIA (my account is back with Proton, but I'll use that for other things, I guess. Their stance against 'automated' methods is laughable).

qdm12:
Oh what the hell. Honestly Proton keeps on surprising me, in a bad way. It's sad because they could really excel, they just chose en-shitification I suppose 🤷  Were you running with 5 tcp+udp ports? Maybe the user could pick N TCP and M UDP ports, with a warning on unused ports? Although if it's a listening port, there could be zero traffic for a while, that's the whole point of listening port, how odd...

Lebowski89:
> Oh what the hell. Honestly Proton keeps on surprising me, in a bad way. It's sad because they could really excel, they just chose en-shitification I suppose 🤷 Were you running with 5 tcp+udp ports? Maybe the user could pick N TCP and M UDP ports, with a warning on unused ports? Although if it's a listening port, there could be zero traffic for a while, that's the whole point of listening port, how odd...

Hi,

Sorry for posting here the other day. I should have made a separate topic or replied to an existing ProtonVPN issue topic. I was pretty peeved (at Proton), and picked the first open ProtonVPN tagged topic I saw. 

How I was running Gluetun at the time of the blocking:

https://github.com/Lebowski89/homelab/blob/main/group_vars/all/services/gluetun.yml

(Ansible variables - not the actual compose file, but you can see env_variables/ports, etc)

At the time of the blocking, my Gluetun instance had been unhealthy for days and when checking the container log it had been trying to connect that entire time every few seconds (I should have recorded the log to file, sorry). So it's either something about that, or Proton simply blocked prior to that and that's why Gluetun was unhealthy in the first place. Heck, I've seen (on Reddit) people reporting that Proton had blocked them recently for simply downloading/uploading too much data. I had done some pretty big transfers (500GBish) before the block (via Jdownloader 2), so it could even just be that (I wish ProtonVPN were more specific with why they blocked). ah well, it was my wakeup call to move away from Proton, anyhow.

Dockerfile
```diff
@@ -164,11 +164,20 @@ ENV VPN_SERVICE_PROVIDER=pia \
...
     PMTUD_ICMP_ADDRESSES=1.1.1.1,8.8.8.8 \
     PMTUD_TCP_ADDRESSES=1.1.1.1:443,8.8.8.8:443,1.1.1.1:53,8.8.8.8:53,[2606:4700:4700::1111]:53,[2001:4860:4860::8888]:53,[2606:4700:4700::1111]:443,[2001:4860:4860::8888]:443 \
+    # VPN port forwarding
+    VPN_PORT_FORWARDING=off \
+    VPN_PORT_FORWARDING_LISTENING_PORTS=0 \
+    VPN_PORT_FORWARDING_PORTS_COUNT=1 \
+    VPN_PORT_FORWARDING_PROVIDER= \
+    VPN_PORT_FORWARDING_STATUS_FILE="/tmp/gluetun/forwarded_port" \
+    VPN_PORT_FORWARDING_UP_COMMAND= \
+    VPN_PORT_FORWARDING_DOWN_COMMAND= \
```
qdm12: This is duplicated, fix this

PR #3259, opened 2026-03-27
Document WireGuard healthcheck auto-restart variables

wickedyoda:
## Summary
- document the WireGuard healthcheck and auto-restart environment variables in the README compose example
- clarify the health env block in the Dockerfile
- note that the full tunnel healthcheck runs every 5 minutes and restarts the VPN on failure

## Testing
- not run (documentation-only change)


qdm12:
This is already documented at https://github.com/qdm12/gluetun-wiki/blob/main/setup/options/healthcheck.md

Issue #788, opened 2022-01-10
Feature request: Support hostnames on WIREGUARD_ENDPOINT_IP 


fishnux:
### What's the feature 🧐

Currently, WIREGUARD_ENDPOINT_IP only accepts IP addresses. Some VPN providers specify a domain name instead of an IP address on their Wireguard configuration files

### Extra information and references

```
cannot read server selection settings: environment variable WIREGUARD_ENDPOINT_IP: cannot parse IP address: example.com
```

ahmaddxb:
I also required this option

qdm12:
For now just use `docker run --rm alpine:3.15 nslookup yourvpnserver.com` to translate it to an IP address.

The current situation is gluetun doesn't do any network interaction until the VPN is connected to avoid leaking anything. I put special attention to this and to have the firewall setup as quickly as possible too.

Now this issue and others such as querying VPN provider web API to obtain Wireguard keys need to access internet before the VPN is up so it will ultimately be an option, but I can't say how long it will take me to get to it. Thanks for your understanding and of course let's keep this issue opened!

misery:
Same here to connect to my local wireguard server at home. I'm using DDNS for this.

misery:
@qdm12 Is there any work around to use ddns for own wireguard server? Any automatic shell script that could fetch the new IP and send it to the control server?
Every night the connection is broken.

TMD20:
I'm using this as a middle as a intermediate vpn
services:
  ipgen:
    image: wbitt/network-multitool:alpine-extra
    container_name: ipgen
    networks:
      - mynetwork
    command:  ["bin/bash","-c","while true;do dig europe.all.vpn.airdns.org +short |nc -l 4000;done"]


Change the entry point to this on the vpn service
    entrypoint:
          ["/bin/sh", "-c" , export VPN_ENDPOINT_IP=$$(timeout  0.1 nc ipgen 4000|shuf -n 1)&&/gluetun-entrypoint]

I imagine with this if the connection failed the healthtest, maybe it would remake the connection. 
However I just configure this today. 

It at least able to grab the IP on initial connect

pranjaljain0:
I found a way. If anyone wants help can reach out. I will post it.
Thanks

pranjaljain0:
Yes I tried it for wireguard custom and it wasnt working

pranjaljain0:
[...]

PR #3273, opened 2026-04-04
fix(wireguard): support IPv6 address formatting from config files

drewwells:
# Description

Wireguard setup with an ipv6 address was failing to parse

# Issue

ipv6 addresses in wireguard are referenced like this, the code can not parse the [] in the text
```
[Peer]
Endpoint = [2222:6666:eeee:8888::10]:51820
```
erroneously, the logs point to this issue but it is unrelated to the ipv6 parsing https://github.com/qdm12/gluetun/issues/788

# Assertions

### DIFF
--- a/internal/configuration/sources/files/wireguard.go
+++ b/internal/configuration/sources/files/wireguard.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"strings"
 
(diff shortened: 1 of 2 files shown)


drewwells:
This gets it closer. There's no ipv6 addresses in the server selection though, so more work is needed to get ipv6 end to end.

PR #3308, opened 2026-05-01
fix handshake window failure for openvpn

BenDutton:
# Description

Fixes openvpn failing due to incorrectly using `handshake-window` arg instead of `hand-window` introduced in f615e3c (see https://openvpn.net/community-docs/community-articles/openvpn-2-5-manual.html#:~:text=the%20ca%20file.-,%2D%2Dhand%2Dwindow%20n,-Handshake%20Window%20%2D%2D%20the)

# Issue

https://github.com/qdm12/gluetun/issues/3306

# Assertions

### DIFF
--- a/internal/provider/custom/openvpnconf.go
+++ b/internal/provider/custom/openvpnconf.go
@@ -76,7 +76,7 @@ func modifyConfig(lines []string, connection models.Connection,
 	modified = append(modified, "pull-filter ignore \"auth-token\"") // prevent auth failed loop
 	modified = append(modified, "auth-retry nointeract")
 	modified = append(modified, "suppress-timestamps")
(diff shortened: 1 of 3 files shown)


qdm12:
Thanks for this, although in the panic I did revert the commit with https://github.com/qdm12/gluetun/commit/7a74d4f462d2e64c7f0a8253f9d9a496a6056d79 😉 
I did add a bunch of end to end testing for openvpn so this never happens again...

- protonvpn c47cc90181d1ca8b885924a6208f2bd41a096034
- private internet access 4106951340ea674df5f460e0a81c6735b8350cf7
- airvpn 28ac8b859e9ca03d38567210e322d8ae63bb7909

And hand-window re-added in a0f6b208f789d07d5b6dc0aa1667b0447966de4c

Anyway thanks for the quick PR still 💪 !

Issue #1986, opened 2023-11-28
Bug: custom OpenVPN configuration file should be parsed when reading settings


qdm12:
### Is this urgent?

None

### CPU arch

None

### VPN service provider

Custom

### What are you using to run the container

docker run

### What is the version of Gluetun

v3.36.0+

### What's the problem 🤔

Original issue mentioning this issue: https://github.com/qdm12/gluetun/issues/1967

Openvpn configuration file parsing happens way after the initial startup, and so values in the settings are often shown as wrong. Functionally, it does work though. For example the OpenVPN protocol shows as `udp` because the OpenVPN protocol defaults to `udp` if left unset, eventhough the `proto tcp-client` is within the OpenVPN custom configuration file (and does functionally takes precedence over the udp value).

This is mostly due to:

1. settings source precedence: it's secret files then files then environment variables
2. because of 1., the read order of sources is the same, so the 'files source' isn't aware of the 'environment variables' source. This causes the files source (where the custom openvpn config should be parsed) to not be aware of the values of `VPN_SERVICE_PROVIDER` nor `OPENVPN_CUSTOM_CONFIG`, causing a bit of a cycle dependency

With gosettings v0.4.0, each individual setting would be read from each source until it's found, moving the 'merging of settings' from the highest level to the lowest level. In this case, we could easily plug in the `VPN_SERVICE_PROVIDER` and `OPENVPN_CUSTOM_CONFIG` values into the Openvpn config file parsing, which would likely be outside of the gosettings reader due to its rather custom nature.

### Share your logs (at least 10 lines)

```plain text
Irrelevant
```

PR #3327, opened 2026-05-11
fix(openvpn/extract): trim spaces in config lines before parsing

immanuwell:
Small parser papercut fix for custom OpenVPN configs.

Repro:
```go
lines := []string{"  proto tcp", "\tremote 1.2.3.4 443 tcp"}
_, err := extractDataFromLines(lines) // before: remote line not found
```

OpenVPN itself accepts those indented directives, so Gluetun should read them too. Tiny but kinda annoying when a config is formatted that way.

Fixed by trimming each line before matching `proto`, `remote`, and `port`.

Checked:
```sh
go test ./internal/openvpn/extract -run Test_extractDataFromLines/leading_whitespace -count=1
go test ./internal/openvpn/extract
go build ./...
```

Related parsing area: #1986

### DIFF
--- a/internal/openvpn/extract/extract.go
+++ b/internal/openvpn/extract/extract.go
@@ -53,6 +53,8 @@ func extractDataFromLines(lines []string) (
 func extractDataFromLine(line string) (
 	ip netip.Addr, port uint16, protocol string, err error,
 ) {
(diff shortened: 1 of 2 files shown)


qdm12 (approved):
Thanks!!

PR #3333, opened 2026-05-17
feat(socks5): add SOCKS5 proxy support with configuration and logging.

adjscent:
Working for me. Added SOCKS5 support with authentication.

Shadowsocks is good but it breaks a lot of socks5 clients.


qdm12:
closed in favor of no-dependency #3336

Issue #2848, opened 2025-07-10
Bug: Public IP endpoint sometimes stops functioning after a VPN restart


focustense:
### Is this urgent?

No

### Host OS

Linux (TrueNAS)

### CPU arch

x86_64

### VPN service provider

NordVPN

### What are you using to run the container

docker-compose

### What is the version of Gluetun

Running version latest built on 2025-01-22T08:30:14.628Z (commit 13532c8)

### What's the problem 🤔

Every so often the WireGuard connection fails, which gluetun detects with its health check and restarts accordingly. However, sometimes it gets into a state where it loses track of the public IP and is no longer able to refresh it.

Note that when this happens, _the proxy itself is still working_. If I have another container set up with network `service: gluetun`, it is still able to get out, and still has the correct IP from the VPN. This issue is _only_ with gluetun's internal IP address detection and the `/v1/publicip/ip` endpoint, which is used by tools like Homepage.

In the logs below, the 260-byte responses to `ip` are the valid responses, which eventually transition into these 17-byte empty responses. Once the empty responses start, they seem to continue indefinitely unless I restart the container.

It does look as though there is some sort of retry happening, as per #2325, but at some point it gives up and never tries again.

I want to stress that the bug report is not about the "program has been unhealthy" lines. I just assume that either the internet connection or the VPN connection flakes out occasionally for a few seconds or minutes at a time, so this is all expected. Not a big deal. The bug is that the `ip` endpoint never recovers, even after the WireGuard connection does.

### Share your logs (at least 10 lines)

```plain text
gluetun  | 2025-07-09T21:26:15Z INFO [healthcheck] healthy!
gluetun  | 2025-07-09T23:02:20Z INFO [http server] 200 GET /ip wrote 260B to 172.16.19.5:47670 in 34.642µs
gluetun  | 2025-07-09T23:05:21Z INFO [http server] 200 GET /ip wrote 260B to 172.16.19.5:38248 in 29.459µs
gluetun  | 2025-07-09T23:06:36Z INFO [http server] 200 GET /ip wrote 260B to 172.16.19.5:39414 in 26.61µs
gluetun  | 2025-07-10T01:56:16Z INFO [http server] 200 GET /ip wrote 260B to 172.16.19.5:53498 in 32.845µs
gluetun  | 2025-07-10T01:56:35Z INFO [http server] 200 GET /ip wrote 260B to 172.16.19.5:41252 in 26.348µs
[...]

Issue #1855, opened 2023-09-10
Bug: Control server returning empty public_ip


Davidmoffatt12:
### Is this urgent?

No

### Host OS

Synology DSM 7.2

### CPU arch

None

### VPN service provider

Private Internet Access

### What are you using to run the container

Other

### What is the version of Gluetun

2023-08-24T09:29.123Z(commit 1ac031e)

### What's the problem 🤔

API is returning no data for a public IP.

API Error: Invalid data
URL: http://IP:8888/v1/publicip/ip
Response Data:
{"public_ip":""}

### Share your logs

```plain text
2023/09/10 19:43:03	stdout	2023-09-10T19:43:03-03:00 INFO [http server] 200 GET /ip wrote 17B to 172.20.0.1:55778 in 71.325µs
2023/09/10 19:42:49	stdout	2023-09-10T19:42:49-03:00 INFO [http server] 200 GET /ip wrote 17B to 172.20.0.1:55680 in 66.902µs
2023/09/10 19:42:29	stdout	2023-09-10T19:42:29-03:00 INFO [http server] 200 GET /ip wrote 17B to 172.20.0.1:55586 in 59.911µs
2023/09/10 19:41:22	stdout	2023-09-10T19:41:22-03:00 INFO [http server] 200 GET /ip wrote 17B to 172.20.0.1:55408 in 63.314µs
2023/09/10 19:40:51	stdout	2023-09-10T19:40:51-03:00 INFO [http server] 200 GET /ip wrote 17B to 172.20.0.1:55300 in 59.162µs
2023/09/10 19:39:50	stdout	2023-09-10T19:39:50-03:00 INFO [http server] 200 GET /ip wrote 17B to 172.20.0.1:55090 in 82.231µs
2023/09/10 19:39:39	stdout	2023-09-10T19:39:39-03:00 INFO [http server] 200 GET /ip wrote 17B to 172.20.0.1:55022 in 73.738µs
2023/09/10 19:34:19	stdout	2023-09-10T19:34:19-03:00 INFO [dns] generate keytag query _ta-4a5c-4f66. NULL IN
2023/09/10 19:28:28	stdout	2023-09-10T19:28:28-03:00 INFO [http server] 200 GET /ip wrote 17B to 172.20.0.1:53460 in 124.194µs
[...]

Issue #3000, opened 2025-11-20
Bug:Startup check failed with airVPN "failed to pass the healthcheck: startup check:failed: dialing: dial tcp4: lookup github.com"


Asylox:
### Is this urgent?

No

### Host OS

Proxmox running Ubuntu server VM

### CPU arch

x86_64

### VPN service provider

AirVPN

### What are you using to run the container

docker-compose

### What is the version of Gluetun

latest

### What's the problem 🤔

Hello,

as you mention in the wiki, i've tried every solution in the FAQ, I tried changing ports, changing servers, adding/removing information such as the server region and city, changing `HEALTH_VPN_DURATION_INITIAL`, and changing the MTU value. Nothing worked; I couldn't mount my gluetun container.

I hope it's not something obvious that's been right in front of my eyes for the past two days. I'd feel foolish.

### Share your logs (at least 10 lines)

```plain text
gluetun     | 2025-11-20T20:10:47+01:00 INFO [storage] merging by most recent 20962 hardcoded servers and 20962 servers read from /gluetun/servers.json
gluetun     | 2025-11-20T20:10:47+01:00 INFO Alpine version: 3.22.2
gluetun     | 2025-11-20T20:10:47+01:00 INFO OpenVPN 2.5 version: 2.5.10
gluetun     | 2025-11-20T20:10:47+01:00 INFO OpenVPN 2.6 version: 2.6.14
gluetun     | 2025-11-20T20:10:47+01:00 INFO IPtables version: v1.8.11
gluetun     | 2025-11-20T20:10:47+01:00 INFO Settings summary:
gluetun     | ├── VPN settings:
gluetun     | |   ├── VPN provider settings:
gluetun     | |   |   ├── Name: airvpn
gluetun     | |   |   └── Server selection settings:
gluetun     | |   |       ├── VPN type: wireguard
gluetun     | |   |       └── Wireguard selection settings:
gluetun     | |   |           └── Server public key: PyLCXAQT8KkM4T+dUsOQfn+Ub3pGxfGlxkIApuig+hk=
gluetun     | |   └── Wireguard settings:
gluetun     | |       ├── Private key: WFv...0o=
gluetun     | |       ├── Pre-shared key: taU...RE=
gluetun     | |       ├── Interface addresses:
gluetun     | |       |   └── 213.152.161.183/32
gluetun     | |       ├── Allowed IPs:
gluetun     | |       |   ├── 0.0.0.0/0
[...]

PR #3388, opened 2026-07-09
fix(publicip): retry fetching information on failure

### LINKED ISSUE #3300: Bug: `getting public IP address information: fetching information: all fetchers failed: ... context deadline exceeded`
### Is this urgent?

No

### Host OS

Debian Trixie

### CPU arch

x86_64

### VPN service provider

ProtonVPN

docker-compose

Running version latest built on 2026-02-20T17:16:46.684Z (commit 0c3e5d9)

### What's the problem 🤔

Control server returns empty IP address. Logs indicate that the context timeout was exceeded for all ip fetchers and for version checking. As you can see in the logs below, Gluetun identifies an external IP address; however, `v1/publicip/ip` returns `{"public_ip":""}`. 

My auto-port-updater (mjmeli/qbittorrent-port-forward-gluetun-server) works, and qBittorrent allows uploads. The lack of a public IP in the control server response bothers Homepage and falsely alarms my monitoring. Therefore I consider this problem not urgent, and recognize others may disagree.

Restarting by sending `{"status":"stopped"}` and `{"status":"running"}` to the control server fixes the problem. 

[Issue #1855](https://github.com/qdm12/gluetun/issues/1855) and [issue #2848](https://github.com/qdm12/gluetun/issues/2848) appear to be the same thing, but both have been closed. 

I also see problems related to [issue #2997](https://github.com/qdm12/gluetun/issues/2997) (portforward reverts to 0 and shuts down incoming connections), but I believe that to be a separate issue.

```plain text
2026-04-25T18:38:21Z INFO [storage] creating /gluetun/servers.json with 20604 hardcoded servers
2026-04-25T18:38:21Z INFO Alpine version: 3.22.3
2026-04-25T18:38:21Z INFO OpenVPN version: 2.6.16
2026-04-25T18:38:21Z INFO IPtables version: v1.8.11
2026-04-25T18:38:21Z INFO Settings summary:
├── VPN settings:
|   ├── VPN provider settings:
|   |   ├── Name: protonvpn
|   |   ├── Server selection settings:
|   |   |   ├── VPN type: wireguard
|   |   |   ├── Countries: ****
|   |   |   ├── Port forwarding only servers: yes
|   |   |   └── Wireguard selection settings:
|   |   └── Automatic port forwarding settings:
|   |       ├── Redirection listening port: disabled
|   |       ├── Use port forwarding code for current provider
|   |       └── Forwarded port file path: /tmp/gluetun/forwarded_port
|   ├── Wireguard settings:
|   |   ├── Private key: ****
|   |   ├── Interface addresses:
|   |   |   └── ****/32
|   |   ├── Allowed IPs:
|   |   |   ├── 0.0.0.0/0
|   |   |   └── ::/0
|   |   └── Network interface: tun0
|   |       └── MTU: use path MTU discovery.
.
.
.
├── Public IP settings:
|   ├── IP file path: /tmp/gluetun/ip
|   ├── Public IP data base API: ipinfo
|   └── Public IP data backup APIs:
|       ├── ifconfigco
|       ├── ip2location
|       └── cloudflare
└── Version settings:
    └── Enabled: yes
2026-04-25T18:38:21Z INFO [routing] default route found: interface eth0, gateway 172.24.0.1, assigned IP 172.24.0.2 and family v4
.
.
.
2026-04-25T18:38:32Z INFO [http server] 200 GET /v1/portforward wrote 11B to 172.24.0.2:55680 in 50.53µs
2026-04-25T18:38:32Z INFO [dns] ready
2026-04-25T18:38:32Z INFO [http server] 200 GET /v1/portforward wrote 11B to 172.24.0.2:55696 in 36.95µs
2026-04-25T18:38:33Z INFO [http server] 200 GET /v1/portforward wrote 11B to 172.24.0.2:55698 in 30.618µs
2026-04-25T18:38:35Z INFO [http server] 200 GET /v1/portforward wrote 11B to 172.24.0.2:55700 in 33.726µs
2026-04-25T18:38:37Z ERROR [vpn] getting public IP address information: fetching information: all fetchers failed: ipinfo: Get "https://ipinfo.io/": context deadline exceeded
ifconfig.co: Get "https://ifconfig.co/json": context deadline exceeded
ip2location: Get "https://api.ip2location.io/": context deadline exceeded
cloudflare: Get "https://speed.cloudflare.com/meta": context deadline exceeded
2026-04-25T18:38:38Z INFO [http server] 200 GET /v1/portforward wrote 11B to 172.24.0.2:55712 in 37.718µs
2026-04-25T18:38:42Z ERROR [vpn] cannot get version information: Get "https://api.github.com/repos/qdm12/gluetun/commits": context deadline exceeded
2026-04-25T18:38:42Z INFO [port forwarding] starting
2026-04-25T18:38:42Z INFO [port forwarding] gateway external IPv4 address is 79.127.184.193
2026-04-25T18:38:43Z INFO [port forwarding] port forwarded is 51792
2026-04-25T18:38:43Z INFO [firewall] setting allowed input port 51792 through interface tun0...
```

### Share your configuration

```yml
services:
  gluetun:
    image: qmcgaw/gluetun
    container_name: gluetun
    cap_add:
      - NET_ADMIN
    devices:
      - /dev/net/tun:/dev/net/tun
    networks:
      - ripperrnet
    environment:
      - TZ=${TZ}
      - HTTP_CONTROL_SERVER_AUTH_CONFIG_FILEPATH=/config/auth.toml
      - VPN_SERVICE_PROVIDER=protonvpn
      - VPN_TYPE=wireguard
      - WIREGUARD_PRIVATE_KEY=${WIREGUARD_PRIVATE_KEY}
      - SERVER_COUNTRIES=redacted,comma,separated,list # Best privacy laws
      - SECURE_CORE_ONLY=off # Enable for better privacy, bolstering less-private server countries.
      - PORT_FORWARD_ONLY=on # TODO: does port forwarding work with secure core?
      - VPN_PORT_FORWARDING=on
      - BORINGPOLL_GLUETUNCOM=on # Poll AI slop scamsite
    ports:
      - ${GLUETUN_SERVER_PORT}:${GLUETUN_SERVER_PORT}
      - ${QBT_WEBUI_PORT}:${QBT_WEBUI_PORT}
      - ${PROWLARR_PORT}:${PROWLARR_PORT}
      - ${FLARESOLVERR_PORT}:${FLARESOLVERR_PORT}
    volumes:
      - /etc/localtime:/etc/localtime:ro
      - ${APP_DIR}/gluetun:/gluetun
      - ${APP_DIR}/gluetun/config:/config
    restart: unless-stopped
    logging:
      driver: "json-file"
      options:
        max-size: "50m"
        max-file: "3"
    labels:
      homepage.group: Network
      homepage.name: Gluetun
      homepage.icon: gluetun
      homepage.href: "http://${OVERKILL}:${GLUETUN_SERVER_PORT}"
      homepage.weight: 1000
      homepage.description: Manages the ProtonVPN connection
      homepage.widget.type: gluetun
      homepage.widget.version: 2
      homepage.widget.fields: '["public_ip", "country", "port_forwarded"]'
      homepage.widget.url: "http://${OVERKILL}:${GLUETUN_SERVER_PORT}/"
      homepage.widget.key: ${GLUETUN_API_KEY}
```

yorah:
# Description

The public IP information is fetched only once per tunnel-up (`onTunnelUp` → `RunOnce`). That single fetch runs seconds after reconnection, when the tunnel is often still flaky, and if it fails, the data stays empty until the *next* tunnel-up, potentially days later, with `/v1/publicip/ip` serving `{"public_ip":""}` the whole time (breaking Homepage widgets and monitoring, as reported in #3300).

Real-world example (v3.41.1, WireGuard/AirVPN): after a ~25 min provider outage, the healthcheck-triggered reconnect succeeded, but the tunnel-up fetch timed out on all four sources (`all fetchers failed: … Client.Timeout exceeded`). The tunnel stabilized seconds later, yet the IP stayed empty for days because nothing ever retried.

This PR makes the fetch self-healing:
- on failure, a retry is scheduled with exponential backoff: 10s, doubling, capped at 5 minutes, until success;
- a new tunnel-up trigger supersedes any pending retry and resets the backoff;
- retries stop when the tunnel goes down (`ClearData`) or if the public IP feature is disabled.

No new settings (no wiki change needed), and zero extra API load when fetches succeed. Includes two unit tests (pass with `-race`); `golangci-lint` clean; `docker build .` succeeds.

# Issue

Fixes #3300 (also previously reported as #1855, #2848).

# Assertions

### DIFF
--- a/internal/publicip/data.go
+++ b/internal/publicip/data.go
@@ -13,6 +13,10 @@ func (l *Loop) GetData() (data models.PublicIP) {
 // ClearData is used when the VPN connection goes down
 // and the public IP is not known anymore.
 func (l *Loop) ClearData() (err error) {
+	// Stop retrying to fetch the public IP information until the
+	// next tunnel up trigger, since the VPN tunnel is going down.
+	l.retryWanted.Store(false)
+
 	l.ipDataMutex.Lock()
 	defer l.ipDataMutex.Unlock()
 	l.ipData = models.PublicIP{}
--- a/internal/publicip/loop.go
+++ b/internal/publicip/loop.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"net/netip"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/qdm12/gluetun/internal/configuration/settings"
@@ -20,12 +21,19 @@ type Loop struct {
 	ipData        models.PublicIP
 	ipDataMutex   sync.RWMutex
 	fetcher       *api.ResilientFetcher
+	// retryWanted is true when the last fetch failed and a retry
+	// is scheduled. It is set to false when the VPN tunnel goes down
+	// (see ClearData) so retries stop until the next tunnel up trigger.
+	retryWanted atomic.Bool
 	// Fixed injected objects
 	httpClient *http.Client
 	logger     Logger
 	// Fixed parameters
 	puid int
 	pgid int
+	// Retry backoff parameters, set as fields for test purposes.
+	retryInitialBackoff time.Duration
+	retryMaxBackoff     time.Duration
 	// Internal channels and locks
 	// runCtx is used to detect when the loop has exited
 	// when performing an update
@@ -40,6 +48,11 @@ type Loop struct {
 	timeNow func() time.Time
 }
 
+const (
+	defaultRetryInitialBackoff = 10 * time.Second
+	defaultRetryMaxBackoff     = 5 * time.Minute
+)
+
 func NewLoop(settings settings.PublicIP, puid, pgid int,
 	httpClient *http.Client, logger Logger,
 ) (loop *Loop, err error) {
@@ -49,13 +62,15 @@ func NewLoop(settings settings.PublicIP, puid, pgid int,
 	}
 
 	return &Loop{
-		settings:   settings,
-		httpClient: httpClient,
-		fetcher:    api.NewResilient(fetchers, logger),
-		logger:     logger,
-		puid:       puid,
-		pgid:       pgid,
-		timeNow:    time.Now,
+		settings:            settings,
+		httpClient:          httpClient,
+		fetcher:             api.NewResilient(fetchers, logger),
+		logger:              logger,
+		puid:                puid,
+		pgid:                pgid,
+		retryInitialBackoff: defaultRetryInitialBackoff,
+		retryMaxBackoff:     defaultRetryMaxBackoff,
+		timeNow:             time.Now,
 	}, nil
 }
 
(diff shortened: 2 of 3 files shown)


qdm12:
Thanks for the contribution. Unfortunately that's a lot of added complexity for a bandaid fix which does not address the true issue here.

> That single fetch runs seconds after reconnection, when the tunnel is often still flaky

It should not be "flaky". The "full" healthcheck (TLS+DNS check) passed before fetching the public ip information, so the tunnel should be already working well. If there was an outage or unhealthy connections beforehand, it shouldn't matter for this new connection.

On top of this, in issue #3300 you can see there are other issues such as `cannot get version information` so focusing exclusively on the public ip part is not good enough. We should address the root cause, mostly related to the http client timeout.

If this is still unresolved with the latest image which *should* have a 15s timeout (double check with log timestamps), we can proceed with the bandaid fix for the public ip but with a fixed number of retries (3) and no retry wait time, to keep code simple - if it still fails after 3 x 15s retries, it probably won't work ever.

yorah:
Thanks for your feedback!

To be more precise, the few times I observed this behavior was when AirVPN had an outage. And I think the "flakiness" was on their side (due to everyone trying to reconnect to the serve rthat just went up at the same time).

I will re-check my logs when I get back home, and report back. I remember I am using 3.41.1, which I believe is the last release.

qdm12:
try the latest image It's about 7 months ahead of v3.41.1 and not the same as v3 or v3.41.1

yorah:
So after one week of running the latest image (I was previously pinned to 3.41.1), I haven't had the issue anymore. I'll close this PR as unneeded, and comment back if I ever encounter the problem again. Thanks!

yorah:
Reopening: I got this again on the latest image (commit 78f6076, built 2026-07-19).

This was the AirVPN outage scenario I mentioned earlier. It started here, with the small ICMP check failing, followed by path MTU discovery getting nothing back and a TCP dial to a bare IP timing out:

```
01:45:05 WARN [vpn] restarting VPN because it failed to pass the healthcheck: small periodic check: all check tries failed:
01:45:11 INFO [MTU discovery] reverting VPN interface tun0 MTU to 1320 (due to: PMTUD failed with both ICMP and TCP)
01:45:17 WARN [vpn] restarting VPN because it failed to pass the healthcheck: startup check: all check tries failed: parallel attempt 1/2 failed: dialing: dial tcp4: lookup github.com: i/o timeout, parallel attempt 2/2 failed: dialing:
dial tcp4 104.16.133.229:443: i/o timeout
```

Nothing was reachable through the tunnel for about ten minutes, by name or by IP, and the VPN restarted 30 times in that window. Then the connection came back:

```
01:56:30 WARN [vpn] restarting VPN because it failed to pass the healthcheck: ...
01:56:48 ERROR [vpn] getting public IP address information: fetching information: all fetchers failed:
ipinfo: Get "https://ipinfo.io/": context deadline exceeded
ifconfig.co: Get "https://ifconfig.co/json": context deadline exceeded
ip2location: Get "https://api.ip2location.io/": dial tcp [2606:4700:20::681a:681]:443: connect: network is unreachable
cloudflare: Get "https://speed.cloudflare.com/meta": context deadline exceeded
```

The tunnel passed its startup healthcheck at 01:56:30, and the public IP fetch still failed about 15 seconds later on all four APIs. The 15s timeout is clearly applied and does not help, because the connection was not slow, it was
still not carrying traffic reliably in those first seconds after reconnect. So a passing healthcheck does not mean the fetch will succeed.

My guess is that this is a thundering herd effect: when an AirVPN server comes back, every client that was on it reconnects at the same time, and the exit stays congested for a short while afterwards. That would explain why three of
the four APIs hit the deadline rather than failing outright, and why it clears up on its own shortly after.

Worth noting that the periodic healthcheck already does exactly what this PR proposes: withRetries with escalating 10s, 15s, 30s timeouts, with the comment "20s timeout in case the connection is under stress". The public IP fetch is a
strictly heavier operation, a full HTTPS GET with a response body against third party hosts rather than a TLS handshake against an anycast edge, and it gets a single attempt. In my log the healthcheck handshaked with cloudflare.com:443
successfully while the GET to speed.cloudflare.com/meta timed out in the same window. Same network, same seconds, just a heavier request.

What do you think about that analysis? I can do other tests if you have other ideas.

Also happy to rewrite this PR as you suggested, with 3 fixed retries and no wait time (althouhg in a thundering herd problem, adding some wait could be beneficial). Let me know and I will force-push.

yorah:
On addressing the root cause, here is a proposal that also covers the version check. Tell me if you like it before I write it.

Both failing (cannot get version information from #3000, and mine) fetches sit in `onTunnelUp`, back to back under `vpnCtx`, both idempotent GETs with a 15s deadline and no retry. The version one is worse, since l.versionInfo = false means one failure loses it until restart.

Proposal: promote your `withRetries` from internal/healthcheck/checker.go into a small shared package, generic over the result:

```
type Policy struct {
    Timeouts []time.Duration // per-attempt deadline; len = max attempts
    Waits    []time.Duration // optional pause before attempt i; nil for none
}

func Do[T any](ctx context.Context, policy Policy, logger Logger, name string,
    op func(ctx context.Context, attempt int) (T, error)) (T, error)
```

Three users: healthcheck unchanged with {10s, 15s, 30s} and no waits, publicip wrapping `FetchInfo` above `ResilientFetcher`, and version wrapping `GetMessage`.

I looked at putting a retrying `RoundTripper` on the shared httpClient instead, but all attempts would then share the single 15s request context, so every call site deadline would need widening anyway, and the port forwarding calls are
not all idempotent.

This also lets me drop the `retryWanted` atomic and the `ClearData` change. `vpnCtx` is already cancelled on tunnel down, so retries stop by themselves. That was the unnecessary complexity you flagged.

If the shape works I will rewrite the PR this way. If you would rather keep it to publicip with 3 fixed retries and no shared package, say so and I will do that instead. Or if you have other ideas, I'm also interested :)

yorah:
For info, I got the problem again today, using the very latest version of gluetun, so I think this fix is still warranted. Ping me whenever if you want me to rework it as per my proposals above.

yorah:
Still getting the problem from time to time, whenever you have the time to talk about it, I'm ready to go back to the solution phase. No rush.

Issue #3358, opened 2026-06-04
Feature request: restricted internet when VPN tunnel is down


qdm12:
### What's the feature 🧐

Gluetun is designed to not allow any internet traffic outside the VPN tunnel at any time, especially to prevent connected containers from leaking data outside the tunnel. To do this it uses the firewall (see `internal/firewall` and `internal/firewall/iptables`). However this adds many limitations, for example it cannot resolve VPN server hostnames or communicate with a VPN provider API endpoint to obtain credentials or servers information.

However, there is already one check in place to check for internet IPv6 support which goes outside the VPN tunnel only once at program start, defined in `internal/netlink/ipv6.go` in the `dialAddrThroughFirewall` function. It does this by only allowing traffic for this narrow request through the firewall, so nothing else can leak out. We want to expand this a bit with this.

The goal is ONLY to implement a package `internal/restrictednet` with two objectives:

1. A way to DNS resolve domain names using DNS over HTTPs through the firewall
2. A way to run https requests through the firewall

To have a restrictive output firewall, we want to narrow the filtering using a source ip, **source port**, destination ip and destination port. Ideally we would use ip packet marking and matching but unfortunately many older/odd kernels do not support that, so we rely on binding to a source port first and allowing traffic from that source port only, ephemerally.

The DNS resolution should use the DoH server(s) corresponding to what the user picked with `DNS_UPSTREAM_RESOLVERS`, no matter what `DNS_UPSTREAM_RESOLVER_TYPE` is. The package `github.com/qdm12/dns/v2/pkg/provider` has each DoH provider with a list of both IPv4 and IPv6 addresses for each DoH server URL. One can use the exported function `NewResolverHardcodedIPs` from `github.com/qdm12/dns/v2/pkg/doh` to create a dialer with these hardcoded IPs to avoid having to do a plaintext lookup. For example:

```go
dohServers := []provider.DoHServer{...}
ipVersion := "ipv6"
httpTransport := http.DefaultTransport.(*http.Transport).Clone()
dialer := &net.Dialer{
	Resolver: doh.NewResolverHardcodedIPs(dohServers, ipVersion),
}
httpTransport.DialContext = dialer.DialContext
return &http.Client{
	Transport: httpTransport,
}
```

[...]

PR #3393, opened 2026-07-13
feat(pia): native WireGuard support with port forwarding

### LINKED ISSUE #3070: Bug: PIA Port Forwarding fails with VPN_SERVICE_PROVIDER=custom (WireGuard): Panic or Region Validation Error
### Is this urgent?

No

### Host OS

Debian Trixie (Linux)

### CPU arch

None

### VPN service provider

Private Internet Access

docker-compose

Running version latest built on 2025-12-24T20:33:57.715Z (commit a10349e)

### What's the problem 🤔

I am trying to use Private Internet Access (PIA) with WireGuard using a manual configuration file (VPN_SERVICE_PROVIDER=custom), while keeping the automatic Port Forwarding feature enabled.

There is a conflict in the configuration logic that prevents this setup from working:

Scenario A: If I do not specify SERVER_REGIONS, the Port Forwarding service starts but panics immediately because it expects a server name/region to be present in the internal state (which is empty for custom provider).
Scenario B: If I do specify SERVER_REGIONS (e.g., Netherlands) to prevent the panic, the container refuses to start because region selection is not allowed for the custom provider.
To Reproduce

Generate a valid wg0.conf for PIA (e.g. using official PIA scripts).
Configure docker-compose.yml with VPN_SERVICE_PROVIDER=custom, VPN_TYPE=wireguard, and VPN_PORT_FORWARDING=on.
Set VPN_PORT_FORWARDING_PROVIDER=private internet access and provide credentials.
Expected behavior
It should be possible to use custom WireGuard configuration (using a PIA config file) and still utilize the built-in PIA Port Forwarding mechanism. Gluetun should allow specifying the region manually (e.g., via SERVER_REGIONS or a dedicated variable) when in custom mode, so the Port Forwarding script knows which API endpoint to contact without triggering a validation error or a panic.

```plain text
Scenario A (Panic - without SERVER_REGIONS):

INFO [port forwarding] starting
panic: server name cannot be empty

goroutine 98 [running]:
github.com/qdm12/gluetun/internal/provider/privateinternetaccess.(*Provider).PortForward(...)
    github.com/qdm12/gluetun/internal/provider/privateinternetaccess/portforward.go:32 +0x7a5
...

Scenario B (Validation Error - with SERVER_REGIONS=Netherlands):

ERROR VPN settings: provider settings: server selection: for VPN service provider custom: the region specified is not valid: one or more values is set but there is no possible value available
```

### Share your configuration

```yml
gluetun:
    image: qmcgaw/gluetun
    cap_add:
      - NET_ADMIN
    environment:
      - VPN_SERVICE_PROVIDER=custom
      - VPN_TYPE=wireguard
      # Port Forwarding settings
      - VPN_PORT_FORWARDING=on
      - VPN_PORT_FORWARDING_PROVIDER=private internet access
      - VPN_PORT_FORWARDING_USERNAME=myuser
      - VPN_PORT_FORWARDING_PASSWORD=mypass
      # - SERVER_REGIONS=Netherlands # Uncommenting this causes Validation Error, commenting it out causes Panic
    volumes:
      - ./data/gluetun/wg0.conf:/gluetun/wireguard/wg0.conf:ro
```

neilcorp2kx:
## Summary

This adds **native WireGuard support for Private Internet Access (PIA), including port forwarding**, plus a small healthcheck robustness improvement that the native path depends on.

Today PIA is OpenVPN-only in gluetun, and PIA + WireGuard + port forwarding is a long-standing gap (#3070, #2147): the custom-provider workaround panics with `server name cannot be empty`, and PIA cannot be selected as a WireGuard provider at all. The underlying reason is that PIA does not hand out static WireGuard configs — it registers a **fresh WireGuard key per connection** through its own API, which does not fit gluetun's normally-static WireGuard model.

This PR implements that dynamic registration inside the PIA provider so the whole flow is just credentials + region:

```yaml
environment:
  - VPN_SERVICE_PROVIDER=private internet access
  - VPN_TYPE=wireguard
  - OPENVPN_USER=...            # or _SECRETFILE
  - OPENVPN_PASSWORD=...        # or _SECRETFILE
  - SERVER_REGIONS=CA Vancouver
  - VPN_PORT_FORWARDING=on
  - VPN_PORT_FORWARDING_PROVIDER=private internet access
```

No externally-generated WireGuard config, no static keys, no helper scripts. Server selection and key rotation happen automatically on every (re)connect.

Closes #3070.

## What's in it

Three focused commits:

1. **`feat(firewall)`** — scoped temporary bootstrap connection allowances. A registry of narrowly-scoped (destination IP + protocol + port + firewall mark, physical interface) temporary outbound allowances needed *before* the tunnel is up, with retain-on-failure cleanup, idempotent deletion under a bounded non-cancelable context, and a sweep on shutdown / before reconnect. iptables gains `-m mark --mark` emit+parse so the rules are mark-scoped and delete symmetrically.
2. **`feat(pia)`** — native PIA WireGuard + port forwarding. At connect time the provider fetches PIA's live server list, selects a WireGuard server for the requested region/name/hostname, obtains a token, generates an ephemeral Curve25519 key pair, registers the public key (`addKey`) over TLS pinned to the server CN via the bundled PIA CA, and builds the WireGuard connection from the response (re-registering each reconnect). Port forwarding works natively (gateway from the `addKey` `server_vip`).
3. **`feat(healthcheck)`** — retry the startup TCP+TLS check within a 60s budget (2s backoff) instead of a single 6s attempt, so a tunnel whose DNS takes a few seconds to become ready doesn't trigger a restart loop. General improvement, not PIA-specific.

## Design / safety notes

- **Killswitch integrity was the top priority.** The pre-tunnel token / server-list / `addKey` calls are the only new outbound traffic before the tunnel exists. They go through the temporary, mark-tagged, exact-destination allowances above, resolved by a bootstrap dialer pinned to the physical default route, and are torn down before the tunnel comes up. Discovery/registration failure fails closed (the tunnel does not start). There is no fail-open to arbitrary destinations.
- **Secrets:** the ephemeral private key, the auth token and account credentials are never logged (info logging shows only the derived public key, shortened) or persisted; the port-forward token file is kept `0600`.
- Malformed/unsupported provider data (invalid/unspecified/IPv6 `server_ip`/`server_vip`/`peer_ip`) is rejected with errors rather than panicking.

## Testing

- Unit tests added across the firewall (rollback, deletion-failure retry, idempotence, mark scoping, reconnect de-duplication, mark parsing edge cases), PIA WireGuard (server-list parsing, region/name/hostname selection, `addKey` request/response mapping), healthcheck retry, and settings validation.
- Built and run end-to-end against a live PIA account: native config connects over WireGuard, obtains a forwarded port, passes healthchecks with 0 restarts, and DNS/egress resolve to the PIA endpoint. Post-startup firewall state was inspected to confirm every temporary bootstrap allowance is torn down (only the active WireGuard endpoint allowance remains).
- It has been running on a real seedbox to soak before this is taken out of draft.

## Known / deferred (pre-existing, not introduced here)

- PIA WireGuard reuses `OpenVPN.User`/`Password` for credentials, which are serialized by the authenticated `/v1/vpn/settings` control endpoint like all other settings — a pre-existing exposure, called out here because the native path now depends on those fields. A redacted control DTO would be a separate change.
- The `addKey` client trusts the bundled PIA CA additively on top of the system pool rather than pinning PIA-only. TLS name verification is enabled and there is no `InsecureSkipVerify`.

## A note on how this was developed

This was built and hardened with heavy, deliberate use of AI tooling — a combination of OpenAI Codex and Anthropic Claude — for implementation, iterative debugging against a live PIA endpoint, and multiple adversarial code-review passes focused on killswitch/leak safety, with all findings driven to resolution and re-verified. Flagging this transparently; every change has been reviewed and validated by a human, and I'm happy to walk through any part of the design or rationale.

---

Opening as a **draft** intentionally — soaking on a live setup for a bit before marking it ready. Feedback on the approach (especially the bootstrap-firewall handling and whether a redacted control DTO should be bundled) very welcome.


qdm12:
Finally a decent solution, thanks! Although the iptables marking mechanism is risky, given it's not always available on the user's kernel. Instead you should more simply use https://github.com/passteque/gluetun/pull/3361 (reasoning behind this at https://github.com/passteque/gluetun/issues/3358)

unrealtournament:
I've been testing this for a day and it works but the connection died after around 24h and didn't come back automatically, started getting spammed in the logs with:

```
gluetun      | 2026-07-23T21:40:43+02:00 INFO [port forwarding] starting
gluetun      | 2026-07-23T21:40:43+02:00 ERROR [port forwarding] starting port forwarding service: getting VPN assigned IP address: network interface tun0 not found: route ip+net: no such network interface - retrying in 5s
```

Also after reconnecting manually (docker compose down + up) it seems like the saved forwarded port is not properly refreshed, I got connected to a different server in the same region but it tried to reuse the old port (which is obviously not valid for the new IP). I had to manually delete the `piaportforward.json` file and restart again.

I'm not too familiar with gluetun config so this last part might just be something you can fix with some config options, but I didn't find anything at a glance.

My config:
```
      - VPN_SERVICE_PROVIDER=private internet access
      - VPN_TYPE=wireguard
      - OPENVPN_USER=<redacted>
      - OPENVPN_PASSWORD=<redacted>
      - SERVER_REGIONS=France
      - VPN_PORT_FORWARDING=on
      - VPN_PORT_FORWARDING_PROVIDER=private internet access

```

Thank you for the PR! Will report back if it happens again and I have any extra info.

unrealtournament:
More logs, this time it took 6 days for the same thing to happen:

```
gluetun      | 2026-07-23T22:32:34+02:00 INFO [ip getter] Public IP address is <redacted> - source: ipinfo+ifconfig.co+ip2location+cloudflare)
gluetun      | 2026-07-23T22:32:34+02:00 INFO [vpn] There is a new release v3.41.1 (v3.41.1) created 162 days ago
gluetun      | 2026-07-23T22:32:34+02:00 INFO [port forwarding] starting
gluetun      | 2026-07-23T22:32:40+02:00 INFO [port forwarding] Port forwarded data expires in 62 days
gluetun      | 2026-07-23T22:32:40+02:00 INFO [port forwarding] port forwarded is 46626
gluetun      | 2026-07-23T22:32:40+02:00 INFO [firewall] setting allowed input port 46626 through interface tun0...
gluetun      | 2026-07-23T22:32:40+02:00 INFO [port forwarding] writing port file /tmp/gluetun/forwarded_port
gluetun      | 2026-07-24T22:32:31+02:00 INFO [dns] downloading hostnames and IP block lists
gluetun      | 2026-07-25T22:32:32+02:00 INFO [dns] downloading hostnames and IP block lists
gluetun      | 2026-07-26T22:32:32+02:00 INFO [dns] downloading hostnames and IP block lists
gluetun      | 2026-07-27T22:32:32+02:00 INFO [dns] downloading hostnames and IP block lists
gluetun      | 2026-07-28T22:32:33+02:00 INFO [dns] downloading hostnames and IP block lists
gluetun      | 2026-07-29T10:19:07+02:00 WARN [vpn] restarting VPN because it failed to pass the healthcheck: small periodic check: all check tries failed:
gluetun      |  attempt 1 (5000ms): timed out waiting for ICMP echo reply from 1.1.1.1
gluetun      |  attempt 2 (5000ms): timed out waiting for ICMP echo reply from 8.8.8.8
gluetun      |  attempt 3 (5002ms): timed out waiting for ICMP echo reply from 1.1.1.1
gluetun      |  attempt 4 (10000ms): timed out waiting for ICMP echo reply from 8.8.8.8
gluetun      |  attempt 5 (10000ms): timed out waiting for ICMP echo reply from 1.1.1.1
gluetun      |  attempt 6 (10000ms): timed out waiting for ICMP echo reply from 8.8.8.8
gluetun      |  attempt 7 (15001ms): timed out waiting for ICMP echo reply from 1.1.1.1
gluetun      |  attempt 8 (15000ms): timed out waiting for ICMP echo reply from 8.8.8.8
gluetun      |  attempt 9 (15000ms): timed out waiting for ICMP echo reply from 1.1.1.1
gluetun      |  attempt 10 (30001ms): timed out waiting for ICMP echo reply from 8.8.8.8
gluetun      | 2026-07-29T10:19:07+02:00 INFO [vpn] 👉 See https://github.com/qdm12/gluetun-wiki/blob/main/faq/healthcheck.md
gluetun      | 2026-07-29T10:19:07+02:00 INFO [vpn] DO NOT OPEN AN ISSUE UNLESS YOU HAVE READ AND TRIED EVERY POSSIBLE SOLUTION
gluetun      | 2026-07-29T10:19:07+02:00 INFO [vpn] stopping
gluetun      | 2026-07-29T10:19:07+02:00 ERROR [vpn] stopping port forwarding: server name not set
gluetun      | 2026-07-29T10:19:07+02:00 INFO [vpn] starting
gluetun      | 2026-07-29T10:19:13+02:00 INFO [firewall] removing allowed port 46626...
gluetun      | 2026-07-29T10:19:13+02:00 INFO [port forwarding] clearing port file /tmp/gluetun/forwarded_port
gluetun      | 2026-07-29T10:19:13+02:00 ERROR [port forwarding] binding port: Get "https://10.29.0.1:19999/bindPort?payload=<payload>&signature=<signature>": context deadline exceed
ed
gluetun      | 2026-07-29T10:19:13+02:00 INFO [port forwarding] starting
gluetun      | 2026-07-29T10:19:13+02:00 ERROR [port forwarding] starting port forwarding service: getting VPN assigned IP address: network interface tun0 not found: route ip+net: no
 such network interface - retrying in 5s
gluetun      | 2026-07-29T10:19:17+02:00 ERROR [vpn] finding a VPN server: resolving PIA server list host: lookup serverlist.piaservers.net: i/o timeout
gluetun      | 2026-07-29T10:19:17+02:00 INFO [vpn] retrying in 15s
gluetun      | 2026-07-29T10:19:18+02:00 INFO [port forwarding] starting
gluetun      | 2026-07-29T10:19:18+02:00 ERROR [port forwarding] starting port forwarding service: getting VPN assigned IP address: network interface tun0 not found: route ip+net: no such network interface - retrying in 5s
...
(endless loop)
```

Harland70:
I spent some time testing this branch against a real PIA account (CA Toronto, `PORT_FORWARD_ONLY=true`)
because I'd like to see PIA WireGuard land. It connects and forwards a port fine, but I hit a
reproducible failure and, while chasing it, three other things that look like they'd block a merge.
Sharing it all in case it's useful. (Analysis done with AI assistance. Every claim below is from a
run I did myself, and I've tried to flag the bits I could not prove.)

## 1. A killed tunnel never recovers

`GetWireguardConnection` resolves `serverlist.piaservers.net` on every connect. gluetun's DNS loop
only starts on tunnel-up, and its upstream socket is unmarked, so once the tunnel is down that
lookup can't succeed, and the reconnect needs it. It retries forever:

```
WARN  [vpn] restarting VPN because it failed to pass the healthcheck
ERROR [vpn] finding a VPN server: resolving PIA server list host:
            lookup serverlist.piaservers.net: i/o timeout
```

Reproduced with `ip link delete tun0` and letting the healthcheck react. Still unhealthy, no
interface, no IP after 5+ minutes.

I think this is what @unrealtournament is reporting above. The trigger is whatever kills the
tunnel, but the reason it never comes back is this.

**Control**, to make sure I wasn't blaming the wrong thing: same tree, reverting *only* the
resolver, everything else identical. Never recovered through +320s. With a tunnel-independent
resolver wired in, it recovers. I'd treat "recovers vs never recovers" as the result. I can't
attribute the exact timing, since the 60s startup budget and the 15s-doubling backoff are both in
play.

## 2. `restrictednet` looks like the fix, and it's already merged

@qdm12's `internal/restrictednet` (#3361 / #3358) does DoH resolution and HTTPS through narrow
temporary firewall openings, i.e. exactly a lookup that doesn't depend on the tunnel. I wired
`ResolveName` in as the `lookupNetIP` passed to `GetWireguardConnection` and the recovery problem
goes away.

Worth noting this branch is based on `93cc5a4` (2026-06-29) and `restrictednet` merged after that
(`78f6076`), so it isn't available here yet. A rebase onto master would be needed regardless. That
would also address the `fwmark` concern raised above, since `restrictednet` is the source-port
approach described in #3358.

⚠️ My wiring is a proof of concept and **not** something I'd propose merging. It only replaces the
resolver, so `bootstrapDialContext` (`SO_MARK`) is still load-bearing for the serverlist fetch, the
token fetch and addKey, and `OpenHTTPS`/`OpenHTTPSByHostname` presumably want to cover those three
too. It also has real bugs: it drops the `"ip4"` argument, and `serverlist.piaservers.net` has AAAA
records; and it reads `DNS.Providers` without honouring `UpstreamPlainAddresses`, so it can
substitute Cloudflare for a user's own resolver. I mention it only as evidence the approach works.
Happy to share the diff if it's useful, but it needs doing properly.

## 3. Port forwarding isn't stopped when the tunnel drops

```
ERROR [vpn] stopping port forwarding: server name not set
```

`stopPortForwarding` sends `Settings{VPNIsUp: false}` with a zero `Service` struct, which now clears
`ServerName`. Validation runs against the *pre-update* `VPNIsUp` (still true), so it fails, the
settings are never applied, and `service.Stop()` isn't reached. `KeepPortForward` then keeps calling
`bindPort` through a dead tunnel for the whole outage.

A/B of `Settings.updateWith` against merge-base `93cc5a4`:

| branch | result |
|---|---|
| `93cc5a4` | `err=<nil>`, `serverName="vancouver439"` |
| this branch | `err=server name not set`, `serverName=""` |

## 4. Saved port-forward data has no server identity

This is the other half of @unrealtournament's report (reconnecting to a different server reuses a
port that isn't valid, needing a manual `piaportforward.json` delete). Reproduced live. After a
reconnect:

```
INFO [port forwarding] Found saved forwarded port data for port 45511
INFO [port forwarding] port forwarded is 45511      <- different server than issued it
```

The saved file holds only `["expires_at","port","signature","token"]`, nothing identifying the
server or gateway. The reuse gate checks expiry only, and PIA's tokens last ~2 months, so it
practically never refreshes. A non-`errPortBusy` bind failure doesn't invalidate the file either, so
the loop replays the same stale payload every 5s indefinitely.

Adding `ServerName`/`Gateway` to the saved struct and rejecting a mismatch would cover it. Worth
noting this branch makes it much more likely, since `selectWireguardServer` re-picks from PIA's live
list on every connect, so "same region, different server" becomes normal rather than rare.

## 5. Smaller things

- `findAPIIP` applies the OpenVPN gateway octet-rewrite (`x.y.128.1`, `x.y.0.1`) to the WireGuard
  `server_vip`. For PIA WireGuard the PF gateway *is* the VIP (their own `port_forwarding.sh` uses
  `PF_GATEWAY=$WG_SERVER_VIP`), so the VIP itself is never probed. The fixtures in this PR use
  `server_vip: 10.13.161.1`, which I think its own code wouldn't reach.
- `internal/provider/privateinternetaccess/updater/wireguard_test.go:152` doesn't parse
  (`missing ',' before newline in argument list`), so that package fails `[setup failed]`, which
  also means `Test_selectWireguardServer` isn't running.
- `internal/vpn/wireguard_test.go` uses `new("...")`/`new(uint32(...))`, which needs go1.26 while
  `go.mod` declares 1.25, so `go vet ./...` exits 1.
- `gofmt -l` flags a few of the changed files.

Thanks for putting this together. Happy to re-test any revision.

m4r1k:
Hey there!

I reviewed the current head of this PR and prepared a follow-up branch against `neilcorp2kx:feat/pia-wireguard-native`, PR at https://github.com/neilcorp2kx/gluetun/pull/1

The follow-up keeps this PR as the implementation base, but changes the bootstrap and reconnect path to use the `internal/restrictednet` package that has since merged into master. It also addresses the failures reported earlier:
- removes the custom `SO_MARK`/marked-iptables bootstrap implementation
- resolves the PIA server list and opens the server-list, token, and `addKey` HTTPS connections through source-port-scoped restricted networking
- fixes the port-forwarding stop transition so `VPNIsUp=false` is validated after applying the new state and the keepalive service is actually stopped
- cycles through all matching live WireGuard servers rather than remaining on the first API result
- tries the `addKey` `server_vip` directly before the legacy PIA gateway address rewrites
- rejects static private keys, pre-shared keys, and interface addresses for PIA's dynamic registration, and disables unsupported IPv6 routes
- synchronizes the work with current master, repairs the broken updater test syntax and Go-version mismatch, and removes the unrelated global healthcheck retry change

The saved port-forward payload does not need a server identity. PIA documents payload/signature reuse across servers. The authenticated reconnect test confirmed this behavior: the first Vancouver endpoint issued port `44241`; after deleting `tun0`, gluetun stopped port forwarding, registered against a different Vancouver endpoint, reused port `44241`, and the port was externally reachable again.

Live verification covered the original recovery failure directly:
1. Native PIA WireGuard connected in kernelspace with the firewall enabled.
2. PIA egress and the forwarded port were verified from outside the tunnel container.
3. `tun0` was deliberately deleted.
4. The healthcheck failed as expected and triggered a restart.
5. Port forwarding stopped without the previous `server name not set` error.
6. Server discovery, token retrieval, and `addKey` succeeded while the old tunnel was down.
7. A fresh tunnel came up on a different endpoint, public egress returned, and the same forwarded port was externally reachable.
8. No temporary bootstrap rules remained in `iptables OUTPUT` after either connection.

Lint, generated-mock checks, cross-compilation, focused race tests, restrictednet integration tests, and the final image build pass. The full test-container run has the same netlink/PMTUD failures on current master and the follow-up branch on this SELinux host; audit logs show denied module auto-load requests for `rtnl-link-wireguard` and `ipt_mark`.

I also ran the traditional PIA OpenVPN path with the same image. OpenVPN initialized normally, public egress worked, port forwarding produced an externally reachable TCP port, and the firewall contained only the expected OpenVPN endpoint and tunnel rules.

AI was used for analysis and code generation. I reviewed the resulting changes and verification output.

Please @neilcorp2kx take a look at the improvments made, likewise @qdm12 hopefully we can get this in before 3.42 🤞

neilcorp2kx:
Thanks @m4r1k. I have reviewed the follow-up and am going to test the updated branch against a live PIA account.

The test plan is:

- controlled recovery runs where tun0 is deliberately deleted, followed by verification that the tunnel, PIA egress, port forwarding, saved-port reuse, and qBittorrent port synchronization all recover
- a continuous seven-day soak, since the earlier reports appeared after roughly 24 hours and six days
- CA Vancouver first, with France included so we cover the region from the original report

I will post the controlled recovery results within 24 hours of deploying the candidate and the final soak result no later than September 1, 2026. I will hold off merging the follow-up or marking this upstream PR ready until those results are in.

neilcorp2kx:
Controlled recovery testing is complete on the follow-up candidate at `831255decc9195aa3bb6a946501cb503a24ddcf7`.

Test setup: Docker on Linux, native PIA WireGuard, PIA port forwarding enabled, and qBittorrent sharing Gluetun's network namespace. For each cycle I deleted `tun0` while leaving the containers running, then allowed Gluetun to detect and recover without manual intervention.

Results:

- CA Vancouver: 3/3 recoveries, in 37s, 48s, and 25s.
- France: 3/3 recoveries, in 25s, 40s, and 42s.
- In all 6 cycles, `tun0` returned, the PIA port-forward lease rebound, and qBittorrent's listen port matched the forwarded port.
- A TCP probe from outside Gluetun's network namespace was reachable after a 5-second settling interval in all 6 cycles.
- One Vancouver probe taken immediately when the port-forward callback completed failed, then passed 5 seconds later. This looks like a brief post-rebind/socket settling window rather than persistent loss.
- The previous `server name not set` error did not recur during port-forward shutdown in any cycle.

The stack is back on CA Vancouver and healthy on the same pinned candidate. The 7-day soak began at 2026-08-24 22:28:33 UTC and runs through 2026-08-31 22:28:33 UTC. I will post the final soak result by September 1. I am still holding off on merge until that completes.

m4r1k:
Hey @neilcorp2kx,

Apologies but I fear a week of testing will further skew this PR from main and it will constantly require rebase it, would you accept a shorter testing? @qdm12 what's your take here?

qdm12:
I'm mostly AFK for the next 2 weeks so there shouldn't be any commits to master. This PR will be reviewed and merged post v3.42.0 which should be (finally) done in mid September (I think there is one pending fix PR left for it). So take your time and don't worry about rebasing. I'll try to review it on the go soon. Thanks !

m4r1k:
Thanks @qdm12, last rebase was massive. if dev slows down a bit, we can get this properly tested. Finger crossed then for the testing.

neilcorp2kx:
Thank you @m4r1k and @qdm12. I'm looking forward to getting this added! I'll keep you both posted.

neilcorp2kx:
Final runtime test result: **PASS**

Tested candidate: `831255decc9195aa3bb6a946501cb503a24ddcf7`

The candidate ran for 7 days and 1 hour with qBittorrent sharing Gluetun's network namespace and providing normal workload traffic.

Controlled recovery testing:

- CA Vancouver: 3/3 forced `tun0` recovery cycles passed in 37s, 48s, and 25s.
- France: 3/3 passed in 25s, 40s, and 42s.
- Every cycle restored WireGuard, rebound the PIA forwarded port, and synchronized qBittorrent.
- External TCP reachability returned after a 5-second settling interval in every cycle.

The soak also produced one natural health-check failure on August 31. Gluetun recovered without container intervention:

- VPN restart began at 04:11:41 MDT.
- WireGuard setup completed at 04:11:43.
- PIA port forwarding and qBittorrent resynchronization completed at 04:11:44.
- Total recovery time was approximately 3 seconds.

VM310 rebooted once during the test. Docker brought the same candidate container back about 10 seconds after shutdown, and WireGuard, port forwarding, and qBittorrent initialized normally.

Final state:

- Gluetun healthy with Canadian egress.
- qBittorrent still routed through the candidate.
- Forwarded port synchronized and externally reachable.
- Zero logged errors.
- No recurrence of `server name not set`.
- Twenty Cloudflare DNS-over-TLS connection resets occurred. Gluetun renewed each connection automatically, and DNS remained operational.

From a runtime-testing perspective, I consider this candidate passed. I recommend integrating the follow-up changes, updating the branch against current `master`, and rerunning CI. Another week-long soak should only be necessary if conflict resolution changes the PIA, WireGuard, firewall, or port-forwarding code.

neilcorp2kx:
Update after synchronizing this branch with current `master`:

- Merged the follow-up fixes from neilcorp2kx/gluetun#1.
- Merged upstream `master` at `bc5e219b` and resolved the resulting PIA validation incompatibility. Dynamic PIA WireGuard still receives its interface address at runtime, while user-supplied static PIA keys and addresses remain rejected.
- The PR head is now `d8b12511c2d56a0f6e1a7a27ef7f78afe7169b67`, and GitHub reports it as mergeable with no conflicts.

Validation on the exact head commit:

- `golangci-lint`: 0 issues
- generated mocks check: passed
- `go build ./...`: passed
- restricted-network integration tests: passed
- Darwin and Windows cross-builds: passed
- `devrun` and `ci` module tests: passed
- race-enabled full package suite: every package passed except the pre-existing external `internal/dns/Test_leakCheck`. The same failure reproduces on clean upstream `master` at `bc5e219b`; `ipleak.net` now returns `ip` as an array where the current test expects a map.

Runtime validation on VM310 using an image built from `d8b12511`:

- CA Vancouver startup: healthy; PIA WireGuard, public egress, port forwarding, and qBittorrent port update all passed.
- Three forced `tun0` losses on CA Vancouver recovered completely in 38s, 57s, and 27s.
- France startup passed, followed by a forced `tun0` loss that recovered completely in 48s.
- Each recovery restored Gluetun health, Gluetun and qBittorrent VPN egress, the forwarded-port file, and the qBittorrent listen-port update.
- Restored CA Vancouver as the steady state; the stack is healthy on `d8b12511` with qBittorrent routed through the new Gluetun container.

This targeted post-sync validation supplements the earlier [seven-day soak report](https://github.com/passteque/gluetun/pull/3393#issuecomment-5486492263).

michgauz:
Can't wait to see it live. Thanks @neilcorp2kx!

neilcorp2kx:
Marking this ready for review. Since the draft phase: the marked-iptables bootstrap was replaced with the merged `internal/restrictednet` approach (per @qdm12's suggestion, via @m4r1k's follow-up), the tunnel-recovery and port-forward-refresh failures reported above are fixed and covered by the controlled recovery tests, and the branch completed a 7-day production soak (results posted 2026-09-01). Head is synced with current `master` and mergeable. Happy to rebase again if needed once v3.42.0 lands.

Harland70:
Re-tested my earlier report against the current head (`d8b12511c2d56a0f6e1a7a27ef7f78afe7169b67`),
since the ready-for-review note says the tunnel-recovery and port-forward-refresh failures I hit are
now fixed. Same setup as before: real PIA account, CA Toronto, `VPN_TYPE=wireguard`,
`PORT_FORWARD_ONLY=true`, image built locally from that commit. (Analysis done with AI assistance.
Every number below is from a run I did myself.)

Short version: the recovery failure I reported is genuinely gone, throughput is good, but port
forwarding does not recover on the first attempt.

## 1. The tunnel-recovery failure is fixed

My original report was that a killed tunnel never comes back, because `GetWireguardConnection`
resolves `serverlist.piaservers.net` on every connect while gluetun's DNS loop only starts on
tunnel-up. I could not reproduce that any more.

Killed the tunnel three times with `ip link set tun0 down`:

| run | recovered in | new endpoint |
|---|---|---|
| 1 | ~20s | 66.56.80.188 |
| 2 | 31s | 66.56.80.195 |
| 3 | 62s | 66.56.80.174 |

Each reconnect landed on a different PIA server, so the serverlist lookup is clearly succeeding
while the tunnel is down. That was the exact path that used to deadlock. This one looks solved.

## 2. Port forwarding does not recover on the first attempt

This is the part I would want another pair of eyes on. On run 2 the tunnel came back fine but port
forwarding failed, left the port file empty for about a minute, and only recovered after a second
full VPN restart triggered by the healthcheck.

```
11:54:59 INFO  [port forwarding] stopping
11:54:59 INFO  [port forwarding] clearing port file /tmp/gluetun/forwarded_port
11:55:00 INFO  [wireguard] Connecting to 66.56.80.195:1337
11:55:01 INFO  [port forwarding] starting
11:55:01 INFO  [port forwarding] Found saved forwarded port data for port 56186
11:55:07 ERROR [vpn] starting port forwarding service: port forwarding for the first time:
               binding port: Get "https://10.25.128.1:19999/bindPort?payload=...&signature=...":
               context deadline exceeded
11:56:01 WARN  [vpn] restarting VPN because it failed to pass the healthcheck: small periodic check:
               all check tries failed:
               attempt 1 (0ms): with DNS server 116.202.176.26:53: lookup github.com on
               127.0.0.1:53: write udp 172.17.0.3:33008->116.202.176.26:53: write: operation not permitted
11:56:03 INFO  [port forwarding] port forwarded is 56186
```

Two things stand out to me. The `bindPort` call to the PIA gateway times out after the tunnel is
back up, and the healthcheck's DNS write gets `operation not permitted`, which reads like the same
class of problem as my original report: a socket that is not permitted through the firewall during
the transition window, just now on a different path. It does self-heal, but it costs an extra VPN
restart cycle and roughly 64 seconds with no forwarded port.

I only caught this by reading the logs across all three runs. The end state looked completely clean
(`forwarded_port` back to 56186, container healthy), so a spot check after the fact would have
missed it.

## 3. Sustained throughput

My earlier soak ran over a week with zero failures, but it was essentially idle, about 10 MB a day
of health probes, so it never tested throughput. Fixing that gap this time. 29 MiB pulls from
`speed.cloudflare.com`, three through the tunnel and two direct, moments apart:

```
tunnel:  138.0 Mbps, 139.8 Mbps, 138.7 Mbps
direct:  184.5 Mbps, 192.5 Mbps
```

That is about 0.73 of the direct line, sustained, with no stalls. For comparison, my production
OpenVPN setup against the same provider and region has been measuring 0.00 to 0.07 of direct on the
same test today, which is what pushed me to look at WireGuard in the first place.

## Summary

- Tunnel recovery: fixed, could not reproduce my original failure.
- Port forwarding: recovers, but not on the first attempt, and there is a ~64s window with no port.
- Throughput: good, roughly 0.73 of direct with no stalls.

Happy to re-run any of this or test a fix for the port forwarding case.

m4r1k:
@qdm12, could you advise on the intended behavior here before we change anything?

Following [Harland70's report](https://github.com/passteque/gluetun/pull/3393#issuecomment-5618454276), I traced the initial port-forward failure. The PIA WireGuard tunnel recovered, but one `/bindPort` request timed out. Port forwarding did not retry until the later DNS health failure restarted the VPN. The other two forced tunnel recoveries succeeded on their first port-forward attempt.

The current port-forward loop appears to distinguish an initial start caused by `UpdateWith` from a failure of an already-running service:

```go
serviceRunError, err = l.service.Start(runCtx)
if updateReceived {
    if err != nil {
        err = fmt.Errorf("starting port forwarding service: %w", err)
    }
    updateResult <- err
} else if err != nil {
    l.logger.Errorf("starting port forwarding service: %s - retrying in %s", err, retryDelay)
    retryAfter = time.After(retryDelay)
}
```

[`internal/portforward/loop.go`](https://github.com/neilcorp2kx/gluetun/blob/d8b12511c2d56a0f6e1a7a27ef7f78afe7169b67/internal/portforward/loop.go#L131-L143)

The VPN caller logs that initial error and continues:

```go
err = l.startPortForwarding(data)
if err != nil {
    l.logger.Error(err.Error())
}
```

[`internal/vpn/tunnelup.go`](https://github.com/neilcorp2kx/gluetun/blob/d8b12511c2d56a0f6e1a7a27ef7f78afe7169b67/internal/vpn/tunnelup.go#L139-L142)

PIA retries the initial bind only for `errPortBusy`; a timeout is returned immediately:

```go
err = bindPort(ctx, client, gateway, payload, signature)
if err == nil {
    break
} else if !errors.Is(err, errPortBusy) {
    return nil, fmt.Errorf("binding port: %w", err)
}
```

[`internal/provider/privateinternetaccess/portforward.go`](https://github.com/neilcorp2kx/gluetun/blob/d8b12511c2d56a0f6e1a7a27ef7f78afe7169b67/internal/provider/privateinternetaccess/portforward.go#L84-L90)

Is the synchronous `UpdateWith` failure intentionally one-shot, while only failures from an already-running service use the generic retry loop? What's your take for the right fix here?

internal/configuration/settings/portforward.go
```diff
@@ -53,6 +53,10 @@ type PortForwarding struct {
 	Username string `json:"username"`
 	// Password is only used for Private Internet Access port forwarding.
 	Password string `json:"password"`
+	// ServerName is only used for Private Internet Access port forwarding
+	// TLS certificate verification. It overrides the server name obtained
+	// from the VPN connection when set.
+	ServerName string `json:"server_name"`
```
qdm12: make this a pointer, and it should default to the empty string.

internal/configuration/settings/portforward.go
```diff
@@ -236,6 +246,9 @@ func (p *PortForwarding) read(r *reader.Reader) (err error) {
 		return err
 	}
 
+	p.ServerName = r.String("VPN_PORT_FORWARDING_SERVER_NAME",
+		reader.ForceLowercase(false))
```
qdm12: i dont think we care about casing here, so remove the reader.ForceLowercase(false)

internal/configuration/settings/portforward_test.go
```diff
@@ -17,3 +19,18 @@ func Test_PortForwarding_String(t *testing.T) {
...
+		Environ: []string{"VPN_PORT_FORWARDING_SERVER_NAME=vancouver439"},
+	})
+	settingsReader := reader.New(reader.Settings{Sources: []reader.Source{source}})
+
+	var settings PortForwarding
+	err := settings.read(settingsReader)
+
+	assert.NoError(t, err)
+	assert.Equal(t, "vancouver439", settings.ServerName)
+}
```
qdm12: Remove this test, it feels unneeded.

internal/configuration/settings/portforward.go
qdm12: Why not use the `SERVER_NAMES` as it's currently done instead of using a new option?

cmd/gluetun/main.go
```diff
@@ -446,27 +451,51 @@ func _main(ctx context.Context, buildInfo models.BuildInformation,
...
+		defaultIPv4Interface := ""
+		for _, defaultRoute := range defaultRoutes {
+			if defaultRoute.Family == netlink.FamilyV4 {
+				defaultIPv4Interface = defaultRoute.NetInterface
+				break
+			}
+		}
+		if defaultIPv4Interface == "" {
+			return fmt.Errorf("creating restricted network client: IPv4 default interface is not found")
+		}
```
qdm12: use the NetInterface field from `defaultRoutes`

cmd/gluetun/main.go
```diff
@@ -446,27 +451,51 @@ func _main(ctx context.Context, buildInfo models.BuildInformation,
...
+			upstreamResolvers[i] = upstreamResolver
+		}
+		ipv6Supported := false
+		restrictedClient = restrictednet.New(restrictednet.Settings{
+			DefaultInterface:  defaultIPv4Interface,
+			IPv6Supported:     &ipv6Supported,
+			Firewall:          firewallConf,
+			UpstreamResolvers: upstreamResolvers,
+		})
+	}
```
qdm12: move this somewhere else. Maybe a `Setup` function in internal/restrictednet`?

internal/configuration/settings/portforward.go
qdm12: Is the server name somehow extractable from the hostname perhaps? Or through PIA's API?

internal/configuration/settings/vpn.go
```diff
@@ -72,6 +80,21 @@ func (v *VPN) Validate(filterChoicesGetter FilterChoicesGetter, ipv6Supported bo
 	return nil
 }
 
+func (v VPN) validatePIAWireguard() error {
+	switch {
+	case *v.OpenVPN.User == "":
+		return errors.New("username is empty: set OPENVPN_USER")
+	case *v.OpenVPN.Password == "":
+		return errors.New("password is empty: set OPENVPN_PASSWORD")
```
qdm12: maybe worth adding new option names since `OPENVPN_`* options are kind of misleading as names

internal/configuration/settings/vpn.go
```diff
@@ -72,6 +80,21 @@ func (v *VPN) Validate(filterChoicesGetter FilterChoicesGetter, ipv6Supported bo
...
+	case len(v.Provider.ServerSelection.Regions) == 0 &&
+		len(v.Provider.ServerSelection.Names) == 0 &&
+		len(v.Provider.ServerSelection.Hostnames) == 0:
+		return errors.New("server selection is empty: set SERVER_REGIONS, SERVER_NAMES or SERVER_HOSTNAMES")
```
qdm12: that shouldn't be an error, it would then default to pick a server at random

internal/configuration/settings/vpn.go
```diff
@@ -57,6 +59,12 @@ func (v *VPN) Validate(filterChoicesGetter FilterChoicesGetter, ipv6Supported bo
 			return fmt.Errorf("OpenVPN settings: %w", err)
 		}
 	case vpn.Wireguard:
+		if v.Provider.Name == providers.PrivateInternetAccess {
+			err = v.validatePIAWireguard()
+			if err != nil {
+				return fmt.Errorf("private internet access settings: %w", err)
+			}
+		}
```
qdm12: move this into the v.Wireguard.validate function

internal/configuration/settings/wireguard.go
```diff
@@ -60,17 +60,14 @@ var regexpInterfaceName = regexp.MustCompile(`^[a-zA-Z0-9_]+$`)
...
-	if *w.PrivateKey == "" {
-		return errors.New("private key is not set")
-	}
-	_, err = wgtypes.ParseKey(*w.PrivateKey)
-	if err != nil {
-		err = fmt.Errorf("private key is not valid: %w", err)
-		if vpnProvider == providers.Nordvpn &&
-			err.Error() == "wgtypes: incorrect key size: 48" {
-			err = fmt.Errorf("%w - you might be using your access token instead of the Wireguard private key", err)
+	dynamicPIAWireguard := vpnProvider == providers.PrivateInternetAccess && !amneziawg
```
qdm12: instead let's just have a validatePIA method with everything in there, and not have a `dynamicPIAWireguard ` bool flag. That would endup being a headache when we add more wireguard providers requiring API network to get keys

internal/configuration/settings/wireguard.go
```diff
@@ -138,6 +135,39 @@ func (w Wireguard) validate(vpnProvider string, ipv6Supported, amneziawg bool) (
...
 }
 
+func (w Wireguard) validatePIAStaticSettings() error {
+	switch {
+	case *w.PrivateKey != "":
+		return errors.New("private key must not be set for Private Internet Access")
+	case *w.PreSharedKey != "":
+		return errors.New("pre-shared key must not be set for Private Internet Access")
+	case len(w.Addresses) > 0:
+		return errors.New("interface addresses must not be set for Private Internet Access")
```
qdm12: remove the "for private internet access" suffix for each, and instead wrap the error at the caller with `fmt.Errorf("private internet access settings: %w", err)`

internal/configuration/settings/wireguard.go
```diff
@@ -138,6 +135,39 @@ func (w Wireguard) validate(vpnProvider string, ipv6Supported, amneziawg bool) (
...
+	if err == nil {
+		return nil
+	}
+	err = fmt.Errorf("private key is not valid: %w", err)
+	if vpnProvider == providers.Nordvpn &&
+		err.Error() == "wgtypes: incorrect key size: 48" {
+		err = fmt.Errorf("%w - you might be using your access token instead of the Wireguard private key", err)
+	}
+	return err
+}
```
qdm12: revert to how it was, given it's used only once, unless it's the linter complaining

internal/configuration/settings/wireguardselection.go
```diff
@@ -41,7 +41,7 @@ func (w WireguardSelection) validate(vpnProvider string) (err error) {
 	switch vpnProvider {
 	case providers.Airvpn, providers.Fastestvpn, providers.Ivpn,
 		providers.Mullvad, providers.Nordvpn, providers.Protonvpn,
-		providers.Surfshark, providers.Windscribe:
+		providers.PrivateInternetAccess, providers.Surfshark, providers.Windscribe:
```
qdm12: nit have it before Protonvpn just for the sake of alpha-sorted

internal/models/server.go
```diff
@@ -13,29 +13,30 @@ import (
...
+	Number           uint16       `json:"number,omitempty"`
+	ServerName       string       `json:"server_name,omitempty"`
+	Hostname         string       `json:"hostname,omitempty"`
+	TCP              bool         `json:"tcp,omitempty"`
+	UDP              bool         `json:"udp,omitempty"`
+	OvpnX509         string       `json:"x509,omitempty"`
+	RetroLoc         string       `json:"retroloc,omitempty"` // TODO remove in v4
+	MultiHop         bool         `json:"multihop,omitempty"`
+	WgPubKey         string       `json:"wgpubkey,omitempty"`
+	WireguardDynamic bool         `json:"wireguard_dynamic,omitempty"`
```
qdm12: i think that's the case for all wireguard servers and PIA, so probably no need for an extra field?

internal/models/wireguard.go
```diff
@@ -0,0 +1,12 @@
...
+import "net/netip"
+
+// WireguardConnection contains provider-generated Wireguard connection data.
+type WireguardConnection struct {
+	Connection Connection
+	PrivateKey string
+	Addresses  []netip.Prefix
+	DNSServers []netip.Addr
+	Gateway    netip.Addr
+}
```
qdm12: Ideally have this in the pia package if possible, since it's specific to pia

internal/models/wireguard.go
```diff
@@ -0,0 +1,12 @@
+package models
+
+import "net/netip"
+
+// WireguardConnection contains provider-generated Wireguard connection data.
+type WireguardConnection struct {
+	Connection Connection
+	PrivateKey string
+	Addresses  []netip.Prefix
+	DNSServers []netip.Addr
```
qdm12: we don't care about that one really

```suggestion
```

internal/portforward/service/settings.go
```diff
@@ -16,6 +17,7 @@ type Settings struct {
 	UpCommand      string
 	DownCommand    string
 	Interface      string // needed for PIA, PrivateVPN and ProtonVPN, tun0 for example
+	Gateway        netip.Addr
```
qdm12: can't this be found in the routing tables automatically (as it is currently)? Check the master branch there was a recent tiny 1 line change on that.

internal/provider/privateinternetaccess/updater/api.go
```diff
@@ -60,9 +74,8 @@ func fetchAPI(ctx context.Context, client *http.Client) (
 		return data, err
 	}
 
-	// remove key/signature at the bottom
-	i := bytes.IndexRune(b, '\n')
-	b = b[:i]
```
qdm12: sure, but let's revert unneeded change

internal/provider/privateinternetaccess/updater/api.go
```diff
@@ -34,7 +36,19 @@ func fetchAPI(ctx context.Context, client *http.Client) (
...
 	const url = "https://serverlist.piaservers.net/vpninfo/servers/v7"
+	return fetchAPIFromURL(ctx, client, url)
+}
 
+func fetchWireguardAPI(ctx context.Context, client *http.Client) (
+	data apiData, err error,
+) {
+	const url = "https://serverlist.piaservers.net/vpninfo/servers/v6"
+	return fetchAPIFromURL(ctx, client, url)
+}
```
qdm12: Why use both v6 and v7? Aren't the wireguard servers on v7?? If not, can't we just use v6 for everything? I would rather avoid having two api calls to update servers.

internal/provider/privateinternetaccess/updater/hosttoserver.go
```diff
@@ -9,17 +9,19 @@ import (
...
+	server, ok := nts[key]
 	if !ok {
 		change = true
-		server.VPN = vpn.OpenVPN
+		server.VPN = vpnType
 		server.ServerName = name
 		server.Hostname = hostname
 		server.Region = region
 		server.PortForward = portForward
+		server.WireguardDynamic = vpnType == vpn.Wireguard
```
qdm12: yeah this feels unneeded.

internal/provider/privateinternetaccess/wireguard.go
```diff
@@ -0,0 +1,214 @@
...
+const wireguardRegistrationPort uint16 = 1337
+
+// RegisterWireguard registers a fresh ephemeral key with the selected PIA
+// server and returns all provider-generated connection settings.
+func (p *Provider) RegisterWireguard(ctx context.Context, connection models.Connection,
+	username, password string, restrictedClient common.RestrictedClient,
+) (wireguardConnection models.WireguardConnection, err error) {
+	switch {
+	case restrictedClient == nil:
+		return wireguardConnection, errors.New("restricted network client is not set")
```
qdm12: remove, cannot happen

internal/storage/formatting.go
```diff
@@ -19,11 +19,7 @@ func noServerFoundError(selection settings.ServerSelection) (err error) {
 
 	messageParts = append(messageParts, "VPN "+selection.VPN)
 
-	protocol := constants.UDP
-	if selection.OpenVPN.Protocol == constants.TCP {
-		protocol = constants.TCP
-	}
-	messageParts = append(messageParts, "protocol "+protocol)
+	messageParts = append(messageParts, openVPNProtocolParts(selection)...)
```
qdm12: revert

internal/storage/formatting.go
```diff
@@ -113,10 +109,7 @@ func noServerFoundError(selection settings.ServerSelection) (err error) {
 		messageParts = append(messageParts, part)
 	}
 
-	if *selection.OpenVPN.PIAEncPreset != "" {
-		part := "encryption preset " + *selection.OpenVPN.PIAEncPreset
-		messageParts = append(messageParts, part)
-	}
+	messageParts = append(messageParts, openVPNEncryptionPresetParts(selection)...)
```
qdm12: revert to how it was, add modification to if statement if needed.

internal/storage/formatting.go
```diff
@@ -146,16 +139,38 @@ func noServerFoundError(selection settings.ServerSelection) (err error) {
 		messageParts = append(messageParts, "tor only")
 	}
 
+	messageParts = append(messageParts, targetIPParts(selection)...)
+
+	message := "for " + strings.Join(messageParts, "; ")
+
+	return fmt.Errorf("no server found: %s", message)
```
qdm12: revert, this is unneeded change

internal/vpn/portforward.go
```diff
@@ -28,6 +28,7 @@ func (l *Loop) startPortForwarding(data tunnelUpData) (err error) {
 		Service: service.Settings{
 			PortForwarder:  data.portForwarder,
 			Interface:      data.vpnIntf,
+			Gateway:        data.gateway,
```
qdm12: check master branch, this might be done already

internal/vpn/run.go
```diff
@@ -44,29 +46,35 @@ func (l *Loop) Run(ctx context.Context, done chan<- struct{}) {
...
 				enabled:   settings.Type != vpn.Wireguard || *settings.Wireguard.MTU == 0,
 				vpnType:   settings.Type,
 				network:   connection.Protocol,
+				ipv6:      l.isIPv6Used(settings),
 				icmpAddrs: settings.PMTUD.ICMPAddresses,
 				tcpAddrs:  settings.PMTUD.TCPAddresses,
 			},
 			serverIP:       connection.IP,
-			serverName:     connection.ServerName,
+			serverName:     serverName,
```
qdm12: why not just the `connection.ServerName`?

internal/vpn/wireguard.go
```diff
@@ -4,45 +4,112 @@ import (
...
 	settings settings.VPN, ipv6SupportLevel netlink.IPv6SupportLevel, logger wireguard.Logger) (
-	wireguarder *wireguard.Wireguard, connection models.Connection, err error,
+	wireguarder *wireguard.Wireguard, connection models.Connection,
+	gateway netip.Addr, err error,
 ) {
 	ipv6Internet := ipv6SupportLevel == netlink.IPv6Internet
-	connection, err = providerConf.GetConnection(settings.Provider.ServerSelection, ipv6Internet)
+	dynamicProvider, dynamic := providerConf.(provider.DynamicWireguardProvider)
+	if dynamic {
+		connection, err = dynamicProvider.GetWireguardConnection(ctx,
```
qdm12: rename GetWireguardConnection with GetAPIWireguardConnection to make it clear it's doing API calls

internal/vpn/wireguard.go
```diff
@@ -4,45 +4,112 @@ import (
...
+			return nil, models.Connection{}, netip.Addr{}, fmt.Errorf("parsing registered private key: %w", err)
+		}
+		logger.Info(fmt.Sprintf("PIA addKey Wireguard configuration: endpoint %s, peer public key %s, "+
+			"client public key %s, interface addresses %s",
+			wireguardSettings.Endpoint, shortWireguardKey(wireguardSettings.PublicKey),
+			shortWireguardKey(clientPrivateKey.PublicKey().String()),
+			wireguardAddresses(wireguardSettings.Addresses)))
+		if len(wireguardConnection.DNSServers) > 0 {
+			logger.Debugf("Wireguard provider DNS servers: %v", wireguardConnection.DNSServers)
+		}
```
qdm12: move this to a function

internal/vpn/wireguard.go
```diff
@@ -4,45 +4,112 @@ import (
...
 
-	wireguardSettings := buildWireguardSettings(connection, settings.Wireguard, ipv6SupportLevel.IsSupported())
+	var wireguardSettings wireguard.Settings
+	if dynamic {
+		wireguardConnection, err := dynamicProvider.RegisterWireguard(ctx, connection,
+			*settings.OpenVPN.User, *settings.OpenVPN.Password, restrictedClient)
+		if err != nil {
+			return nil, models.Connection{}, netip.Addr{}, fmt.Errorf("registering Wireguard connection: %w", err)
+		}
+		connection = wireguardConnection.Connection
```
qdm12:
```suggestion
		connection, err = dynamicProvider.RegisterWireguard(ctx, connection,
			*settings.OpenVPN.User, *settings.OpenVPN.Password, restrictedClient)
		if err != nil {
			return nil, models.Connection{}, netip.Addr{}, fmt.Errorf("registering Wireguard connection: %w", err)
		}
```

internal/vpn/wireguard.go
```diff
@@ -4,45 +4,112 @@ import (
...
+		gateway = wireguardConnection.Gateway
+		clientPrivateKey, err := wgtypes.ParseKey(wireguardSettings.PrivateKey)
+		if err != nil {
+			return nil, models.Connection{}, netip.Addr{}, fmt.Errorf("parsing registered private key: %w", err)
+		}
+		logger.Info(fmt.Sprintf("PIA addKey Wireguard configuration: endpoint %s, peer public key %s, "+
+			"client public key %s, interface addresses %s",
+			wireguardSettings.Endpoint, shortWireguardKey(wireguardSettings.PublicKey),
+			shortWireguardKey(clientPrivateKey.PublicKey().String()),
+			wireguardAddresses(wireguardSettings.Addresses)))
```
qdm12: Probably to remove before merge, that's just internal/confusing information to the end user.

internal/vpn/wireguard.go
qdm12: dns servers pushed should be ignored

internal/vpn/wireguard.go
```diff
@@ -4,45 +4,112 @@ import (
...
+	return buildWireguardSettings(registration.Connection, userSettings, piaIPv6Supported)
+}
+
+func shortWireguardKey(key string) string {
+	const visibleCharacters = 8
+	if len(key) <= 2*visibleCharacters {
+		return key
 	}
+	return key[:visibleCharacters] + "..." + key[len(key)-visibleCharacters:]
+}
```
qdm12: Why would you hide a public key 🤔 Anyway just remove.

PR #3409, opened 2026-07-26
fix(protonvpn): adopt reassigned forwarded ports instead of restarting port forwarding

### LINKED ISSUE #3408: Bug: ProtonVPN port rotation leaves port forwarding permanently at 0 and the client on a stale port
### Is this urgent?

Yes - port forwarding ends up permanently disabled, and the torrent client is left listening on a port that is not forwarded.

### Host OS

Linux (Kubernetes, containerd)

### CPU arch

x86_64

### VPN service provider

ProtonVPN

Kubernetes

v3.41.1

### What's the problem?

With ProtonVPN + WireGuard + `VPN_PORT_FORWARDING=on`, gluetun obtains a forwarded port, then the next NAT-PMP renewal returns a **different** port. gluetun logs that as an `ERROR`, clears the port file, restarts the loop, obtains the new port, writes the port file, and then **immediately clears it again**.

The steady state after that is:

- `GET /v1/portforward` returns `{"port":0}`
- `VPN_PORT_FORWARDING_UP_COMMAND` runs only for the *first* port and never for the replacement, so the downstream client keeps the stale port forever
- No inbound connections, because the port the client listens on is not the port that is actually forwarded

In my case this silently persisted for **7 days** before I noticed. The port-forwarding loop never recovered on its own, and there was no further log output at all after the sequence below.

### Logs

Immediately after a fresh container start (IPs redacted, ports verbatim):

```
12:49:35 INFO  [port forwarding] starting
12:49:35 INFO  [port forwarding] gateway external IPv4 address is <gateway-A>
12:49:35 INFO  [port forwarding] port forwarded is 37928
12:49:35 INFO  [port forwarding] writing port file /tmp/gluetun/forwarded_port
12:49:35 INFO  [port forwarding] Connecting to 127.0.0.1:8090... connected.
12:49:35 INFO  [port forwarding] HTTP request sent, awaiting response... 200 OK
12:50:20 INFO  [port forwarding] clearing port file /tmp/gluetun/forwarded_port
12:50:20 ERROR [port forwarding] external port changed: 37928 changed to 60557
12:50:20 INFO  [port forwarding] starting
12:50:21 INFO  [port forwarding] gateway external IPv4 address is <gateway-B>
12:50:21 INFO  [port forwarding] port forwarded is 60557
12:50:21 INFO  [port forwarding] writing port file /tmp/gluetun/forwarded_port
12:50:21 INFO  [port forwarding] clearing port file /tmp/gluetun/forwarded_port
```

Note that the reported gateway external IPv4 address is **not the same** between the two attempts, which appears to be what produces the different port on renewal.

Resulting state, ~10 minutes later, with no further log lines:

```
port file            : 37928
GET /v1/portforward  : {"port":0}
client listen_port   : 37928
last forwarded port  : 60557   (per the log above)
```

So three sources disagree, and the one the client is using is not the one that was last forwarded.

### Steps to reproduce

1. Run gluetun v3.41.1 with ProtonVPN over WireGuard, `VPN_PORT_FORWARDING=on`, `PORT_FORWARD_ONLY=on`, `SERVER_COUNTRIES=Netherlands`.
2. Set a `VPN_PORT_FORWARDING_UP_COMMAND` that pushes `{{PORTS}}` into a downstream client.
3. Wait for the first renewal (~45s here).
4. Observe the `external port changed` error, then check `GET /v1/portforward`.

### Configuration

```yaml
VPN_SERVICE_PROVIDER: protonvpn
VPN_TYPE: wireguard
SERVER_COUNTRIES: Netherlands
PORT_FORWARD_ONLY: "on"
VPN_PORT_FORWARDING: "on"
VPN_PORT_FORWARDING_UP_COMMAND: <pushes {{PORTS}} to the client's API>
FIREWALL_OUTBOUND_SUBNETS: <two RFC1918 ranges>
FIREWALL_INPUT_PORTS: <three WebUI ports>
```

### What I would expect

A changed port on renewal is a normal ProtonVPN behaviour, not a fatal condition. I would expect gluetun to:

1. Treat the new port as the current port rather than an error.
2. Keep the port file populated with the new port instead of clearing it as the last action.
3. Re-run `VPN_PORT_FORWARDING_UP_COMMAND` for the new port so the downstream client follows it.
4. Keep `/v1/portforward` consistent with the port file.

Happy to work on a PR if you can confirm the intended semantics - specifically whether the up-command is meant to re-run on every port change, and whether the clear-then-write ordering around a rotation is deliberate.

quadseven:
Fixes #3408.

## Why

The ProtonVPN gateway may hand back a **different external port** when an existing NAT-PMP mapping is refreshed. `addPortMappingTCPUDP` treated that as an error:

```go
} else if externalPort != 0 && externalPort != 1 && externalPort != assignedExternalPort {
    return 0, 0, fmt.Errorf("%s external port requested as %d but received %d", ...)
}
```

That error propagated out of `KeepPortForward`, so `Start`'s goroutine ran `cleanup()` and pushed the error to the loop, which tore down and rebuilt the whole port forwarding service. Observed effect on a live setup:

- the firewall rules and port file were dropped on every reassignment
- `VPN_PORT_FORWARDING_UP_COMMAND` had already run for the *previous* port and did not run again for the replacement, so the downstream client stayed on a port that was no longer forwarded
- `GET /v1/portforward` ended up reporting `0` while the port file and the client still held the stale port

Worth noting that the external-port check could only ever fire on the refresh path. `PortForward` calls `addPortMappingTCPUDP` with `externalPort` of `1` (first, symmetric mapping) or `0` (extra mappings), and both are excluded by the condition. So the check existed solely to reject the case this PR now handles.

Size: S

## What changed

- `utils.PortForwardObjects` gains an optional `OnPortsChanged` callback so a provider can hand newly assigned ports back to the port forwarding service.
- `Service.onPortsChanged` applies them in place: it calls the existing `cleanup()` to close the firewall for the previous ports, then the existing `onNewPorts()` to allow the new ones, rewrite the port file and re-run the up command. This is the same sequence `SetPortsForwarded` already uses, so no new firewall logic is introduced.
- `protonvpn.KeepPortForward` compares each refreshed mapping against the previous one, logs any reassignment, and calls `OnPortsChanged` once per refresh if anything moved.
- `addPortMappingTCPUDP` no longer rejects a changed external port; the caller now decides. The internal-port check and the TCP/UDP consistency check are unchanged.
- When `OnPortsChanged` is nil, `KeepPortForward` returns the new `ErrPortsReassigned` instead, preserving today's restart behaviour rather than silently forwarding a port the client is not listening on.

## Ordering and the firewall

`onPortsChanged` deliberately removes the old allowed port **before** adding the new one, so a reassignment cannot leave a hole open on a port that is no longer forwarded. `Test_Service_onPortsChanged` asserts that ordering with `gomock.InOrder`. Nothing outside the forwarded-port rules is touched, so the kill switch and outbound-subnet handling are unaffected.

`onPortsChanged` takes only `portMutex`, never `startStopMutex`. It is invoked from the `KeepPortForward` goroutine, and `Stop` holds `startStopMutex` while waiting for that goroutine to return, so taking it here would deadlock. This is noted in a comment on the method.

## Acceptance criteria

- A reassigned external port is adopted without restarting the port forwarding service.
- The firewall closes the previous port before opening the new one.
- The port file and `GET /v1/portforward` report the newly assigned port.
- The up command runs for the newly assigned port.
- A provider whose service does not supply the callback still returns an error.
- Every other qBittorrent-facing behaviour is unchanged when no reassignment occurs.

## Test plan

- `go build ./...` clean.
- `go vet ./internal/portforward/... ./internal/provider/...` clean.
- New `Test_Service_onPortsChanged` and `Test_Service_onPortsChanged_keepsPreviousPortsOnFailure`, covering the adopt path, the firewall ordering, the port file contents, and the failure path where the new port cannot be allowed.
- New `Test_portsMapToString` covering no-change, single reassignment, multiple ports, partial change and unknown internal port.
- `mockgen` directive extended from `Logger` to `Logger,PortAllower`; `mocks_test.go` regenerated with the same command.
- `go test ./internal/portforward/... ./internal/provider/...` all pass.

Two notes on the local run, for transparency:

- `internal/netlink`, `internal/tun` and `internal/natpmp` fail for me, but they fail identically on an unmodified checkout of `master` (verified 5/5 runs for `Test_Client_AddPortMapping/rpc_error`, which expects an i/o timeout and gets a connection refused). These look macOS-specific and unrelated.
- I could not add a unit test for the reassignment branch of `KeepPortForward` itself, because it constructs `natpmp.New()` internally and the client is not injectable. Happy to add one if you would like that refactored as part of this PR.

## Out of scope

- Why the gateway reassigns the port. In my case successive NAT-PMP requests reported different gateway external addresses, which suggests the endpoint is load balanced, but this PR only makes gluetun handle the reassignment correctly.
- The other providers' `KeepPortForward` implementations, which do not renew a port mapping this way.


qdm12:
closing this in favor of handling port forwarding port changes using the upper layers of the port forwarding loop

PR #3415, opened 2026-07-28
feat(ci): attach provenance and SBOM attestations to the published image

kobihikri:
Hi, and thanks for gluetun.

`.github/workflows/ci.yml` publishes the image, but the pushed manifest carries no provenance or SBOM attestation. Someone pulling it cannot check that it was built by this workflow, from this repository, at that tag.

gluetun is an unusually good case for this. People deploy it specifically to route other containers' traffic through a VPN — it is chosen *because* it sits in the trust path, and it is handed VPN credentials to do the job. An image whose entire purpose is to be trusted with your network egress is one worth being able to verify.

The change is two lines on the build step:

```yaml
          push: ...
          provenance: mode=max
          sbom: true
```

BuildKit attaches both to the image manifest, so they travel with the image. **No permissions change is needed** — nothing has to gain `id-token`.

```
docker buildx imagetools inspect <image>:<tag> --format '{{ json .Provenance }}'
```

Two caveats: `mode=max` records build arguments (`provenance: true` gives a smaller record if any have been sensitive), and attestations add an extra manifest to the index, which the registry supports.

No SLSA level claimed.

Disclosure: I used AI assistance to help spot this and prepare the change, and I read the workflow myself.


qdm12 (approved):
Awesome! I appreciate the

- PR description 😺 
- honesty with AI usage 🏅 
- reviewing what AI produced 🙏 

And obviously the idea, I've been wanting to do that for a while and never found the motivation to address it, little did I know it would be that simple (configuration wise)! So thanks for that 💯!!!

kobihikri:
**Correction to my PR description — after you already merged it, which makes it more important rather than less.**

I wrote that the pushed manifest "carries no provenance or SBOM attestation". Half of that was wrong.

**Provenance was already there.** For public repositories `docker/build-push-action` adds provenance with `mode=max` by default — [Docker's docs](https://docs.docker.com/build/ci/github-actions/attestations/): *"Public repos: provenance attestations with `mode=max` are automatically added"*. So the `provenance: mode=max` line made existing behaviour explicit rather than adding it.

**The SBOM was genuinely new** — *"SBOM attestations aren't automatically added to the image"* — so that half of the change did what I said.

I also claimed in the caveats that `provenance: true` gives "a smaller record". Also wrong: `true` resolves to max on a public repo; the smaller setting is `provenance: mode=min`.

Nothing about the merged change is harmful — an explicit provenance line is still worth having, because the behaviour will not shift quietly if the default ever changes, and the SBOM is a real addition. But you merged on a description that overstated what it did, and you should know that.

Apologies. It was caught by a maintainer reviewing the same change on another project.

qdm12:
No problem, awesome! Good to know too and thanks docker for default enabling provenance I guess 🤷

Issue #279, opened 2020-10-30
Feature request: Export metrics to Prometheus


qdm12:
1. What's the feature?

- Export metrics to Prometheus on:
    - Scrap iptables data to get network information (blocked data, vpn'ed data etc.)
    - Health information, such as when connectivity drops
    - VPN server information
    - the Go written Shadowsocks
    - the (Go) HTTP proxy
    - the Go written DNS server (blocked by #137)
    - ~the numbers found from #251~
    - Maybe add some bandwidth tests but that depends on the test server, so maybe not
- ~Add instructions on how to set this up with Grafana to have nice graphs~ See https://github.com/qdm12/gluetun-wiki/blob/latest/setup/advanced/metrics.md
- ~Add a Grafana dashboard example to the repository~ See https://github.com/passteque/gluetun/blob/master/metrics/grafana.json

2. Why do you need this feature?

Because metrics are fun

3. Extra information?

dorsegal:
That would be awesome!

SR-G:
So any news on this ?
I would also find this very helpful. Especially the "connection lost", "connected country", ... informations.

Maybe "start small" / do "baby steps" regarding this new evolution, as the feature list is quite long and may be overhelming ?

What would be the "architecture" for these metrics ? Something embedded inside gluetun code itself (and exposed by ... the control server, on a /metrics path, maybe ?), or an external separated program (a `gluetun-exporter` go standalone project, able to externally gather (in some way) all the metrics) ?

TheGameProfi:
Hey everyone 👋

Since there hasn’t been much activity or communication lately, I went ahead and created a Prometheus Exporter myself.
It’s still a work in progress — I’m currently experimenting with getting it to auto-configure, but the core functionality is already there.

Feel free to check it out and let me know what you think:
👉 [Gluetun-Exporter](https://github.com/thecfu/gluetun-exporter)

I’m totally open to feedback, ideas, or suggestions! 😊

SR-G:
Interesting, thanks !

But at very quick glance i'm not so sure how it has to be launched / what is the principe (the README.md is not 100% clear about that), especially with Standalone mode.

I build the container i executed it with : 

```
docker run --rm --name "gluetun-exporter" \
  -e"GLUETUN_HOST=localhost" \ <-- same with "gluetun"
  -e"GLUETUN_PORT=8000" \
  -e"EXPORTER_PORT=8001" \
  -e"EXPORTER_ROLE=gluetun" \
[...]

PR #3421, opened 2026-08-01
feat(metrics): `gluetun_tun_rx_bytes_total` and `gluetun_tun_tx_bytes_total` metrics

### LINKED ISSUE #251: Feature request: vnstat like traffic monitor
1. What's the feature?
Ability to find out how much data has been sent/received through the vpn tunnel and keeping track of daily/monthly totals.

2. Why do you need this feature?
Limited data plans from ISP

3. Extra information?
Maybe there's already a way to find this out without adding a new feature?

sugata100:
# Description

# Issue

Fixes #251 

Refers to #279 

# Assertions


qdm12:
Nice, but let's use 0fef7b24ee62d7423296d1b1824d2dd97eab4ce6 which sets up basic code to have a Prometheus server (no actual metrics yet), so that these metrics can be part of the Prometheus server data instead of the control server.

cmd/gluetun/main.go
```diff
@@ -471,6 +472,11 @@ func _main(ctx context.Context, buildInfo models.BuildInformation,
 		"vpn", goroutine.OptionTimeout(time.Second))
 	go vpnLooper.Run(vpnCtx, vpnDone)
 
+	err = tunstats.New(prometheusRegistry, vpnLooper)
```
qdm12, line 475: maybe create a metrics gatherer from the `internal/metrics` package, which would itself register a sub-gatherer from the tunstats package? That way we don't need to call a bunch of `.New` constructors for each metrics subpackages like `tunstats`

internal/metrics/tunstats/collector.go
```diff
@@ -0,0 +1,101 @@
+// Package tunstats offers a New function returning a Prometheus
+// collector that exposes the cumulative number of bytes received and
+// sent over the VPN network interface (TUN for OpenVPN, and the
+// Wireguard interface for Wireguard and AmneziaWG). The interface name
```
qdm12, line 4: remove the parenthese, it's a tun interface for all protocols.

internal/metrics/tunstats/sysfs.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+)
+
+const defaultSysfsNetPath = "/sys/class/net"
```
qdm12, line 11: define this constant in the New constructor, no need for global scope

internal/metrics/tunstats/collector.go
```diff
@@ -0,0 +1,101 @@
...
+
+// Collector exposes the VPN network interface received and sent byte
+// counters as Prometheus metrics.
+type Collector struct {
+	descRxBytes *prometheus.Desc
+	descTxBytes *prometheus.Desc
+	// getVPNSettings returns the current VPN settings.
+	getVPNSettings func() settings.VPN
+	// sysfsNetPath is the sysfs base path containing the network
+	// interfaces. It defaults to /sys/class/net.
```
qdm12, line 38: not in tests so lets remove

```suggestion
	// interfaces.
```

internal/metrics/tunstats/collector.go
```diff
@@ -0,0 +1,101 @@
...
+// statistics cannot be read (for example when the VPN is not
+// connected).
+func (c *Collector) Collect(ch chan<- prometheus.Metric) {
+	interfaceName := resolveInterfaceName(c.getVPNSettings())
+	if interfaceName == "" {
+		return
+	}
+	rxBytes, txBytes, err := c.readInterfaceStats(interfaceName)
+	if err != nil {
+		return
```
qdm12, line 79: we should log a warning here

internal/metrics/tunstats/collector.go
```diff
@@ -0,0 +1,101 @@
...
+func resolveInterfaceName(vpnSettings settings.VPN) (interfaceName string) {
+	switch vpnSettings.Type {
+	case vpn.OpenVPN:
+		interfaceName = vpnSettings.OpenVPN.Interface
+	case vpn.Wireguard:
+		interfaceName = vpnSettings.Wireguard.Interface
+	case vpn.AmneziaWg:
+		interfaceName = vpnSettings.AmneziaWg.Wireguard.Interface
+	}
+	return interfaceName
```
qdm12, line 100: panic on default, the vpn settings should enforce the type is one of these 3.

internal/metrics/tunstats/collector.go
```diff
@@ -0,0 +1,101 @@
...
+
+// Collect implements the prometheus.Collector interface. It emits
+// nothing if the VPN network interface cannot be resolved, or its
+// statistics cannot be read (for example when the VPN is not
+// connected).
+func (c *Collector) Collect(ch chan<- prometheus.Metric) {
+	interfaceName := resolveInterfaceName(c.getVPNSettings())
+	if interfaceName == "" {
+		return
+	}
```
qdm12, line 76: remove this should never happen

internal/metrics/tunstats/collector.go
```diff
@@ -0,0 +1,101 @@
...
+	ch <- c.descRxBytes
+	ch <- c.descTxBytes
+}
+
+// Collect implements the prometheus.Collector interface. It emits
+// nothing if the VPN network interface cannot be resolved, or its
+// statistics cannot be read (for example when the VPN is not
+// connected).
+func (c *Collector) Collect(ch chan<- prometheus.Metric) {
+	interfaceName := resolveInterfaceName(c.getVPNSettings())
```
qdm12, line 73: we might want to use the netlink package to double check the link exists. Sometimes if the VPN is starting/restarting, the link would be deleted temporarily. If it does not exist, then we can write 0 to the metrics and not log a warning.

internal/metrics/tunstats/sysfs.go
```diff
@@ -0,0 +1,52 @@
...
+	rxBytes, txBytes uint64, err error,
+) {
+	rxBytes, err = c.readSysfsCounter(c.counterPath(interfaceName, counterRxBytes))
+	if err != nil {
+		return 0, 0, fmt.Errorf("reading %s: %w", counterRxBytes, err)
+	}
+	txBytes, err = c.readSysfsCounter(c.counterPath(interfaceName, counterTxBytes))
+	if err != nil {
+		return 0, 0, fmt.Errorf("reading %s: %w", counterTxBytes, err)
+	}
```
qdm12, line 30: that's kind of inefficient to read the file twice, let's just read the file once and then parse it fully

internal/metrics/tunstats/collector_test.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+}
+
+func Test_New_Gather_NoMetrics(t *testing.T) {
+	t.Parallel()
+
+	testCases := map[string]struct {
+		vpnSettings settings.VPN
+	}{
+		"interface absent": {
```
qdm12, line 136: for all subtests names, use udnerscores instead of spaces so `interface_absent` here

qdm12 (commented):
Self review / ignore these comments - I'm working on this branch and pushing to it.

PR #3424, opened 2026-08-03
feat(wireguard): `WIREGUARD_GSO` option to disable TUN vnet hdr batching

postmaxin:
# Description

Since wireguard-go introduced vectorized TUN I/O, it unconditionally opens the TUN device with `IFF_VNET_HDR` and uses GRO/GSO batched reads and writes, with no way to opt out. Some kernels misbehave on this path: notably vendor NAS kernels such as ASUSTOR ADM (Linux "6.6.x" — the vendor scrubs the exact version string), which advertise `IFF_VNET_HDR` support but reject GRO-coalesced writes with `EINVAL`. When that happens, wireguard-go repeatedly logs

```
Failed to write packets to TUN device: write /dev/net/tun: invalid argument
```

under load and ends up retrying packets individually, wasting up to half of the available bandwidth. The breakage is undetectable at TUN creation time: the kernel accepts the flags and only fails on write.

This PR adds `WIREGUARD_DISABLE_GSO` (default `off`, no behavior change unless opted in). When `on`, the TUN device is created without `IFF_VNET_HDR`, so wireguard-go detects no vnet hdr support and falls back to simple single-packet reads and writes — the same behavior all versions had before vectorized I/O. The AmneziaWG userspace path honors the same option.

## Testing

- Running in production on an ASUSTOR NAS (the affected platform) with ProtonVPN WireGuard. In an A/B test on the same host and mirror, stock gluetun (GSO enabled) sustained ~50 Mbit/s while this build with `WIREGUARD_DISABLE_GSO=on` sustained ~82 Mbit/s (indicative only — VPN server assignment may differ between connections).
- The `EINVAL` log spam that motivated this change (spring 2026) no longer reproduces on the current ADM kernel, suggesting ASUSTOR has since shipped the mainline fix — but vendor kernels can carry the regression indefinitely, and the same regression hit mainline LTS (~6.6.44) for a window: kernel bisection documented in tailscale/tailscale#13041, which shows the identical symptom in Tailscale's embedded wireguard-go.
- `go build ./...`, `golangci-lint run` (v2.4.0, 0 issues) and `go test ./...` pass; new table-test case covers `DisableGSO=true` and the default-MTU fallback in `buildWireguardSettings`.

## A note on how this was developed

This was developed and iterated with substantial help from Anthropic Claude (implementation, AGENTS.md conformance pass, and testing against a live ASUSTOR deployment), with all changes reviewed and validated by a human. Happy to walk through any part of the design.

# Issue

No matching gluetun issue found (searched for the error string, GSO, vnet_hdr, ASUSTOR). Closest ecosystem reference: https://github.com/tailscale/tailscale/issues/13041 — same wireguard-go error and slowdown, bisected to a mainline kernel regression in the `virtio_net_hdr` GSO path.

# Assertions

* [x] I am aware that any changes to settings should be reflected in the [wiki](https://github.com/qdm12/gluetun-wiki/) — happy to submit a wiki PR for `WIREGUARD_DISABLE_GSO` once this is reviewed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


internal/configuration/settings/wireguard.go
```diff
@@ -46,6 +46,13 @@ type Wireguard struct {
 	// It defaults to "auto" and cannot be the empty string
 	// in the internal state.
 	Implementation string `json:"implementation"`
+	// DisableGSO creates the WireGuard TUN device without IFF_VNET_HDR,
+	// preventing wireguard-go from enabling its GRO/GSO batch-write path.
+	// This is needed on some kernels (e.g. certain NAS devices) that claim
+	// IFF_VNET_HDR support but return EINVAL when wireguard-go writes
+	// GRO-coalesced packets with virtio_net_hdr structs under load.
+	// It defaults to false and cannot be nil in the internal state.
+	DisableGSO *bool `json:"disable_gso"`
```
qdm12: can we flip this to GSO which defaults to true/on? I just prefer something on/off versus something_disable on/off

internal/configuration/settings/wireguard.go
```diff
@@ -46,6 +46,13 @@ type Wireguard struct {
 	// It defaults to "auto" and cannot be the empty string
 	// in the internal state.
 	Implementation string `json:"implementation"`
+	// DisableGSO creates the WireGuard TUN device without IFF_VNET_HDR,
+	// preventing wireguard-go from enabling its GRO/GSO batch-write path.
+	// This is needed on some kernels (e.g. certain NAS devices) that claim
+	// IFF_VNET_HDR support but return EINVAL when wireguard-go writes
+	// GRO-coalesced packets with virtio_net_hdr structs under load.
```
qdm12: would it be possible to log a warning when encountering the tun error you got, to suggest turning gso off? I guess that would involve a custom logger injected to wireguard-go which would compare each log line, but I havent digged where the error you got is coming from

internal/configuration/settings/wireguard.go
```diff
@@ -265,5 +279,11 @@ func (w *Wireguard) read(r *reader.Reader, amneziaWG bool) (err error) {
 	if err != nil {
 		return err
 	}
+
+	w.DisableGSO, err = r.BoolPtr("WIREGUARD_DISABLE_GSO")
```
qdm12: flip to WIREGUARD_GSO

internal/wireguard/settings.go
```diff
@@ -48,6 +48,11 @@ type Settings struct {
 	// Implementation is the implementation to use.
 	// It can be auto, kernelspace or userspace, and defaults to auto.
 	Implementation string
+	// DisableGSO creates the TUN device without IFF_VNET_HDR so that
+	// wireguard-go falls back to single-packet reads and writes instead
+	// of its GRO/GSO batch I/O path.
+	// See WIREGUARD_DISABLE_GSO for details.
```
qdm12: flip to GSO

internal/wireguard/run.go
```diff
@@ -147,7 +147,7 @@ func Run(ctx context.Context, waitError chan<- error, ready chan<- struct{},
 
 func setupKernelSpace(ctx context.Context,
 	interfaceName string, netLinker NetLinker, mtu uint32,
-	cleanups *cleanup.Cleanups, logger Logger) (
+	_ bool, cleanups *cleanup.Cleanups, logger Logger) (
```
qdm12, line 150: do you have wireguard in your kernel on your odd system? I guess not but curious!

internal/wireguard/offload_linux.go
```diff
@@ -0,0 +1,49 @@
...
+	ifr.SetUint16(unix.IFF_TUN | unix.IFF_NO_PI) // intentionally omit IFF_VNET_HDR
+	if err := unix.IoctlIfreq(tunFD, unix.TUNSETIFF, ifr); err != nil {
+		unix.Close(tunFD)
+		return nil, fmt.Errorf("setting TUN flags: %w", err)
+	}
+	if err := unix.SetNonblock(tunFD, true); err != nil {
+		unix.Close(tunFD)
+		return nil, fmt.Errorf("setting nonblock: %w", err)
+	}
+	return os.NewFile(uintptr(tunFD), "/dev/net/tun"), nil
```
qdm12, line 52: wrap the error with `creating tun fd file: %w`

qdm12 (commented):
Nice PR 💪 
I appreciate the AI+Human effort 💯 
and the fully/partly human written PR description too!

What a relief in this day and age 😸

internal/wireguard/run.go
postmaxin: It is not supported on ASUSTOR 5.1, hence the awful log spew and bandwidth loss.

qdm12 (dismissed):
Awesome! 👍

qdm12:
Sorry to be annoying on the x-compilation compatibility, can you fix:

```
 > [xcompile 2/2] RUN GOOS=windows go build -o /dev/null ./...:
18.40 # github.com/qdm12/gluetun/internal/amneziawg
18.40 internal/amneziawg/run.go:45:25: undefined: amneziatun.CreateTUNFromFile
```

Thanks!

qdm12:
merged, thanks! 👍

PR #3429, opened 2026-08-12
Fix broken star history chart in README

OctoBored:
The star history chart in the README is currently broken due to GitHub's stargazer API restrictions. This changes the chart to use star-history.dera.page, an alternative that relies on a different data source requiring no API token, restoring the chart for all visitors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ services:
 
 ## Fun graphs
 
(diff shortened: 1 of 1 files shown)


qdm12 (approved):
thankssss!

PR #3464, opened 2026-09-08
fix(natpmp): retry on connection refused errors, like timeouts - Fix #3462

### LINKED ISSUE #3462: Bug: one refused NAT-PMP datagram tears down ProtonVPN port forwarding, ~11 times per hour
### Is this urgent?

No, it always recovers on its own. But it's cumbersome.

### Host OS

Ubuntu 24.04 · x86_64 · docker compose

### VPN service provider

ProtonVPN (WireGuard, NAT-PMP-enabled key, P2P server)

`v3.41.3` — follow-up to #3418, which was closed with "This should be fixed in `v3.41.3`". The
mapping does come back now, so that part is fixed. What remains is that a transient failure still
triggers a full teardown.

### What's the problem?

`KeepPortForward` refreshes the mapping every 45 s for a 60 s lease. When the gateway refuses a
single datagram (`recvfrom: connection refused`), the whole port forwarding service is torn down
and rebuilt: port file cleared, firewall rule dropped, external address re-queried, new mapping
requested, `VPN_PORT_FORWARDING_UP_COMMAND` re-run.

Proton then hands back **the same port** — nothing had actually been lost.

Measured over 10 h 52 on a single uninterrupted tunnel (no reconnection, healthcheck `healthy`
throughout, same forwarded port from start to finish):

| | |
|---|---|
| teardown/rebuild cycles | **121** |
| shortest / median / longest | **5 s / 27 s / 253 s** |
| total time with no forwarded port | **12.5 %** |
| times the port actually changed | **0** |

So `GET /v1/portforward` reports no port for minutes at a time, and the up command runs ~11 times
an hour with an unchanged port — which makes a torrent client rebind its listening socket just as
often, dropping incoming peer connections each time. Switching ProtonVPN server lowers the rate but
does not remove it.

### Root cause

`internal/natpmp/rpc.go` retries only on timeouts:

```go
var netErr net.Error
if errors.As(err, &netErr) && netErr.Timeout() {
    connectionDuration *= 2
    continue
}
return nil, fmt.Errorf("reading from udp connection: %w", err)
```

`initialConnectionDuration = 250ms` / `maxTries = 9` is RFC 6886 §3.1 exactly — the RFC anticipates
a **silent** gateway (lost datagram), not one that **refuses**. On Linux an ICMP port-unreachable
surfaces as `ECONNREFUSED`, which is not a timeout, so the loop gives up immediately. The error then
propagates out of `KeepPortForward`, and `portforward/service/start.go` treats that as a crash →
`cleanup()` → restart. Both the retry branch and the `return` date back to the original NAT-PMP
support (#1543); the escalation to a full restart came with the run loop rework (#1874).

### Suggested fix

Treat `ECONNREFUSED` as retryable in `rpc.go`, like a timeout. An ICMP port-unreachable does not
prove the service is gone, only that nothing was listening *at that instant*. This covers every
NAT-PMP provider and keeps the hiccup invisible to the layers above.

Alternatively, retry the refresh in `KeepPortForward` before escalating — there are 15 s of
remaining lease to work with — or skip `cleanup()` and the up/down commands when the rebuild yields
the same port. #3409 already established "adopt instead of restart" for a *reassigned* port; this
asks for the same where nothing changed at all.

### Logs

One cycle, trimmed (gateway IP redacted; last line is my own up command):

```
22:18:22 INFO  [port forwarding] clearing port file /gluetun/forwarded_port
22:18:22 ERROR [port forwarding] adding port mapping: executing remote procedure call: reading from udp connection: read udp 10.2.0.2:37153->10.2.0.1:5351: recvfrom: connection refused
22:18:22 INFO  [port forwarding] starting
22:18:22 ERROR [port forwarding] starting port forwarding service: port forwarding for the first time: getting external IPv4 address: ... recvfrom: connection refused - make sure you have +pmp at the end of your OpenVPN username - retrying in 5s
22:18:27 INFO  [port forwarding] starting
22:18:28 ERROR [port forwarding] ... recvfrom: connection refused - retrying in 5s
22:18:33 INFO  [port forwarding] starting
22:18:41 INFO  [port forwarding] gateway external IPv4 address is <redacted>
22:18:49 INFO  [port forwarding] port forwarded is 46640
22:18:49 INFO  [port forwarding] writing port file /gluetun/forwarded_port
22:18:49 INFO  [port forwarding] up command: listening port set to 46640
```

27 s without a forwarded port, ending on the port it started with. Note the initial acquisition
*does* retry (`retrying in 5s`); only the refresh does not.

### Configuration

```yaml
environment:
  - VPN_SERVICE_PROVIDER=protonvpn
  - VPN_TYPE=wireguard
  - WIREGUARD_PRIVATE_KEY=<redacted>
  - SERVER_COUNTRIES=<redacted>
  - VPN_PORT_FORWARDING=on
  - VPN_PORT_FORWARDING_PROVIDER=protonvpn
  - VPN_PORT_FORWARDING_STATUS_FILE=/gluetun/forwarded_port
  - VPN_PORT_FORWARDING_UP_COMMAND=/set-listening-port.sh {{PORTS}}
  - FIREWALL_OUTBOUND_SUBNETS=<docker subnet>
```

### Related

- #3418 — same `recvfrom: connection refused` on refresh, closed as fixed in `v3.41.3`. The mapping
  does come back now; the teardown on a transient failure is what is left.
- #3408 / #3409 — the *reassigned port* case: different trigger, same escalation path, resolved by
  adopting the new port instead of restarting.
- #1543 — original ProtonVPN NAT-PMP support, where the timeout-only retry and the `return` come from.
- #1874 — run loop rework, where an error from `KeepPortForward` became a crash + `cleanup()`.

qdm12:
# Type

- [x] it is AI generated 🤖 and I did review it 👨👩
- [ ] it is humanly written like the good old days 👨‍🎨👩‍🎨

# Description

`KeepPortForward` refreshes the NAT-PMP mapping every 45 s for a 60 s lease. When the gateway refused a single datagram (`recvfrom: connection refused`, the Linux surface of an ICMP port unreachable), the RPC failed immediately because only timeouts were retried. That error then escalated to a full port forwarding teardown and rebuild — port file cleared, firewall rule dropped, up command re-run — even though nothing had been lost and the gateway hands back the very same port.

This implements the fix suggested in the issue: connection refused errors are now retried in `internal/natpmp/rpc.go` like timeouts, covering both the initial acquisition and the refresh, for every NAT-PMP provider. Since a refused read fails instantly (no deadline wait), it waits for the current connection duration before retrying, in a context aware way, keeping exactly the same doubled retry pace as the timeout case and the ~64 s retry window of RFC 6886 §3.1.

A transient hiccup is now invisible to the layers above: the refresh succeeds on a retry well within the 15 s of remaining lease, so there is no `cleanup()`, no port file clearing, and no spurious up command. If the gateway PMP service is genuinely down, the RPC fails after the full retry window like before, and the service restarts.

Two related adjustments:

- the final error is renamed `connection timeout:` to `connection failed:`, since it can now also report connection refused and source IP mismatch attempts;
- the ProtonVPN `+pmp` hint matched with `strings.HasSuffix(err, "connection refused")`, which no longer matches now that the final error ends with the deduplicated attempts list, so it now uses `strings.Contains`.

Tests: new `read_connection_refused_retried` case in `Test_Client_rpc` using a closed UDP port (the OS answers with ICMP port unreachable, i.e. connection refused on read), proving refused attempts are retried and reported; plus a `allocateClosedUDPPort` test helper and a per-test `maxRetries` field.

# Issue (optional)

Fix #3462

### DIFF
--- a/internal/natpmp/helpers_test.go
+++ b/internal/natpmp/helpers_test.go
@@ -6,31 +6,54 @@ import (
 	"testing"
 	"time"
 
+	"github.com/qdm12/gluetun/internal/pmtud/constants"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"golang.org/x/sys/unix"
 )
 
 // enough for slow machines for local UDP server.
 const initialConnectionDuration = 3 * time.Second
 
-// allocateClosedUDPPort returns the port of a UDP socket which is
-// closed immediately, so that no process is listening on it.
-// The OS sends back an ICMP port unreachable message for datagrams
-// sent to that port, which surfaces on the sender as a connection
-// refused error on read.
-func allocateClosedUDPPort(t *testing.T) (port uint16) {
+// reserveClosedPort returns a UDP port which is reserved by a TCP socket
+// which is bound but not listening, so that no UDP process is listening on
+// it and the OS cannot assign the port to another process. The OS sends
+// back an ICMP port unreachable message for datagrams sent to that port,
+// which surfaces on the sender as a connection refused error on read.
+func reserveClosedPort(t *testing.T) (port uint16) {
 	t.Helper()
 
-	conn, err := net.ListenUDP("udp", nil)
+	fd, err := unix.Socket(constants.AF_INET, constants.SOCK_STREAM, constants.IPPROTO_TCP)
 	require.NoError(t, err)
+	t.Cleanup(func() {
+		err := unix.Close(fd)
+		assert.NoError(t, err)
+	})
 
-	localAddress, ok := conn.LocalAddr().(*net.UDPAddr)
-	require.True(t, ok, "listening address is not UDP")
+	addr := &unix.SockaddrInet4{
+		Port: 0,
+		Addr: [4]byte{127, 0, 0, 1},
+	}
 
-	err = conn.Close()
-	require.NoError(t, err)
+	err = unix.Bind(fd, addr)
+	if err != nil {
+		_ = unix.Close(fd)
+		t.Fatal(err)
+	}
 
-	return uint16(localAddress.Port) //nolint:gosec
+	sockAddr, err := unix.Getsockname(fd)
+	if err != nil {
+		_ = unix.Close(fd)
+		t.Fatal(err)
+	}
+
+	sockAddr4, ok := sockAddr.(*unix.SockaddrInet4)
+	if !ok {
+		_ = unix.Close(fd)
+		t.Fatal("not an IPv4 address")
+	}
+
+	return uint16(sockAddr4.Port) //nolint:gosec
 }
 
 type udpExchange struct {
--- a/internal/natpmp/helpers_test.go
+++ b/internal/natpmp/helpers_test.go
@@ -6,55 +6,12 @@ import (
 	"testing"
 	"time"
 
-	"github.com/qdm12/gluetun/internal/pmtud/constants"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"golang.org/x/sys/unix"
 )
 
 // enough for slow machines for local UDP server.
 const initialConnectionDuration = 3 * time.Second
-
-// reserveClosedPort returns a UDP port which is reserved by a TCP socket
-// which is bound but not listening, so that no UDP process is listening on
-// it and the OS cannot assign the port to another process. The OS sends
-// back an ICMP port unreachable message for datagrams sent to that port,
-// which surfaces on the sender as a connection refused error on read.
-func reserveClosedPort(t *testing.T) (port uint16) {
-	t.Helper()
-
-	fd, err := unix.Socket(constants.AF_INET, constants.SOCK_STREAM, constants.IPPROTO_TCP)
-	require.NoError(t, err)
-	t.Cleanup(func() {
-		err := unix.Close(fd)
-		assert.NoError(t, err)
-	})
-
-	addr := &unix.SockaddrInet4{
-		Port: 0,
-		Addr: [4]byte{127, 0, 0, 1},
-	}
-
-	err = unix.Bind(fd, addr)
-	if err != nil {
-		_ = unix.Close(fd)
-		t.Fatal(err)
-	}
-
-	sockAddr, err := unix.Getsockname(fd)
-	if err != nil {
-		_ = unix.Close(fd)
-		t.Fatal(err)
-	}
-
-	sockAddr4, ok := sockAddr.(*unix.SockaddrInet4)
-	if !ok {
-		_ = unix.Close(fd)
-		t.Fatal("not an IPv4 address")
-	}
-
-	return uint16(sockAddr4.Port) //nolint:gosec
-}
 
 type udpExchange struct {
 	request  []byte
--- a/internal/natpmp/helpers_test.go
+++ b/internal/natpmp/helpers_test.go
@@ -13,6 +13,26 @@ import (
 // enough for slow machines for local UDP server.
 const initialConnectionDuration = 3 * time.Second
 
+// allocateClosedUDPPort returns the port of a UDP socket which is
+// closed immediately, so that no process is listening on it.
+// The OS sends back an ICMP port unreachable message for datagrams
+// sent to that port, which surfaces on the sender as a connection
+// refused error on read.
+func allocateClosedUDPPort(t *testing.T) (port uint16) {
+	t.Helper()
+
+	conn, err := net.ListenUDP("udp", nil)
+	require.NoError(t, err)
+
+	localAddress, ok := conn.LocalAddr().(*net.UDPAddr)
+	require.True(t, ok, "listening address is not UDP")
+
+	err = conn.Close()
+	require.NoError(t, err)
+
+	return uint16(localAddress.Port) //nolint:gosec
+}
+
 type udpExchange struct {
 	request  []byte
 	response []byte
--- a/internal/natpmp/rpc.go
+++ b/internal/natpmp/rpc.go
@@ -83,10 +83,9 @@ func (c *Client) rpc(ctx context.Context, gateway netip.Addr,
 
 		bytesRead, receivedRemoteAddress, err := connection.ReadFromUDP(response)
 		if err != nil {
-			shouldRetry, finalErr := retryAfterReadError(ctx, connectionDuration,
-				c.maxRetries, retryCount, err)
-			if !shouldRetry {
-				return nil, finalErr
+			if retryErr := retryAfterReadError(ctx, connectionDuration,
+				c.maxRetries, retryCount, err); retryErr != nil {
+				return nil, fmt.Errorf("reading from udp connection: %w", retryErr)
 			}
 			connectionDuration *= 2
 			failedAttempts = append(failedAttempts, err.Error())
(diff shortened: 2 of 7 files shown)


internal/natpmp/rpc_test.go
```diff
@@ -132,12 +145,24 @@ func Test_Client_rpc(t *testing.T) {
...
+				serverPort = allocateClosedUDPPort(t)
+			default:
+				remoteAddress := launchUDPServer(t, testCase.exchanges)
+				serverPort = uint16(remoteAddress.Port) //nolint:gosec
+			}
+
+			maxRetries := testCase.maxRetries
+			if maxRetries == 0 {
+				maxRetries = 1
+			}
```
qdm12: just set the maxRetries for every test case instead of having this if condition

internal/natpmp/rpc.go
```diff
@@ -124,6 +123,40 @@ func (c *Client) rpc(ctx context.Context, gateway netip.Addr,
...
+// connection refused error, which is the OS translation of an ICMP port
+// unreachable message, meaning nothing was listening on the gateway port at
+// that instant. A timed out read already waited for the connection duration
+// to elapse, so only a refused read requires an explicit wait, to keep the
+// same retry pace.
+func retryAfterReadError(ctx context.Context, connectionDuration time.Duration,
+	maxRetries, retryCount uint, readErr error,
+) (shouldRetry bool, finalErr error) {
+	if ctx.Err() != nil {
+		return false, fmt.Errorf("reading from udp connection: %w", ctx.Err())
```
qdm12: wrap the error in the caller with the context `reading from udp connection: ` instead of within this function twice. So here for example just return `ctx.Err()`

internal/natpmp/rpc.go
```diff
@@ -124,6 +123,40 @@ func (c *Client) rpc(ctx context.Context, gateway netip.Addr,
...
+	if !isTimeout && !isConnectionRefused {
+		return false, fmt.Errorf("reading from udp connection: %w", readErr)
+	}
+	if !isTimeout && retryCount+1 < maxRetries {
+		waitTimer := time.NewTimer(connectionDuration)
+		select {
+		case <-ctx.Done():
+			if !waitTimer.Stop() {
+				<-waitTimer.C
+			}
```
qdm12: no need to stop the wait timer anymore in recent go versions

internal/natpmp/rpc.go
```diff
@@ -124,6 +123,40 @@ func (c *Client) rpc(ctx context.Context, gateway netip.Addr,
...
+// read error, waiting for the retry pacing when needed. A read error is
+// retryable when it is a timeout, meaning the gateway did not respond, or a
+// connection refused error, which is the OS translation of an ICMP port
+// unreachable message, meaning nothing was listening on the gateway port at
+// that instant. A timed out read already waited for the connection duration
+// to elapse, so only a refused read requires an explicit wait, to keep the
+// same retry pace.
+func retryAfterReadError(ctx context.Context, connectionDuration time.Duration,
+	maxRetries, retryCount uint, readErr error,
+) (shouldRetry bool, finalErr error) {
```
qdm12: we don't need `shouldRetry` which is only true if `finalErr` is `nil`, so just return the `finalErr`. Actually rename `finalErr` to just `err` in this case.

internal/natpmp/rpc.go
```diff
@@ -124,6 +123,40 @@ func (c *Client) rpc(ctx context.Context, gateway netip.Addr,
...
+		return false, fmt.Errorf("reading from udp connection: %w", ctx.Err())
+	}
+	var netErr net.Error
+	isTimeout := errors.As(readErr, &netErr) && netErr.Timeout()
+	isConnectionRefused := errors.Is(readErr, syscall.ECONNREFUSED)
+	if !isTimeout && !isConnectionRefused {
+		return false, fmt.Errorf("reading from udp connection: %w", readErr)
+	}
+	if !isTimeout && retryCount+1 < maxRetries {
+		waitTimer := time.NewTimer(connectionDuration)
```
qdm12: just call it `timer`

internal/natpmp/helpers_test.go
```diff
@@ -13,6 +13,26 @@ import (
...
+func allocateClosedUDPPort(t *testing.T) (port uint16) {
+	t.Helper()
+
+	conn, err := net.ListenUDP("udp", nil)
+	require.NoError(t, err)
+
+	localAddress, ok := conn.LocalAddr().(*net.UDPAddr)
+	require.True(t, ok, "listening address is not UDP")
+
+	err = conn.Close()
```
qdm12: closing the connection would possibly allow the OS to assign the UDP port; we should do something else to make sure we are still binding to the port, even if not listening. Feel free to have a linux only implementation and just have `panic("not implemented")` for other-than-linux platforms.

There might be such implementation already present in the codebase.

internal/natpmp/rpc_test.go
qdm12: Done — `maxRetries` is now set explicitly on every test case, no more defaulting.

internal/natpmp/rpc.go
qdm12: Done — the helper now just returns `ctx.Err()` (or the raw read error), and the caller wraps it once with `reading from udp connection: %w`.

internal/natpmp/rpc.go
qdm12: Done — the `Stop()` call is removed.

internal/natpmp/rpc.go
qdm12: Done — the helper now just returns `err` (nil means the read should be retried), and the caller checks it.

internal/natpmp/rpc.go
qdm12: Done.

internal/natpmp/helpers_test.go
qdm12: Done — there was indeed an existing implementation: `reserveClosedPort` in `internal/pmtud/tcp/helpers_test.go`. It binds a TCP socket without listening, which keeps the port reserved (so the OS cannot assign it to another process) while UDP datagrams sent to that port still get an ICMP port unreachable, i.e. connection refused on the sender (verified empirically). Reused the exact same pattern for the natpmp test helper — it works on both unix and windows via `internal/pmtud/constants`, so no linux-only panic is needed.

internal/natpmp/rpc.go
```diff
@@ -82,16 +83,13 @@ func (c *Client) rpc(ctx context.Context, gateway netip.Addr,
 
 		bytesRead, receivedRemoteAddress, err := connection.ReadFromUDP(response)
 		if err != nil {
-			if ctx.Err() != nil {
-				return nil, fmt.Errorf("reading from udp connection: %w", ctx.Err())
+			if retryErr := retryAfterReadError(ctx, connectionDuration,
```
qdm12: let's call this `fatalErr` - it's not really a retryErr, since nothing was retried in that function. Actually `retryAfterReadError` naming is misleading, this just checks the error and waits, let's rename.

internal/natpmp/helpers_test.go
```diff
@@ -6,13 +6,56 @@ import (
...
+	"github.com/qdm12/gluetun/internal/pmtud/constants"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"golang.org/x/sys/unix"
 )
 
 // enough for slow machines for local UDP server.
 const initialConnectionDuration = 3 * time.Second
 
+// reserveClosedPort returns a UDP port which is reserved by a TCP socket
```
qdm12: since this in helpers_test.go, compilation on i.e. windows will fail. We want compilation to succeed on all platforms including windows, even if running tests/the program fails with a panic.

qdm12: The comment seems off, shouldn't it be a UDP socket instead of TCP?

internal/natpmp/rpc.go
qdm12: Done — renamed the variable to `fatalErr` and the function to `handleReadError`.

internal/natpmp/helpers_test.go
qdm12: Done — split into `helpers_unix_test.go` (`//go:build linux || darwin`) with the real implementation, and `helpers_unspecified_test.go` (`//go:build !linux && !darwin`) with a stub which panics "not implemented". Verified the test binary now cross-compiles for both windows and darwin (`GOOS=windows go test -c` / `GOOS=darwin go test -c`).

qdm12: The socket is deliberately TCP: a bound UDP socket is a UDP listener by definition, so it would receive the client datagrams instead of refusing them (the test would time out rather than get a connection refused). Binding a TCP socket (without listening) only reserves the port number so the OS cannot assign it to another process, while the UDP side of that port stays unreachable. I expanded the comment to make this distinction explicit.

PR #3465, opened 2026-09-08
fix(command): call Wait after pipe reads are complete, to not lose output lines

qdm12:
# Type

- [x] it is AI generated 🤖 and I did review it 👨👩
- [ ] it is humanly written like the good old days 👨‍🎨👩‍🎨

# Description

`cmd.Wait()` was called concurrently with the stdout/stderr pipe reads in `internal/command/start.go`. But `Wait` closes the pipes (it is the one that owns the parent ends created by `StdoutPipe`/`StderrPipe`), and closing a pipe discards the data still buffered in the kernel. So when a command exited while its last output line was still buffered, that line could be silently lost — violating the documented contract of `StdoutPipe`/`StderrPipe` ("It is thus incorrect to call Wait before all reads from the pipe have completed").

This made `Test_Service_runCommand` flaky under full-suite load (missing call to `logger.Info` — hit twice out of two full validation runs), and could silently drop command outputs in production (notably the port forwarding up/down commands run through `RunAndLog`).

The fix waits for all the pipe reads to be complete before calling `Wait`, which is safe since a pipe only reaches EOF once the command (and any process holding its file descriptors) has exited. A side benefit: if a command spawns a long-lived background process inheriting its stdout, the output is now fully drained and the wait error is delivered once the stream is complete, instead of being cut short when the direct child exits.

Found while validating PR #3464, split out per maintainer request.

Reproduction: a parallel stress test running the old pattern with real commands lost 1057 output lines out of 3200 (33%); after the fix, 0/3200.

A real-command regression test `Test_start_real_command` is added (linux, runs 64 commands in parallel); it fails without the fix.

# Issue (optional)

Found while validating #3464 (no dedicated issue).


internal/command/start_linux_test.go
```diff
@@ -0,0 +1,59 @@
+//go:build linux
```
qdm12: We can extend this to unix (linux and darwin)

internal/command/start_linux_test.go
```diff
@@ -0,0 +1,59 @@
...
+	t.Parallel()
+
+	const workers = 64
+
+	var lostLines atomic.Int64
+	var wg sync.WaitGroup
+	for range workers {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
```
qdm12: shouldn't all workers wait on a waitgroup to run the command all at the same time?

internal/command/start_linux_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+			cmd := exec.CommandContext(t.Context(), "/bin/sh", "-c", "echo line")
```
qdm12: shouldn't we try writing multiple lines (not thousands either) in that command?

internal/command/start_linux_test.go
qdm12: Done — the build tag is now `linux || darwin` and the file renamed to `start_unix_test.go` (`/bin/sh` is available on both).

internal/command/start_linux_test.go
qdm12: Done — the workers now wait on a start barrier (a released channel) before launching their command, so all 64 commands start at the same time.

internal/command/start_linux_test.go
qdm12: Done — the command now outputs 100 lines (`echo 0; echo 1; ...; echo 99`), and the test checks every line is received, in order: any missing, out-of-order or wrong line is counted as lost.

PR #3471, opened 2026-09-13
fix(cmd/gluetun): directory mode on /tmp/gluetun should be 0755

evan314159:
# Type

Please tick which one the following applies to your pull request:

- [ ] it is AI generated 🤖 and I did review it 👨👩
- [X] it is humanly written like the good old days 👨‍🎨👩‍🎨

# Description

Correct directory mode on /tmp/gluetun to be 0755 (from 0644).

# Issue (optional)


qdm12:
why ?

evan314159:
The original 0644 is missing the execute bit so non-root users cannot change into the directory.  0755 adds the execute bit.

evan314159:
Also, I checked the rest of the code base and this is the only function that creates directories with the wrong permissions.  The rest use 0755 or 0751.  I renamed the constant to dirPermissions as per ./internal/storage/flush.go to avoid confusion in future.

qdm12:
why do you need the execute bit for all users ? also if that's not done already the directory should be owned by what the user specifies with PUID and PGID

evan314159:
0644 is not a valid directory mode. A 0644 directory can't be traversed even by root, except via DAC_OVERRIDE capability that allows root to bypass permission checks, which is why 0644 appears to work enough with default capabilities, at least until one tries to share the directory with the non-root user.  I hit this running with all capabilities dropped.

Gluetun also chowns the port mapping files in /tmp/gluetun to PUID:PGID.  This further needs DAC_OVERRIDE to update them because with file mode 0644 once they have been chownd to PUID:PGID root no longer has permissions to update them.

In the PR I set 0755 on the directories assuming world-read was intended, since the files themselves are created 0644, and that directory mode 0644 was likely not intended. Can you confirm the intent?

Two alternatives depending on the answer:

If /tmp/gluetun is private by default (admin overrides to share):

* Directory: root:root 0700 initially and the admin changes afterwards as desired
* Files: one of the following is likely correct:
    * root:root 0644: root writes, world reads (avoids needing to chown or chmod the files, depends on the directory permissions to restrict access)
    * root:PGID 0640: root writes, non-root user reads
    * root:PGID 0660: root and non-root both write (matches the current chown PUID:PGID behaviour without depending on DAC_OVERRIDE)

If /tmp/gluetun is meant to be shared with the non-root user by default:

* Directory: root:PGID 0750
* Files: same options as above

Controlling access by PGID is better than by PUID both because it does not depend on DAC_OVERRIDE and because Kubernetes securityContext.fsGroup can be used to fix the volume permissions and assign a secondary GID where no such capability exists for controlling access by PUID.

For /gluetun, if it's meant to be private then 0700 would be appropriate.

Please let me know which model was intended and I'll revise the PR accordingly.

cmd/gluetun/main.go
```diff
@@ -292,12 +292,12 @@ func _main(ctx context.Context, buildInfo models.BuildInformation,
 		logger.Warn(warning)
 	}
 
-	const permission = fs.FileMode(0o644)
-	err = os.MkdirAll("/tmp/gluetun", permission)
+	const dirPermission = fs.FileMode(0o755)
```
qdm12, line 295: rename back to `permission`

qdm12 (approved):
Makes sense, my bad for misunderstanding!

qdm12:
Actually, ideally, I think we should:

- Chown directories to PUID:PGID
- Set permission of directories to 0o770 and files to 0o660

So we have the same permissions for both PUID and PGID.

Now to avoid breaking compatibility, let's keep directories to 0o774 and files to 0o664 (if the world permission was `4` before), with a comment `// v4: set permission to 0 for world group` on each chmod calls

evan314159:
Updated to your specifications.  With default umask 022 to get file modes like 664 we need to use chmod.

cmd/gluetun/main.go
```diff
@@ -290,12 +291,27 @@ func _main(ctx context.Context, buildInfo models.BuildInformation,
...
+	switch {
+	case errors.Is(err, os.ErrNotExist):
+		// TODO v4: remove world permission (0770 or 2770)
+		// TODO v4: chown root:PGID for sharing with nonrootuser
+		const tmpPermission = fs.FileMode(0o774)
+		err = os.MkdirAll("/tmp/gluetun", tmpPermission)
+		if err != nil {
+			return err
+		}
+		// chmod explicitly since MkdirAll is subject to the umask
```
qdm12: uh nice 👍

cmd/gluetun/main.go
```diff
@@ -290,12 +291,27 @@ func _main(ctx context.Context, buildInfo models.BuildInformation,
...
+		if err != nil {
+			return fmt.Errorf("setting /tmp/gluetun permissions: %w", err)
+		}
+	case err != nil:
+		return fmt.Errorf("checking /tmp/gluetun: %w", err)
 	}
-	err = os.MkdirAll("/gluetun", permission)
+
+	const gluetunPermission = fs.FileMode(0o755)
+	err = os.MkdirAll("/gluetun", gluetunPermission)
```
qdm12: let's move this to a local `createDirectories` function, since it's getting pretty big (and `_main` is already pretty too-big)

qdm12: should `/gluetun` get the same treatment as `/tmp/gluetun`? (same checks as switch above)

internal/portforward/service/fs.go
```diff
@@ -19,12 +19,21 @@ func (s *Service) writePortForwardedFile(ports []uint16) (err error) {
...
 		return fmt.Errorf("writing file: %w", err)
 	}
+	// chmod explicitly since WriteFile is subject to the umask
+	err = os.Chmod(filepath, perms)
+	if err != nil {
+		return fmt.Errorf("setting file permissions: %w", err)
+	}
 
+	// TODO v4: set UID to -1 (root:PGID)?
+	// files chowned to PUID require DAC_OVERRIDE to update
```
qdm12: I'm not sure I understand, `0` is for root. But why

> files chowned to PUID require DAC_OVERRIDE to update

If you match PUID to your user id? Sorry if I'm missing something!

PR #3477, opened 2026-09-16
fix(protonvpn): allow the servers updater to log in with an external email account

christophervalles:
### Problem

ProtonVPN accounts whose login is an **external address** (for example a Gmail one) have no Proton username. `POST /core/v4/auth/info` returns an empty `Username` alongside perfectly valid SRP parameters:

```
Code=1000  Version=4  Username=''  Modulus=present  ServerEphemeral=present  Salt=present  SRPSession=present
```

`authInfo` rejects that response, so these accounts can never update their server data:

```
ERROR updating server information: getting protonvpn servers: authentifying with Proton:
getting auth information: data field missing in response: username is empty
```

The practical effect is that the Proton server list silently never updates — `UPDATER_PERIOD` logs `credentials missing`/this error every cycle, and gluetun keeps using the hardcoded list (301 days old in my case), repeatedly dialling servers that no longer exist.

### Why the field isn't needed here

`info.Username` is used in exactly one place, to build the SRP password hash:

```go
srpAuth, err := srp.NewAuth(version, username, []byte(password), saltBase64, modulusPGPClearSigned, serverEphemeralBase64)
```

and `ProtonMail/go-srp`'s `HashPassword` ignores it for modern auth versions:

```go
switch authVersion {
case 4, 3:
	return hashPasswordVersion3(password, salt, modulus)   // userName not passed
case 2:
	return hashPasswordVersion2(password, userName, modulus)
case 1:
	return hashPasswordVersion1(password, userName, modulus)
case 0:
	return hashPasswordVersion0(password, userName, modulus)
}
```

The subsequent `/core/v4/auth` call sends the **email** anyway (`c.auth(ctx, unauthCookie, email, ...)`), not this field.

### Change

Require `Username` only for the legacy auth versions that actually consume it (0–2), and move the `Version` nil-check above it so the version is known at that point. One `case`, no behaviour change for accounts that do return a username.

### Verification

Built from this branch and ran the updater against a version-4 external-email account. Before the patch it fails immediately at `username is empty`. After it, authentication proceeds normally into Proton's regular login flow — in my case reaching human verification, which is a separate, later stage handled elsewhere in this file:

```
INFO updating Protonvpn servers...
WARN Proton is asking for human verification to continue authentication, please open the following URL ...
```

That confirms the SRP exchange itself is fine without the username.


qdm12 (approved):
Thanks!

PR #3481, opened 2026-09-18
fix(httpproxy): forward bytes buffered by the HTTP server before Hijack

zencamp:
## The bug

`internal/httpproxy/https.go`'s CONNECT handler discards the `*bufio.ReadWriter` that `Hijack()` returns, keeping only the raw `net.Conn`:

```go
clientConnection, _, err := hijacker.Hijack()
```

`net/http` keeps a background read running on the connection while a body-less handler (such as CONNECT) executes. `Hijack()` flushes the already-written `200` response to the client *before* it cancels that background read. A client that writes immediately after seeing the `200` — a TLS ClientHello a millisecond or two later, which is common on a fast local network or with a client that pipelines the write — can land inside that window: the background read consumes the first byte(s) of the client's write and hands them back inside the `*bufio.ReadWriter` this handler throws away with `_`.

Forwarding the raw connection instead of that buffer silently drops the leading byte(s), so the destination receives a stream that's off by one (or more) bytes. For a TLS handshake specifically, the ClientHello arrives missing its `0x16` record-type prefix, and the peer rejects it as not TLS — visible client-side as `wrong version number` (OpenSSL/Node) or `first record does not look like a TLS handshake` (uTLS-style clients).

This reproduced at a measurable rate against a real destination in production use of this proxy — roughly 5-7% of CONNECT requests from a typical client, and up to ~21% from a client with sub-millisecond timing between reading the response and writing its first bytes.

## The fix

Keep the `*bufio.Reader` `Hijack()` returns and read the client side through it instead of the bare connection. `*bufio.Reader` isn't an `io.ReadCloser`, so `bufferedClientConn` combines it with the connection's `Close`.

## Test

`hijack_firstbyte_test.go` opens 4000 CONNECT requests against the real handler, writing the "client" bytes immediately after seeing the `200` (matching the timing that triggers the race), and asserts every byte arrives at the destination intact. Against the pre-fix code this fails intermittently in CI-like environments; 0/4000 with the fix applied. The existing `httpproxy` tests were unaffected by the bug because none of them write inside the hijack window, which is why it went unnoticed.

```
=== RUN   Test_handleHTTPS_firstByteNotDropped
    hijack_firstbyte_test.go:101: connections=4000 client_errors=0 dest_errors=0 first_byte_lost=0
--- PASS: Test_handleHTTPS_firstByteNotDropped (2.14s)
```


qdm12:
Also

> The test is the weak part. It relies on winning a race, 4000 sequential dials (~2 s, ephemeral-port churn), and t.Fatalf on any clientErr/destErr — that's flaky under loaded CI. It also ignores repo conventions (no testify table, 
   no t.Parallel, context.WithCancel instead of t.Context(), hand-rolled nopLogger instead of the gomock pattern, not gofumpt-formatted).                                                                                                
   Better: add a deterministic case where the client sends CONNECT + first tunnel bytes in a single write (the catastrophic variant above — 100% failure pre-fix, zero timing dependence), and keep the race variant as a smaller-N      
   (e.g. 500) supplementary check.

